PR #11, opened 2021-07-11
Use example values for CF properties

chase:
This is intended to be merged after https://github.com/mrbbot/node-fetch/pull/1

This uses the example values provided in the documentation instead of mostly empty and some invalid values which likely break some assumptions in scripts that use the information.

I will make another PR next weekend using fast-geoip for slightly more accurate geo-IP information.


test/index.spec.ts
```diff
@@ -263,18 +264,38 @@ test("createServer: includes cf property on request", async (t) => {
...
+    t.regex(body.latitude, /^-?\d+\.\d+/);
+  } else {
+    t.fail();
+  }
+
+  if (body.longitude) {
+    t.regex(body.longitude, /^-?\d+\.\d+/);
+  } else {
+    t.fail();
+  }
```
mrbbot, line 292: Can we condense this a little bit? The empty string will always fail the regexp.
```suggestion
  t.regex(body.latitude ?? "", /^-?\d+\.\d+/);
  t.regex(body.longitude ?? "", /^-?\d+\.\d+/);
```

test/index.spec.ts
```diff
@@ -263,18 +264,38 @@ test("createServer: includes cf property on request", async (t) => {
...
-    tlsCipher: "",
-    tlsClientAuth: null,
-    tlsVersion: "",
-    timezone: "",
-  });
+  const body: NonNullable<RequestInit["cf"]> = JSON.parse(
+    (await request(origin))[0]
+  );
+
+  t.true(body.asn > 0, "ASN is a valid number");
```
mrbbot, line 271: `>` will perform type coercion so `"1" > 0` is true, but we don't want ASN to be a string.
```suggestion
  t.true(typeof body.asn === "number");
```

mrbbot (commented):
👋 Looks good! Just a couple of comments on the tests.

How are you thinking of using `fast-geoip` for more accurate geo-IP information? Will you make a request to a third-party service to get the user's IP address? Ideally, we wouldn't do that, though I guess it would be an optional feature? I'm also a little concerned that `fast-geoip` is quite a large package (~164MB). I know I'm already including the TypeScript compiler (~61 MB) just for CJS modules support, but ideally I wouldn't be. Even as an `optionalDependency` Node would still try to install it.  If you are going to implement it, do you think you could not include it in the package dependencies and just enable the feature if the user has it installed?

test/index.spec.ts
chase: You're absolutely correct that this is overly verbose. I wanted to simply use the non-null assertion operator (`body.latitude!`), but there is a linting rule that does not permit the non-null assertion operator.

Rather than doing nullish coalescing to an empty string (which will always fail the grep as well), I chose to be explicit without violating the linting rule.

Whatever you're happy with works for me, but I think that disabling the linting rule for that test's function would probably be the most straight-forward.

test/index.spec.ts
chase: My message was a poor descriptor, I meant it is a valid autonomous system number (ASN). 0 is an invalid _autonomous system_ number, which was the case I intended to write against. I didn't give much thought to the other cases.

All non-positive, non-integer values are not valid ASNs. There are in fact many non-valid numbers, but I also don't think it is crucial to test the ASN to be exactly to spec, just mostly correct.

This suggest doesn't validate the ASN is greater than 0 nor that it is an integer, so perhaps this is better?
```suggestion
  t.true(Number.isInteger(body.asn));
  t.true(body.asn > 0);
```

mrbbot: Ah right, that last suggestion looks good then.

test/index.spec.ts
mrbbot: I'd prefer not to disable linting rules, so could we use nullish coalescing here?

chase:
> How are you thinking of using `fast-geoip` for more accurate geo-IP information? Will you make a request to a third-party service to get the user's IP address? Ideally, we wouldn't do that, though I guess it would be an optional feature? I'm also a little concerned that `fast-geoip` is quite a large package (~164MB).

Using `fast-geoip` would ensure that this is not connecting to a third-party service at all. I was intentionally trying to avoid this, since it would defeat the purpose of an offline simulator. But you're right, it is one chunky module and as I said in my previous issue, I would want to make this optional since I don't think my use case for it is very common.

It would simply take the actual client IP and provide the necessary information. It is good enough to get cities correct for the most part along with corresponding coordinates, and that is largely my use case.

> Even as an `optionalDependency` Node would still try to install it. If you are going to implement it, do you think you could not include it in the package dependencies and just enable the feature if the user has it installed?

That's pretty much what I was thinking. I wasn't particularly happy with the changes in NPM 7 regarding optionalDependencies and handling of peerDependencies that are marked as optional, but I can understand the rationale behind it.

My intention was to update the documentation and/or README to say it can be enabled by running `npm install --no-save fast-geoip`. Then in the actual code, just try to require it to enable the feature.

chase:
> Where would you be getting the client IP from?

`const ip = req.headers['x-forwarded-for'] || req.socket.remoteAddress;`

Similar to an actual Cloudflare zone, a worker can be applied to specific paths. In such cases, I have a local Nginx that provides a reverse proxy (`x-forwarded-for` comes from it).

> Or are you thinking about using Miniflare as an actual production server?

No, a couple things:

- The same Nginx configuration I mention above as a Docker image on a series of tests as a service. I intend to use this for testing the sanity of workers before deploying to a staging environment.
- I use [ngrok](https://ngrok.com/product) to validate and receive feedback on changes with my clients without deploying to a live environment

chase:
It's also worth mentioning it only shows 127.0.0.1 if you use localhost or 127.0.0.1 to access your example server.

If you access the server from another device on your network, it will have a proper address. Say your computer running this server's local network address is 192.168.1.5 and your cell phone is 192.168.1.128, if you go to `192.168.1.5:3000` on your phone, you will get `::ffff:192.168.1.128` (that's IPv4 to IPv6 translation, but that's what you'll see most likely)

chase:
I've implemented the changes to the tests.

My additions using fast-geoip will be done in another PR if that's alright.

PR #21, opened 2021-07-30
chore: swap semver -> semiver

lukeed:
Replaced the very simple, single usage of semver with a dependency that only deals with the desired use case.

`semiver` is about 10 LOC, making for ~6kB total package size.
Because `semver` does significantly more (99% unused), its total package size is around 90kB.


lukeed:
No problem, this pr and #22 will fail CI though because of the package-lock issue I mentioned to you. Sorry :/

PR #78, opened 2021-10-31
chore: add `engines.node` constraint

lukeed:
Provides install-time Node version checks, preventing users with outdated/unsupported `node` versions from installing miniflare.

This `16.7` is the value reflected as of [v2.0.0-next.1](https://github.com/cloudflare/miniflare/releases/tag/v2.0.0-next.1) & was confirmed by @mrbbot  in conversation

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,6 +54,9 @@
     "vitepress": "^0.20.0",
     "which": "^2.0.2"
   },
+  "engines": {
+    "node": ">=16.7"
+  },
   "volta": {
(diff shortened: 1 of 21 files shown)


mrbbot (commented):
Could you add this to `scripts/scaffold.mjs` too, so newly generated packages also include it?

lukeed:
Ah sorry, hadn't looked at that script. Updated

mrbbot (approved):

PR #79, opened 2021-10-31
fix(http-server): append livereload to end of document;

### LINKED ISSUE #70: Miniflare Live Reload script not included if closing `body` tag is missing
Using `miniflare@next` and `--live-reload`, the Miniflare Live Reload script tag is only written when a `</body>` tag exists. I'm not sure whether this is an HTMLRewriter implementation issue, but the script should probably be written from the `end` event of the document handler.

lukeed:
Scripts are still valid & will still execute, even if outside of the `<html>` tags. The browsers' HTML parsers are incredibly loose/forgiving haha

Here's a valid HTML document, which still logs to the console:

```html
<html>
  <title>hello</title>
  <p>content</p>
</html>
<script defer>
  console.log('loaded')
</script>
```

Closes #70

### DIFF
--- a/packages/http-server/package.json
+++ b/packages/http-server/package.json
@@ -37,7 +37,6 @@
   },
   "dependencies": {
     "@miniflare/core": "2.0.0-next.2",
-    "@miniflare/html-rewriter": "2.0.0-next.2",
     "@miniflare/shared": "2.0.0-next.2",
     "@miniflare/web-sockets": "2.0.0-next.2",
     "kleur": "^4.1.4",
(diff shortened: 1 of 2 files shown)


lukeed:
Yeah, I don't think HTMLRewriter is needed at all anymore, but I felt bad removing it outright :D

mrbbot (approved):

lukeed:
I noticed the same thing while playing with the rewriter in cf playground. Have a note to open an internal issue tomorrow

lukeed:
Oops – total brain fart. The CF playground implementation is okay – at least while returning `rewriter.transform(res)` directly. Because it's streamed, the `transfer-encoding: chunked` header is added and any `content-length` value is omitted, even if you try to force the wrong value.

PR #80, opened 2021-11-01
chore: bump `undici` version;

### LINKED ISSUE #64: Miniflare is not rewriting headers correctly
Let's say we want to use Cloudflare Workers as just a proxy and rewrite headers passed with the `x-` prefix.

**Current behavior**

```js
const { Miniflare } = require('miniflare')

function listener (event) {
  const response = new Response('', { headers: event.request.headers })

  for (const [key, value] of response.headers) {
    response.headers.delete(key)
    response.headers.set(`x-${key}`, value)
  }

  return event.respondWith(response)
}

;(async () => {
  const mf = new Miniflare({
    script: `addEventListener('fetch', ${listener.toString()})`
  })
  const res = await mf.dispatchFetch('http://localhost:8787/', {
    headers: {
      foo: 'bar',
      fooz: 'barz'
    }
  })

  console.log(await Object.fromEntries(res.headers))
  // => { fooz: 'barz', 'x-x-foo': 'bar' }
  // WRONG!
})()
```

**Expected behavior**

CleanShot 2021-10-13 at 16 49 13@2x

lukeed:
The `4.9.3` is the first release that includes this PR: https://github.com/nodejs/undici/pull/1081

- Closes #64

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -31,6 +31,9 @@
         "typescript": "^4.4.4",
         "vitepress": "^0.20.0",
         "which": "^2.0.2"
+      },
+      "engines": {
+        "node": ">=16.7"
       }
     },
     "node_modules/@algolia/cache-browser-local-storage": {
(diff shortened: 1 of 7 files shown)


mrbbot (approved):
Nice! 🎉 Thanks for fixing this. 🙂

PR #95, opened 2021-11-17
[jest-environment-miniflare] Add Buffer to global

dan-lee:
This is to fix an error that came up in my test suite:

> ReferenceError: Buffer is not defined
>
>    at Object.<anonymous> (node_modules/undici/lib/client.js:409:19)
>    at Object.<anonymous> (node_modules/undici/index.js:3:16)


gzuidhof:
I have also ran into this same issue (it's for a library that doesn't check for Buffer in a safe way - which are very common in Node!).

PR #118, opened 2021-12-10
avoid await top level

Kikobeats:
Hello,

This change allows being possible to use `miniflare@v2` in environments where top-level await is not supported


Kikobeats:
Hello, this change is motivated by the fact of running miniflare under cjs.

> Top-level await is currently not supported with the "cjs" output format` then I need to remove needed to replace

After that, I compiled miniflare as cjs module. It's just an experiment but it's working as expected, take a look at:
https://github.com/Kikobeats/miniflare/tree/v2-cjs

packages/core/src/plugins/rust.ts
```diff
@@ -3,28 +3,30 @@ import assert from "assert";
...
-  path.join(generatedDir, "script.wasm")
-);
+  // 3. Copy *_bg.wasm file into worker/generated/script.wasm
+  const wasmName = pkg.files.find((file) => file.endsWith(".wasm"));
+  assert(wasmName);
+  await fs.copyFile(
+    path.join("pkg", wasmName),
+    path.join(generatedDir, "script.wasm")
+  );
+})()
```
mrbbot, line 32: Could you put a semi-colon on the end here, and remove the one at the top of the `async` function? This matches the default style of `prettier` we use.

mrbbot (commented):
Ok, let's go ahead with official CommonJS support. 👍

packages/jest-environment-miniflare/test/index.spec.ts
```diff
@@ -22,13 +22,12 @@ async function findJest() {
...
 
-const jestPath = await findJest();
-
-function runJest(
+async function runJest(
   match: string,
   options: MiniflareOptions = {},
   cwd = fixturesPath
 ): Promise<[exitCode: number, output: string]> {
+  const jestPath = await findJest();
```
mrbbot, line 30: Maybe do this too, so we only `findJest` once?
```suggestion
const jestPathPromise = findJest();

async function runJest(
  match: string,
  options: MiniflareOptions = {},
  cwd = fixturesPath
): Promise<[exitCode: number, output: string]> {
  const jestPath = await findJestPromise;
```

Kikobeats:
> Ok, let's go ahead with official CommonJS support. 👍

that could be awesome, how I can help? 🙂

Kikobeats:
done at https://github.com/cloudflare/miniflare/pull/119

PR #119, opened 2021-12-10
Add CJS support

Kikobeats:
TODO:

-  [ ] Expose packages in a way is still possible to be loaded using `esm`

### DIFF
--- a/packages/cache/package.json
+++ b/packages/cache/package.json
@@ -11,8 +11,7 @@
   ],
   "author": "MrBBot <me@mrbbot.dev>",
   "license": "MIT",
-  "type": "module",
-  "exports": "./dist/src/index.js",
+  "main": "./dist/src/index.js",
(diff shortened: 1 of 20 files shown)


Kikobeats:
@mrbbot what do you think about using run esbuild twice, one for generating `esm` and another one for generating `cjs` output?

In that way, the current functionality of the library will preserve, plus it will add the ability to load it as a require module.

Just needs to generate separate destination folders and link properly with package.json files.

I can do it if you agree with the approach 🙂

Kikobeats:
maybe as separate thing? Like `@miniflare/cjs`.

About the named export, it could be as easy to refactor the code to avoid the default export?

Kikobeats:
@mrbbot happy if you want to make the change, I don't feel too confident with the codebase yet 😿

Kikobeats:
wow that was fast. Excited to test it 🎉

PR #127, opened 2021-12-16
Fix Typo setImmediate -> setImmediatePromise

worenga:


worenga:
@mrbbot i have jest test that tests a durable object and looks like this:

in that file we encounter
 `import { setImmediate as setImmediatePromise } from "timers/promises";`

```
import { MyDurableObject } from './MyDurableObject';
import { MemoryStorage } from '@miniflare/storage-memory';
import { DurableObjectStorage, DurableObjectState, DurableObjectId } from '@miniflare/durable-objects';

let storage = new DurableObjectStorage(new MemoryStorage());
let obj: MyDurableObject;

const createObj = () => {
    return new DurableObject(new DurableObjectState(new DurableObjectId('test', 'test'), storage), {});
};

beforeEach(() => {
    storage = new DurableObjectStorage(new MemoryStorage());
    obj = createObj();
});

it(`works`, async () => {
    const consume1 = await obj.fetch(

....
        }),
...
    );
```

When i run this i get

```
 FAIL  src/durable_objects/TokenStore.test.ts
  ✕ works (7 ms)

  ● works

    ReferenceError: setImmediate is not defined

      36 |         switch (url.pathname) {
      37 |             case '/consume': {
    > 38 |                 return await this.state.blockConcurrencyWhile(async () => {
         |                        ^

      at InputGate.runWithClosed (node_modules/@miniflare/shared/src/sync/gate.ts:106:7)
      at MyDurableObject.fetch (src/durable_objects/MyDurableObject.ts:38:24)
      at Object.<anonymous> (src/durable_objects/MyTest.test.ts:18:22)

```

worenga:
@mrbbot does that matter?
setImmediate is not even defined?

Issue #128, opened 2021-12-17
Environment Issue when including 'promjs'


worenga:
I wanted to test my worker using miniflare.
I realized when i include and use the 'promjs' package running the example breaks the jest tests.

```
 FAIL  test/counter.spec.ts
  ● Test suite failed to run

    TypeError: Cannot read properties of undefined (reading 'ceil')

       7 | var __markAsModule = (target) => __defProp(target, "__esModule", { value: true });
       8 | var __commonJS = (cb, mod) => function __require() {
    >  9 |   return mod || (0, cb[Object.keys(cb)[0]])((mod = { exports: {} }).exports, mod), mod.exports;
         |                                            ^
      10 | };
      11 | var __reExport = (target, module, desc) => {
      12 |   if (module && typeof module === "object" || typeof module === "function") {

      at runInContext2 (node_modules/lodash/lodash.js:1531:27)
      at __require (dist/index.mjs:9:44)
      at __require (dist/index.mjs:9:44)
      at src/index.ts:1:26

```
I took the example from `miniflare-typescript-esbuild-jest` and added promjs, here is what i changed: 
https://github.com/mrbbot/miniflare-typescript-esbuild-jest/pull/4/commits/fc283580ebcfca46c47fe14ed1c0cd925c46665c

Inspecting the lodash import it seems this is somehow referencing native math functions:

```
 1531     var nativeCeil = Math.ceil,
 1532         nativeFloor = Math.floor,
 1533         nativeGetSymbols = Object.getOwnPropertySymbols,
 1534         nativeIsBuffer = Buffer ? Buffer.isBuffer : undefined,
 1535         nativeIsFinite = context.isFinite,
 1536         nativeJoin = arrayProto.join,
 1537         nativeKeys = overArg(Object.keys, Object),
 1538         nativeMax = Math.max,
 1539         nativeMin = Math.min,
 1540         nativeNow = Date.now,
 1541         nativeParseInt = context.parseInt,
 1542         nativeRandom = Math.random,
 1543         nativeReverse = arrayProto.reverse;

```

mrbbot:
[...]

PR #129, opened 2021-12-17
Export Math to global object

worenga:


worenga:
Fixes #128

worenga:
@mrbbot yes, it fixed it.

worenga:
Thanks!

PR #155, opened 2022-01-18
Fixed calling getPrototypeOf(null)

### LINKED ISSUE #154: `ordinaryHasInstance` throws Cannot convert undefined or null to object
https://github.com/cloudflare/miniflare/blob/a651b20d9b268725360982af9a566dd0712db5e7/packages/runner-vm/src/instanceof.ts#L99

The variable `O` is possible to be `null` or `undefined` and calling `getPrototypeOf` that will throw an error.

`Cannot convert undefined or null to object`

I found this bug when return null on resolver with graphql and miniflare v2.1 (work fine on v2.0).

Test with graphql 15, node 16-17

Cerberus:
I tried to make a simple solution to fix #154.

### DIFF
--- a/packages/runner-vm/src/instanceof.ts
+++ b/packages/runner-vm/src/instanceof.ts
@@ -85,7 +85,7 @@ function ordinaryHasInstance(C: any, O: any): boolean {
   // 2. If C has a [[BoundTargetFunction]] internal slot, ... (IGNORED)
   // 3. If Type(O) is not Object, return false.
   if (typeof O !== "object" && typeof O !== "function") return false;
-  if (O === null) return false
+  if (O === null) return false;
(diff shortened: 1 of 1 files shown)


mrbbot (changes requested):
Thank you very much for the PR! 😃 Looks like this is because even though `Type(null)` is `Null`, `typeof null` is `"object"`: https://tc39.es/ecma262/multipage/ecmascript-language-expressions.html#table-typeof-operator-results, oops. 🙃 

Would you be able to move the `=== null` check to just under `3. If Type(O) is not Object...`, so it follows the spec?

```diff
  if (typeof O !== "object" && typeof O !== "function") return false;
+ if (O === null) return false;
```

Issue #165, opened 2022-01-28
500 Internal Server Error / on 304 response and empty ReadableStream


lukaszczerpak:
It seems there is inconsistency between CF and Miniflare. PoC code:

```javascript
addEventListener('fetch', function(event) {
  event.respondWith(handleRequest(event.request))
})
async function handleRequest(request) {
  let url = new URL(request.url)
  const res = await fetch(`http://origin.cloudtest.it${url.pathname}`, {headers: {"if-modified-since": "Wed, 12 Jan 2022 10:43:29 GMT"}})
  return new Response(res.body, {
    headers: {
      ...Object.fromEntries(res.headers),
      "x-workers-hello": "Hello from Cloudflare Workers"
    },
    status: res.status,
    statusText: res.statusText
  })
}
```

Throws the following error:

```
GET /bigfiles/256m 500 Internal Server Error (98.22ms)
[mf:err] GET /bigfiles/256m: TypeError: Response with null body status cannot have body
    at new Response (/Users/lukes/labs/cf-workers/sandbox-lukasz-testing/node_modules/undici/lib/fetch/response.js:160:15)
    at new Response (/Users/lukes/labs/cf-workers/sandbox-lukasz-testing/node_modules/@miniflare/core/src/standards/http.ts:556:13)
    at handleRequest (/Users/lukes/labs/cf-workers/sandbox-lukasz-testing/index.js:22:10)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at EventTarget.[kDispatchFetch] (/Users/lukes/labs/cf-workers/sandbox-lukasz-testing/node_modules/@miniflare/core/src/standards/event.ts:359:13)
    at Server.<anonymous> (/Users/lukes/labs/cf-workers/sandbox-lukasz-testing/node_modules/@miniflare/http-server/src/index.ts:167:20)
```

The origin server returns 304 response with no body (as expected). The reason of the error is that `res.body` is a ReadableStream which is empty but the object itself is not, therefore the `undici` library doesn't like it and throws an error (304 responses don't have response body). 
I was sitting on the fence if this should be reported here or maybe to the dependency's maintainer. In fact, the object type could be detected to behave properly. WDYT?

mrbbot:
Hey! 👋 Thanks for raising this. This is the right place to report the issue. Executing this code in Chrome/Firefox/`undici` gives me a `ReadableStream` `body`, whereas Cloudflare gives `null`:
```js
[...]

PR #172, opened 2022-02-07
Fix null status body response

GregBrimble:
```ts
import { fetch } from '@miniflare/core'

await fetch('https://some/url/which/returns/304')
```

fails because `response.body` always returns a `ReadableStream`, but the `new Response()` constructor expects `null` when setting some status codes.

### DIFF
--- a/packages/core/src/standards/http.ts
+++ b/packages/core/src/standards/http.ts
@@ -480,8 +480,9 @@ export interface ResponseInit extends BaseResponseInit {
 
 const kWaitUntil = Symbol("kWaitUntil");
 
-// From https://github.com/nodejs/undici/blob/3f6b564b7d3023d506cad75b16207006b23956a8/lib/fetch/constants.js#L28, minus 101
-const nullBodyStatus: (number | undefined)[] = [204, 205, 304];
+// From https://github.com/nodejs/undici/blob/3f6b564b7d3023d506cad75b16207006b23956a8/lib/fetch/constants.js#L28
+// https://fetch.spec.whatwg.org/#null-body-status
+const nullBodyStatus: (number | undefined)[] = [101, 204, 205, 304];
 
 const enumerableResponseKeys: (keyof Response)[] = [
   "encodeBody",
@@ -796,7 +797,11 @@ export async function fetch(
       headers,
     });
   } else {
-    res = new Response(baseRes.body, baseRes);
+    // https://fetch.spec.whatwg.org/#null-body-status
+    res = new Response(
+      nullBodyStatus.includes(baseRes.status) ? null : baseRes.body,
+      baseRes
+    );
   }
 
   await waitForOpenInputGate();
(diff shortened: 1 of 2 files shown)


threepointone (approved):

packages/core/src/standards/http.ts
```diff
@@ -796,7 +796,11 @@ export async function fetch(
       headers,
     });
   } else {
-    res = new Response(baseRes.body, baseRes);
+    // https://fetch.spec.whatwg.org/#null-body-status
+    res = new Response(
+      [101, 204, 205, 304].includes(baseRes.status) ? null : baseRes.body,
```
mrbbot, line 801: Nit: `nullBodyStatus` is defined earlier on in this file. This doesn't include `101`, that should probably be added.

```suggestion
      nullBodyStatus.includes(baseRes.status) ? null : baseRes.body,
```

mrbbot (commented):
Looks good! This will close https://github.com/cloudflare/miniflare/issues/165. Do you have time to write a test too: https://github.com/cloudflare/miniflare/blob/master/packages/core/test/standards/http.spec.ts? 🙂

packages/core/src/standards/http.ts
```diff
@@ -796,7 +796,11 @@ export async function fetch(
       headers,
     });
   } else {
-    res = new Response(baseRes.body, baseRes);
+    // https://fetch.spec.whatwg.org/#null-body-status
+    res = new Response(
```
alexandrukis, line 800: I am very new to miniflare, but there are so many things happening in this function.
This is already an improvement in readability IMO:

```js
const body = [101, 204, 205, 304].includes(baseRes.status) ? null : baseRes.body;
res = new Response(body, baseRes);
```

but the whole if/else could be moved to a separate function that would simply return the new response


```js
function toHybridResponse(baseRes: Response): Response { // or any other naming that fits the project conventions
  if (baseRes.type !== "opaqueredirect") {
    const body = [101, 204, 205, 304].includes(baseRes.status) ? null : baseRes.body;
    return new Response(body, baseRes);
  }

  // Unpack opaque responses. This restriction isn't needed server-side,
  // and Cloudflare doesn't support Response types anyway.
  // @ts-expect-error symbol properties are not included in type definitions
  const internalResponse = baseRes[fetchSymbols.kState].internalResponse;
  const headersList = internalResponse.headersList;
  assert(headersList.length % 2 === 0);
  const headers = new Headers();
  for (let i = 0; i < headersList.length; i += 2) {
    headers.append(headersList[i], headersList[i + 1]);
  }
  // Cloudflare returns a body here, but undici aborts the stream so
  // unfortunately it's unusable :(
  return new Response(null, {
    status: internalResponse.status,
    statusText: internalResponse.statusText,
    headers,
  });
}
```

petebacondarwin (approved):
LGTM!

mrbbot (approved):
Thank you! 😃

Issue #192, opened 2022-02-27
http-server doesn't properly terminate BYOB reader


vlovich:
Hi! 👋 
      
Firstly, thanks for your work on this project! 🙂

Today I used [patch-package](https://github.com/ds300/patch-package) to patch `@miniflare/http-server@2.3.0` for the project I'm working on.

I noticed that I had done a `readAtLeast(64kb, new Uint8Array(64kb))` and it hung on a body of 15 bytes. @jasnell had pointed out that the MDN docs don't document that if there's a pending read waiting, you have to `controller.byobRequest?.respond(0)` to have the ongoing read notice the end of the stream & unblock.

Here is the diff that solved my problem:

```diff
diff --git a/node_modules/@miniflare/http-server/dist/src/index.js b/node_modules/@miniflare/http-server/dist/src/index.js
index d7a4ebc..70ffbbc 100644
--- a/node_modules/@miniflare/http-server/dist/src/index.js
+++ b/node_modules/@miniflare/http-server/dist/src/index.js
@@ -404,7 +404,10 @@ async function convertNodeRequest(req, meta) {
       async pull(controller) {
         const { done, value } = await iterator.next();
         if (done) {
-          queueMicrotask(() => controller.close());
+          queueMicrotask(() => {
+              controller.close()
+              controller.byobRequest?.respond(0)
+          });
         } else {
           const buffer = Buffer.isBuffer(value) ? value : Buffer.from(value);
           controller.enqueue(new Uint8Array(buffer));
```

<em>This issue body was [partially generated by patch-package](https://github.com/ds300/patch-package/issues/296).</em>

vlovich:
The same problem appears in `@miniflare/core`:
```
diff --git a/node_modules/@miniflare/core/dist/src/index.js b/node_modules/@miniflare/core/dist/src/index.js
index 2909a6d..f4f25c9 100644
--- a/node_modules/@miniflare/core/dist/src/index.js
+++ b/node_modules/@miniflare/core/dist/src/index.js
@@ -529,8 +529,10 @@ var Body = class {
         } else if (value) {
           return controller.error(new TypeError(buildNotBufferSourceError(value)));
         }
-        if (done)
+        if (done) {
           controller.close();
+          controller.byobRequest?.respond(0);
+        }
       },
       cancel: (reason) => reader.cancel(reason)
     };
```

It may also be present in KV's `convertStoredToGetValue` but I'm not sure (same pattern of calling `controller.close`).

[...]

PR #194, opened 2022-03-01
Fix cases where BYOB readers don't notice end of stream

vlovich:
This resolves issue #192. The R2 project had tests that were hanging
because the await on the BYOB reader was never returning done even
though the controller was being signalled. All credit to @jasnell who
had pointed me towards this in our own internal code that was
polyfilling on Miniflare 1.4.


vlovich:
Open questions:
* Should the same fix be done in KV? packages/kv/src/namespace.ts. These two I know 100% fix the issue in R2 test cases. Probably worthwhile to add some test that reproduces this issue in Miniflare & double-check if it impacts KV. Not sure how.
* There are patterns in tests where controller.close is invoked without doing `controller.byobRequest.respond(0)` - should those also be updated?

vlovich:
Should be fixed.

PR #195, opened 2022-03-01
Add jest environment's missing dependencies

BasixKOR:
Hello! While I use this project with Yarn, I noticed the environment package has some unlisted dependencies. This will ensure the packages are installed correctly (like [jest-environment-node](https://github.com/facebook/jest/blob/main/packages/jest-environment-node/package.json#L20)) but also improves the compatibility with more strict package managers such as pnpm or Yarn Berry.

### DIFF
--- a/packages/jest-environment-miniflare/package.json
+++ b/packages/jest-environment-miniflare/package.json
@@ -45,7 +45,12 @@
     "@miniflare/sites": "2.3.0",
     "@miniflare/storage-memory": "2.3.0",
     "@miniflare/web-sockets": "2.3.0",
-    "miniflare": "2.3.0"
+    "miniflare": "2.3.0",
+    "@jest/environment": ">=27",
+    "@jest/fake-timers": ">=27",
+    "@jest/types": ">=27",
+    "jest-mock": ">=27",
+    "jest-util": ">=27"
   },
   "peerDependencies": {
     "jest": ">=27"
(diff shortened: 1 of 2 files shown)


packages/jest-environment-miniflare/package.json
```diff
@@ -45,7 +45,12 @@
     "@miniflare/sites": "2.3.0",
     "@miniflare/storage-memory": "2.3.0",
     "@miniflare/web-sockets": "2.3.0",
-    "miniflare": "2.3.0"
+    "miniflare": "2.3.0",
+    "@jest/environment": ">=27",
```
mrbbot, line 49: `@jest/environment` and `@jest/types` are only used in [`type`-only imports](https://github.com/cloudflare/miniflare/blob/b887877810a745f2a458d20da0a3dbd9f8a58cf4/packages/jest-environment-miniflare/src/index.ts#L2-L4), so should probably be `devDependencies`.

packages/jest-environment-miniflare/package.json
```diff
@@ -45,7 +45,12 @@
...
+    "@jest/fake-timers": ">=27",
+    "@jest/types": ">=27",
+    "jest-mock": ">=27",
+    "jest-util": ">=27"
```
mrbbot, line 53: I'm not too sure about this, but shouldn't these be `peerDependencies` instead, since we're expecting them to be installed by `jest`?

BasixKOR: Although I followed what Meta did on their `jest-environment-node` the package managers (including npm) doesn't treat the nested dependencies as peer dependencies (AFAIK).

PR #204, opened 2022-03-07
Remove "content-length: 0" from some requests

### LINKED ISSUE #193: Behaviour mismatch: Fetch with `content-length: 0` fails with undici error
Attempting to `fetch` with `content-length: 0` results in an error from `undici`. This is tolerated in real Workers env:

**Steps to reproduce:**
```
addEventListener("fetch", event => {
  event.respondWith(handleRequest(event.request))
})

async function handleRequest(request) {
  return fetch('https://httpbin.org/delete', {
    method: 'DELETE',
    headers: {
      'Accept': 'application/json',
      'Content-length': '0'
    }
  });
}
```

**expected behaviour**:

Request succeeds (worker returns the httpbin 200 response).

**Actual behaviour**

```
[mf:err] GET /: TypeError: fetch failed
    at Object.processResponse (/somedir/node_modules/undici/lib/fetch/index.js:176:23)
    at fetchFinale (/somedir/node_modules/undici/lib/fetch/index.js:737:17)
    at Fetch.mainFetch (/somedir/node_modules/undici/lib/fetch/index.js:712:5)
[mf:err] Cause: RequestContentLengthMismatchError: Request body length does not match content-length header
    at write (/somedir/node_modules/undici/lib/client.js:1433:37)
    at _resume (/somedir/node_modules/undici/lib/client.js:1385:29)
    at resume (/somedir/node_modules/undici/lib/client.js:1244:3)
    at connect (/somedir/node_modules/undici/lib/client.js:1229:3)
GET / 500 Internal Server Error (329.42ms)
```

----

This originates from undici's strictness; they deliberately treat the req as invalid, it's not a bug from their POV. See [undici issue 649](https://github.com/nodejs/undici/issues/649) for example. They offer a non-strict mode to turn this and other behaviours into warnings, but I don't think miniflare supports enabling this mode?

This does not match behaviour of CF deployed workers. Moreover, there is code in the wild that explicitly and deliberately adds `content-length: 0` to certain requests. For example, popular NPM lib [node-http-proxy](https://github.com/http-party/node-http-proxy/blob/9b96cd725127a024dabebec6c7ea8c807272223d/lib/http-proxy/passes/web-incoming.js#L24-L40); I believe Envoy proxy does similar.

---

(this shares some DNA with recent miniflare issues like #183 and #182: undici's design goals, while reasonable in some contexts, don't always seem to align with CF Workers' server-side-oriented `fetch` impl)

orls:
Fixes #193

### DIFF
--- a/packages/core/src/standards/http.ts
+++ b/packages/core/src/standards/http.ts
@@ -700,6 +700,12 @@ const kDefaultHeadersToRemove = [
   "user-agent",
 ];
 
+const methodsExpectingPayload = [
+  "PUT",
+  "POST",
+  "PATCH"
+];
+
 class MiniflareDispatcher extends Dispatcher {
   // dispatch, close & destroy are the only methods a Dispatcher must implement:
   // https://github.com/nodejs/undici/blob/09059fb491b4158a25981eb5598262b43a18c6ae/lib/dispatcher.js
@@ -767,6 +773,13 @@ export async function fetch(
   req.headers.delete("cf-connecting-ip");
   // Add "MF-Loop" header for loop detection
   req.headers.set(_kLoopHeader, String(ctx?.requestDepth ?? 1));
+  // Delete "content-length: 0" from bodyless requests. Some proxies add this,
+  // but undici considers it an error.
+  // See https://github.com/cloudflare/miniflare/issues/193
+  
+  if (!methodsExpectingPayload.includes(req.method) && req.headers.get("content-length") === "0") {
+    req.headers.delete("content-length");
+  }
 
   // Mark default headers for removal that aren't explicitly included
   const removeHeaders: string[] = [];
--- a/packages/core/test/standards/http.spec.ts
+++ b/packages/core/test/standards/http.spec.ts
@@ -960,6 +960,23 @@ test("fetch: removes Host and CF-Connecting-IP headers from Request", async (t) 
     "x-real-ip": "127.0.0.1",
   });
 });
+test("fetch: removes Content-Length 0 from body-less requests", async (t) => {
+  // Should remove content-length: 0 from certain methods because undici sees it
+  // as an error: https://github.com/cloudflare/miniflare/issues/193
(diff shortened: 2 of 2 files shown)


packages/core/test/standards/http.spec.ts
```diff
@@ -960,6 +960,23 @@ test("fetch: removes Host and CF-Connecting-IP headers from Request", async (t)
...
   });
 });
+test("fetch: removes Content-Length 0 from body-less requests", async (t) => {
+  // Should remove content-length: 0 from certain methods because undici sees it
+  // as an error: https://github.com/cloudflare/miniflare/issues/193
+  const upstream = (
+    await useServer(t, (req, res) => res.end(JSON.stringify(req.headers)))
+  ).http;
+  const res = await fetch(upstream, {
+    method: 'DELETE',
```
mrbbot, line 970: Nit: these should be double quotes
```suggestion
    method: "DELETE",
```

packages/core/src/standards/http.ts
```diff
@@ -767,6 +767,13 @@ export async function fetch(
   req.headers.delete("cf-connecting-ip");
   // Add "MF-Loop" header for loop detection
   req.headers.set(_kLoopHeader, String(ctx?.requestDepth ?? 1));
+  // Delete "content-length: 0" from bodyless requests. Some proxies add this,
+  // but undici considers it an error.
+  // See https://github.com/cloudflare/miniflare/issues/193
+  const expectsPayload = ["PUT", "POST", "PATCH"];
```
mrbbot, line 773: Nit: could you move this constant array outside the `fetch` function so it's only created once?

mrbbot (commented):
Hey! 👋 Thanks for the PR! And a test too! 😍 
A couple very minor things, then I'll merge.

orls:
Thanks; fixed the nits!

mrbbot (approved):

PR #205, opened 2022-03-08
generate "real" paths for identifiers when linking modules.

threepointone:
ref: https://github.com/cloudflare/wrangler2/issues/443, https://github.com/cloudflare/wrangler2/issues/416#issuecomment-1054815079

When running the miniflare cli from a different cwd than the project, and the modules are in temporary directories, we can get into a situation where process.cwd() and referenced modules are in `/private/var/...` and `/var/...` On macs, these are symlinks to each other, but node doesn't know that. So we use realpathSync to resolve `identifier` to a "real" path.

--- 

I tried this out locally by editing `node_modules/@miniflare/runner-vm/dist/src/index.js` in wrangler and adding this at line 130 - 

```js
      try {
      // When running the miniflare cli from a different cwd than the project,
      // and the modules are in temporary directories, we can get into a
      // situation where process.cwd() and referenced modules are in /private/var/...
      // and /var/... On macs, these are symlinks to each other, but node doesn't
      // know that. So we use realpathSync to resolve `identifier` to a "real" path.
      // This is wrapped in a try/catch because we also generate path identifiers for
      // node builtins, etc, and we generate a separate error for that further down.
      identifier = import_path.isAbsolute(identifier)
        ? import_fs.realpathSync(identifier)
        : identifier;
    } catch (err) {
      // silently fail
    }
```

and it "fixed" my issues.

### DIFF
--- a/packages/runner-vm/src/linker.ts
+++ b/packages/runner-vm/src/linker.ts
@@ -1,4 +1,4 @@
-import { readFileSync } from "fs";
+import { readFileSync, realpathSync, existsSync } from "fs";
 import fs from "fs/promises";
 import { builtinModules } from "module";
 import path from "path";
@@ -61,10 +61,22 @@ export class ModuleLinker {
     }
 
     const additionalModule = this.additionalModules[spec];
-    const identifier = additionalModule
+    let identifier = additionalModule
       ? spec
       : // Get path to specified module relative to referencing module
         path.resolve(path.dirname(referencing.identifier), spec);
+
+    // When running the miniflare cli from a different cwd than the project,
+    // and the modules are in temporary directories, we can get into a
+    // situation where process.cwd() and referenced modules are in /private/var/...
+    // and /var/... On macs, these are symlinks to each other, but node doesn't
+    // know that. So we use realpathSync to resolve `identifier` to a "real" path.
+    // This is wrapped in a try/catch because we also generate path identifiers for
+    // node builtins, etc, and we generate a separate error for that further down.
+    identifier =
+      path.isAbsolute(identifier) && existsSync(identifier)
+        ? realpathSync(identifier)
+        : identifier;
 
     // If we've already seen a module with the same identifier, return it, to
     // handle import cycles


packages/runner-vm/src/linker.ts
```diff
@@ -61,11 +61,26 @@ export class ModuleLinker {
...
+      // situation where process.cwd() and referenced modules are in /private/var/...
+      // and /var/... On macs, these are symlinks to each other, but node doesn't
+      // know that. So we use realpathSync to resolve `identifier` to a "real" path.
+      // This is wrapped in a try/catch because we also generate path identifiers for
+      // node builtins, etc, and we generate a separate error for that further down.
+      identifier = path.isAbsolute(identifier)
+        ? realpathSync(identifier)
+        : identifier;
+    } catch (err) {
+      // silently fail
```
mrbbot, line 81: Ignoring this silently kinda scares me, but it's probably fine. Do you think it's worth expliciting checking the code is what's expected (e.g. `ENOENT`) and `throw`ing otherwise?

mrbbot (commented):
Hey! 👋 Thanks for looking into this. Does this fix `ignore` requiring a `path.relative`d path, or were 2 copies of the same module being loaded into the module cache, or something else?

Do you think you'd be able to [add a test here](https://github.com/cloudflare/miniflare/blob/master/packages/runner-vm/test/linker.spec.ts) too?

packages/runner-vm/src/linker.ts
threepointone: Instead, I added an `existsSync` check and removed the try/catch altogether

threepointone:
> Does this fix ignore requiring a path.relatived path, or were 2 copies of the same module being loaded into the module cache, or something else?

This specifically fixes the `path.relative()d path` error that `ignores` was throwing 

> Do you think you'd be able to [add a test here](https://github.com/cloudflare/miniflare/blob/master/packages/runner-vm/test/linker.spec.ts) too?

So I tried adding a test that imports a symlinked module, but that passes even without my fix. This bug is super specific to the `/var/...`/`/private/var/...` scenario, and I'm unsure how to reproduce it except in wrangler2. I'll be adding tests for `wrangler dev` fairly soon and I can add coverage there, but I'm unsure otherwise how to do it here. Any ideas on how to do so, if it's critical?

threepointone:
Don't merge this in just yet, I'm going to try some more on wrangler's side and see if I can fix it there directly. I'll ping you once I do.

threepointone:
Ok, I have an alternate fix in wrangler directly here https://github.com/cloudflare/wrangler2/pull/561, so I'm going to close this. wrangler's usecase is very odd and you shouldn't need the liability of that code here. That said, if this error pops up for someone in the future, you'll know what the suggested fix is (or feel free to reopen and merge this).

Issue #180, opened 2022-02-19
The client's `Accept-Encoding` header is not passed in headers, but cf object


evanderkoogh:
The 'Accept-Encoding' header is always set to gzip + deflate. The actual client `Accept-Encoding` header is passed through as `request.cf.clientAcceptEncoding`

evanderkoogh:
Hey @mrbbot, I have been trying to find out what the best way is to do this, but it is a bit hard to find a good place to change it. Would be great if you could point me to where this change could be best made.. is it `http.ts` in the `core/src/standards` maybe?

mrbbot:
Hey! 👋 Thanks for raising this and apologies for the delayed response. I would probably add this around here in the `@miniflare/http-server` package:
https://github.com/cloudflare/miniflare/blob/b887877810a745f2a458d20da0a3dbd9f8a58cf4/packages/http-server/src/index.ts#L100-L121

leader22:
Hi, I've created a PR to fix this partially(for `headers` part, not `cf`). 👉🏻 https://github.com/cloudflare/miniflare/pull/213

But I want to confirm a few things about my fixes.

I tested this code below with `wrangler dev` + `curl http://127.0.0.1:8787`,

```js
addEventListener('fetch', (event) => {
  event.respondWith(handleRequest(event.request));
});

/** @param {Request} request */
async function handleRequest({ headers: _headers }) {
  const headers = {};
  for (const [key, value] of _headers) headers[key] = value;

  return new Response(JSON.stringify(headers, null, 2), {
    headers: { 'content-type': 'application/json' }
  });
}
```

printed JSON like this.(some values are tweaked)

```json
{
  "accept": "*/*",
  "accept-encoding": "gzip",
  "cf-connecting-ip": "123d:4a:56e:b78:a901:23ff:ff45:67c8",
  "cf-ew-preview-server": "https://123m45.cfops.net",
  "cf-ipcountry": "JP",
  "cf-ray": "1e23456c7d8901b2",
  "cf-visitor": "{\"scheme\":\"https\"}",
  "connection": "Keep-Alive",
  "host": "foo.bar.workers.dev",
  "user-agent": "curl/7.77.0",
  "x-forwarded-proto": "https",
  "x-real-ip": "123d:4a:56e:b78:a901:23ff:ff45:67c8"
}
```

> The 'Accept-Encoding' header is always set to gzip + deflate.

[...]

PR #213, opened 2022-03-09
Add accept-encoding in request headers

leader22:
Refs #180.

### DIFF
--- a/packages/http-server/src/index.ts
+++ b/packages/http-server/src/index.ts
@@ -103,8 +103,15 @@ export async function convertNodeRequest(
   req.headers["cf-ipcountry"] ??= meta?.cf?.country ?? "US";
   req.headers["cf-ray"] ??= randomHex(16);
   req.headers["cf-visitor"] ??= `{"scheme":"${proto}"}`;
-  req.headers["accept-encoding"] = "gzip";
   req.headers["host"] = url.host;
+
+  const clientAcceptEncoding = req.headers["accept-encoding"];
+  // This should be fixed
+  req.headers["accept-encoding"] = "gzip";
+  // Instead, set actual encoding if exists
+  if (meta?.cf && clientAcceptEncoding) {
+    meta.cf.clientAcceptEncoding = clientAcceptEncoding;
+  }
 
   // Build Headers object from request
   const headers = _headersFromIncomingRequest(req);
--- a/packages/http-server/src/index.ts
+++ b/packages/http-server/src/index.ts
@@ -103,7 +103,6 @@ export async function convertNodeRequest(
   req.headers["cf-ipcountry"] ??= meta?.cf?.country ?? "US";
   req.headers["cf-ray"] ??= randomHex(16);
   req.headers["cf-visitor"] ??= `{"scheme":"${proto}"}`;
-  req.headers["cf-ew-preview-server"] = `https://${randomHex(7)}.cfops.net`;
   req.headers["accept-encoding"] = "gzip";
   req.headers["host"] = url.host;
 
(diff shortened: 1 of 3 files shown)


packages/http-server/src/index.ts
```diff
@@ -103,6 +103,8 @@ export async function convertNodeRequest(
   req.headers["cf-ipcountry"] ??= meta?.cf?.country ?? "US";
   req.headers["cf-ray"] ??= randomHex(16);
   req.headers["cf-visitor"] ??= `{"scheme":"${proto}"}`;
+  req.headers["cf-ew-preview-server"] = `https://${randomHex(7)}.cfops.net`;
```
Electroid, line 106: I don't think we need to add this header, it's not an official API.

leader22:
@Electroid Thanks for the comment! I removed it.

mrbbot (changes requested):
Hey! 👋 Thanks for the PR and apologies for only just getting to this. This looks great. Thanks for adding a test too! 🙂 

For `cf.clientAcceptEncoding` I'd do something like this slightly further down in the function:
```ts
const clientAcceptEncoding = req.headers["accept-encoding"];
req.headers["accept-encoding"] = "gzip";

const request = new Request(url, {
  ...
  cf: {
    ...meta?.cf,
    clientAcceptEncoding
  },
  ...
});
```

Don't worry about the `Connection` header too. This needs to be removed unfortunately otherwise `undici` wouldn't let us construct the `new Request`.

leader22:
Thank you for the comment and review~!

Originally I did not aim to solve cf related issue 😅 , but I tried and updated. Is it something like this?

packages/http-server/src/index.ts
```diff
@@ -105,6 +105,14 @@ export async function convertNodeRequest(
   req.headers["cf-visitor"] ??= `{"scheme":"${proto}"}`;
   req.headers["host"] = url.host;
 
+  const clientAcceptEncoding = req.headers["accept-encoding"];
+  // This should be fixed
+  req.headers["accept-encoding"] = "gzip";
+  // Instead, set actual encoding if exists
+  if (meta?.cf && clientAcceptEncoding) {
+    meta.cf.clientAcceptEncoding = clientAcceptEncoding;
```
mrbbot, line 113: The problem with doing it like this is that you're mutating the passed `meta` object. I think a nicer way of doing this would be to replace the `const request = new Request` below with:
```ts
const request = new Request(url, {
  ...
  cf: {
    ...meta?.cf,
    clientAcceptEncoding
  },
  ...
});
```

leader22: @mrbbot Thanks again! 🙌🏻 

At first I also tried your snippet but it ended up with such TS errors.

```
Type '{ clientAcceptEncoding: string | string[] | undefined; asn?: number | undefined; botManagement?: IncomingRequestCfPropertiesBotManagement | undefined; ... 17 more ...; tlsClientAuth?: IncomingRequestCfPropertiesTLSClientAuth | undefined; }' is not assignable to type 'IncomingRequestCfProperties | RequestInitCfProperties | undefined'.
  Type '{ clientAcceptEncoding: string | string[] | undefined; asn?: number | undefined; botManagement?: IncomingRequestCfPropertiesBotManagement | undefined; ... 17 more ...; tlsClientAuth?: IncomingRequestCfPropertiesTLSClientAuth | undefined; }' is not assignable to type 'IncomingRequestCfProperties'.
    Types of property 'asn' are incompatible.
      Type 'number | undefined' is not assignable to type 'number'.
        Type 'undefined' is not assignable to type 'number'.

125     cf: {
        ~~
```

And I hesitated to use type cast. But is it OK? (If so, there is no reason to mutate `meta` object anymore. 😅 )

```ts
    cf: {
      ...meta?.cf,
      clientAcceptEncoding,
    } as IncomingRequestCfProperties,
```

mrbbot: Yep, I'd say `as IncomingRequestCfProperties` is fine here. I think this is just TypeScript failing to infer things.

mrbbot (approved):

PR #216, opened 2022-03-11
Bind `this` name to the `webcrypto`

szkl:
The changes in https://github.com/nodejs/node/pull/4178 (released in node v17.7.0) makes the [type check](https://github.com/nodejs/node/blob/master/lib/internal/crypto/webcrypto.js#L701-L703) on `this` in the internal `crypto` module fail.


dangilkerson:
I ran into this issue as well. This PR resolved it for me.

cojo:
Same issue here with this fix working as well!

mrbbot (approved):
Hey! 👋 Thanks for the PR! Looks great! ✅

Issue #211, opened 2022-03-08
Support `[text_blobs]`


mrbbot:
Wrangler allows text bindings to be loaded from files. Miniflare should support these too.
```toml
[text_blobs]
TEXT = "./text.txt"
```
 See https://github.com/cloudflare/wrangler2/pull/518 and https://github.com/cloudflare/wrangler/pull/1543.

threepointone:
Oh ha, I thought miniflare already supported this. I didn't even know about wasm_modules/text_blobs until recently 😅

mrbbot:
Added in #228 🎉

threepointone:
Amazing, thank you! I'll update miniflare in w2 as soon as you have a release.

PR #228, opened 2022-03-30
Add support for `text_blobs`

caass:
I have absolutely no idea what I'm doing here 😎

Opening in hope of closing #211


packages/core/src/plugins/bindings.ts
```diff
@@ -45,6 +45,7 @@ export interface BindingsOptions {
   bindings?: Record<string, any>;
   globals?: Record<string, any>;
   wasmBindings?: Record<string, string>;
+  textBlobs?: Record<string, string>;
```
mrbbot, line 48: Can we rename this to `textBlobBindings`?

packages/core/src/plugins/bindings.ts
```diff
@@ -164,6 +165,16 @@ export class BindingsPlugin
   })
   wasmBindings?: Record<string, string>;
 
+  @Option({
+    type: OptionType.OBJECT,
+    typeFormat: "NAME=PATH",
+    name: "text",
+    description: "Text blob to bind",
+    logName: "Text blobs",
```
mrbbot, line 173: Nit: `logName` should be Title Cased
```suggestion
    logName: "Text Blob Bindings",
```

packages/core/src/plugins/bindings.ts
```diff
@@ -246,8 +257,9 @@ export class BindingsPlugin
     // 1) Wrangler [vars]
     // 2) .env Variables
     // 3) WASM Module Bindings
-    // 4) Service Bindings
-    // 5) Custom Bindings
+    // 4) Text blobs
```
mrbbot, line 260: Nit:
```suggestion
    // 4) Text Blobs Bindings
```

packages/core/src/plugins/bindings.ts
```diff
@@ -164,6 +165,16 @@ export class BindingsPlugin
   })
   wasmBindings?: Record<string, string>;
 
+  @Option({
+    type: OptionType.OBJECT,
+    typeFormat: "NAME=PATH",
+    name: "text",
```
mrbbot, line 171: This corresponds to the name of the command line argument, e.g. `--text`.
Probably best to omit it and just use the default `--text-blob` which is a bit clearer.
```suggestion
```

packages/core/src/plugins/bindings.ts
```diff
@@ -281,12 +293,22 @@ export class BindingsPlugin
       }
     }
 
-    // 4) Load service bindings
+    // 4) Load text blobs from files
+    if (this.textBlobs) {
+      // eslint-disable-next-line prefer-const
+      for (let [name, textPath] of Object.entries(this.textBlobs)) {
+        textPath = path.resolve(this.ctx.rootPath, textPath);
+        bindings[name] = (await fs.readFile(textPath)).toString();
```
mrbbot, line 301: [`fs.readFile`](https://nodejs.org/api/fs.html#fspromisesreadfilepath-options) accepts an optional options argument. Setting this to a string automatically decodes the buffer using that encoding.
```suggestion
        bindings[name] = await fs.readFile(textPath, "utf8");
```

mrbbot (commented):
Looks pretty bang on to me! 🙂Added a few comments. If you could add some tests here too that would be awesome: https://github.com/cloudflare/miniflare/blob/master/packages/core/test/plugins/bindings.spec.ts.

packages/core/test/plugins/bindings.spec.ts
```diff
@@ -259,13 +275,33 @@ test("BindingsPlugin: setup: loads WebAssembly bindings", async (t) => {
   result = await plugin.setup();
   t.not(result.bindings?.ADD, undefined);
 });
+
+test("BindingsPlugin: setup: loads text blob bindings", async (t) => {
+  let plugin = new BindingsPlugin(ctx, {
+    textBlobBindings: { LOREM_IPSUM: loremIpsumPath },
+  });
+  let result = await plugin.setup();
+  t.not(result.bindings?.LOREM_IPSUM, undefined);
```
mrbbot, line 284: Nit: this assertion is redundant, since you're checking the value is a fixed string below.

packages/core/test/plugins/bindings.spec.ts
```diff
@@ -259,13 +275,33 @@ test("BindingsPlugin: setup: loads WebAssembly bindings", async (t) => {
...
+  t.is(result.bindings?.LOREM_IPSUM, loremIpsum);
+
+  // Check resolves text blob bindings path relative to rootPath
+  plugin = new BindingsPlugin(
+    { log, compat, rootPath: path.dirname(loremIpsumPath) },
+    { textBlobBindings: { LOREM_IPSUM: loremIpsumPath } }
```
mrbbot, line 290: `loremIpsumPath` is already an absolute path so this won't be resolved relative to `rootPath`.
```suggestion
    { textBlobBindings: { LOREM_IPSUM: "lorem-ipsum.txt" } }
```

packages/core/test/plugins/bindings.spec.ts
```diff
@@ -259,13 +275,33 @@ test("BindingsPlugin: setup: loads WebAssembly bindings", async (t) => {
...
+  );
+  result = await plugin.setup();
+  t.not(result.bindings?.LOREM_IPSUM, undefined);
```
mrbbot, line 293: Nit: same as above

packages/core/test/plugins/bindings.spec.ts
```diff
@@ -287,6 +323,9 @@ test("BindingsPlugin: setup: loads bindings from all sources", async (t) => {
       B: addModulePath,
       C: addModulePath,
     },
+    textBlobBindings: {
+      D: loremIpsumPath,
+    },
```
mrbbot, line 326: The idea with this test is that the lowest priority binding defines the most bindings, that are then overriden by subsequent higher priority bindings. For `textBlobBindings`, you'd want to define `A`, `B`, `C` as `loremIpsumPath`, for `wasmBindings` you'd define `A`, `B`, `C`, `D`, for `envPath` you'd define `A=env\nB=env\nC=env\nD=env\nE=env` and then for wrangler bindings you'd define `A`, `B`, `C`, `D`, `E`, `F`.

This is probably a little excessive, but it's nice to test every combination 🙂

mrbbot (approved):

PR #257, opened 2022-05-09
Feature - Durable Objects Alarms

CraigglesO:
TODO:

- [x] namespace test cases
- [x] plugin test case(s)
- [x] storage test cases
- [x] improve readme to show alarm code


lucaslago:
Hey, @CraigglesO are you still working on this PR?

CraigglesO:
@lucaslago - Yes sir. I'm setting aside this weekend. Sorry for the delay, I really want this feature too, but I got side tracked with a lot of other problems.

packages/durable-objects/src/namespace.ts
```diff
@@ -71,6 +78,12 @@ export class DurableObjectState {
     readonly storage: DurableObjectStorage
   ) {}
 
+  injectDurableObject(durableObject: DurableObject): void {
+    this[kInstance] = durableObject;
+    // we need to throw an error on "setAlarm" if the "alarm" method does not exist
+    if (!durableObject.alarm) this.storage.alarmExists = false;
```
mrbbot, line 84: I'm not sure about this. Could you technically define the `alarm` handler dynamically after the constructor executes?
```js
export class WeirdObject {
  async fetch() {
    this.alarm = () => { ... };
  }
}
```
Maybe add a reference to the Durable Object instance to the alarm bridge somehow in the `DurableObjectPlugin#getObject()` method, then do the check at `setAlarm`/`getAlarm` time?

packages/durable-objects/src/plugin.ts
```diff
@@ -82,6 +89,16 @@ export class DurableObjectsPlugin
...
   durableObjectsPersist?: boolean | string;
+
+  @Option({
+    type: OptionType.BOOLEAN,
+    name: "do-ignore-alarms",
+    description: "Durable Objects will not monitor or trigger alarms.",
+    logName: "Durable Object Alarms",
+    fromWrangler: ({ miniflare }) => miniflare?.durable_objects_ignore_alarms,
+  })
+  ignoreAlarms?: boolean;
```
mrbbot, line 100: I'd do something like the `--(no-)cache`option here instead for consistency, maybe `--(no-)do-alarms`:
https://github.com/cloudflare/miniflare/blob/981133dda005ccb6533206c8ed826dea7eefe8a3/packages/cache/src/plugin.ts#L99-L106
Note the `(no-)` stuff is automatically added when setting `negatable: true`.

packages/durable-objects/src/namespace.ts
```diff
@@ -71,6 +78,12 @@ export class DurableObjectState {
     readonly storage: DurableObjectStorage
   ) {}
 
+  injectDurableObject(durableObject: DurableObject): void {
```
mrbbot, line 81: If we're keeping this method (see the below comment), it needs to be behind a symbol like `[kAlarm]`/`[kFetch]`. Otherwise, end users could call it in their code:
```js
export class BadObjectt {
  constructor(state) {
    state.injectDurableObject({
      fetch() { return new Response("completely different object") }
    });
  }
}
```

packages/durable-objects/src/plugin.ts
```diff
@@ -165,17 +187,41 @@ export class DurableObjectsPlugin
...
+        const ns = this.getNamespace(storage, objectName);
+        const stub = ns.get(new DurableObjectId(objectName, hexId));
+        // build the controller and context
+        const controller = new ScheduledController(scheduledTime, "alarm");
+        const event = new ScheduledEvent("scheduled", {
+          scheduledTime,
+          cron: "alarm",
+        });
+        const ctx = new ExecutionContext(event);
+        // execute the alarm
```
mrbbot, line 219: You'll want to use the `getObject` method here to access the `DurableObjectState` directly, then call `[kAlarm]` on that. You'll need to wrap this with a `RequestContext#runWith` though, see `fetch` in `DurableObjectStub`. (otherwise things like `fetch`, `crypto.getRandomValues` won't work in alarm handlers).

packages/durable-objects/src/namespace.ts
```diff
@@ -166,6 +202,14 @@ export class DurableObjectStub {
...
     return res;
   }
+
+  async alarm(
+    controller: ScheduledController,
+    ctx: ExecutionContext
+  ): Promise<void> {
+    const state = await this.#factory(this.id);
+    await state[kAlarm](controller, ctx);
+  }
```
mrbbot, line 212: `DurableObjectStub`'s shouldn't expose an `alarm` method. You aren't able to do this for example:
```js
export default {
  fetch(request, env) {
    const id = env.TEST_OBJECT.newUniqueId();
    const stub = env.TEST_OBJECT.get(id);
    await stub.alarm();
  }
}
```
You'll want to call the `[kAlarm]` method on `DurableObjectState` directly. See comments below.

packages/durable-objects/src/shadow.ts
```diff
@@ -162,4 +163,16 @@ export class ShadowStorage extends Storage {
     }
     return { keys, cursor: "" /* unsupported */ };
   }
+
+  async getAlarm(): Promise<number | null> {
+    return this.alarm;
+  }
+
+  async setAlarm(scheduledTime: number): Promise<void> {
```
mrbbot, line 171: This function should accept `Date` too: https://github.com/cloudflare/workers-types/blob/3012f263fb1239825e5f0061b267c8650d01b717/index.d.ts#L487-L490.

packages/durable-objects/src/shadow.ts
```diff
@@ -162,4 +163,16 @@ export class ShadowStorage extends Storage {
...
+    this.alarm = scheduledTime;
```
mrbbot, line 172: Should this do any validation on the `scheduledTime`? e.g. checking if positive integer? What does the real runtime do? I note below you clamp this to `0`. This also means you're safe to use `-1` to signal deletion as below.

packages/durable-objects/src/storage.ts
```diff
@@ -403,9 +451,13 @@ export class DurableObjectStorage implements DurableObjectOperator {
   // be added in the future, but it seemed redundant since most users will be
   // using in-memory storage anyways, and the file system isn't *too* slow.
   readonly #shadow: ShadowStorage;
+  readonly #alarmBridge?: DurableObjectAlarmBridge;
+  // Let's storage know if the parent instance includes an alarm method or not
+  alarmExists = true;
```
mrbbot, line 456: See above comments for alternatives, but if this is staying, it needs to be behind a symbol, so it's not accessible to end-users like so:
```js
export class BadObject {
  constructor(state) {
    state.storage.alarmExists = true;
  }
  // ...but alarm doesn't exist
}
```

packages/durable-objects/src/storage.ts
```diff
@@ -425,7 +477,7 @@ export class DurableObjectStorage implements DurableObjectOperator {
   }
 
   async #txnValidateWrite(txn: DurableObjectTransaction): Promise<boolean> {
-    // This function returns false iff the transaction should be retried
+    // This function returns false if the transaction should be retried
```
mrbbot, line 480: `iff` if shorthand for if and only if.
```suggestion
    // This function returns false iff the transaction should be retried
```

packages/durable-objects/src/storage.ts
```diff
@@ -685,4 +745,53 @@ export class DurableObjectStorage implements DurableObjectOperator {
...
+      options?.allowConcurrency
+    );
+  }
+
+  // setAlarm accepts integer milliseconds since epoch or a js date
+  async setAlarm(
+    scheduledTime: DurableObjectScheduledAlarm,
+    options?: DurableObjectSetAlarmOptions
+  ): Promise<void> {
+    if (!this.alarmExists) throw new Error("Alarm method not set.");
```
mrbbot, line 765: Is this the same error message the real runtime uses?

packages/durable-objects/src/shadow.ts
```diff
@@ -14,6 +14,7 @@ const collator = new Intl.Collator();
 export class ShadowStorage extends Storage {
   readonly readSet?: Set<string>;
   readonly copies = new Map<string, StoredValue | undefined>();
+  alarm: number | null = null;
```
mrbbot, line 17: Shadow storage needs to be an *overlay* over the underlying storage. At the moment, if an alarm is already set (and persistence is on), then you try to `getAlarm` in a transaction, you'll read `null`. Similarly, setting an alarm outside a transaction, then `getAlarm` in a transaction will return `null`:
```js
await this.storage.setAlarm(Date.now() + 30_000);
await this.storage.transaction(async (txn) => {
  console.log(await txn.getAlarm()); // null
});
```
What you need to do is have `alarm` initially `undefined`.
In `ShadowStorage#setAlarm` update it to the new uncomitted value.
Then in `ShadowStorage#getAlarm`, if it's defined return that value, otherwise return the value from the underlying `inner` storage.

packages/durable-objects/src/shadow.ts
```diff
@@ -162,4 +163,16 @@ export class ShadowStorage extends Storage {
     }
     return { keys, cursor: "" /* unsupported */ };
   }
+
+  async getAlarm(): Promise<number | null> {
+    return this.alarm;
```
mrbbot, line 168: I think this needs to add `__MINIFLARE_ALARM__` to the `readSet`?

packages/durable-objects/test/storage.spec.ts
```diff
@@ -984,11 +1237,15 @@ test("transaction: allows delivery of events inside transaction closure even tho
 test("hides implementation details", (t) => {
   const { storage } = t.context;
   t.deepEqual(getObjectProperties(storage), [
+    "alarmExists",
```
mrbbot, line 1240: This is an implementation detail. See above comments.

packages/shared/src/storage.ts
```diff
@@ -77,6 +77,9 @@ export abstract class Storage {
     options: StorageListOptions,
     skipMetadata: true
   ): Awaitable<StorageListResult<StoredKey>>;
+  abstract getAlarm(): Awaitable<number | null>;
+  abstract setAlarm(scheduledTime: number): Awaitable<void>;
+  abstract deleteAlarm(): Awaitable<void>;
```
mrbbot, line 82: You shouldn't need to add these methods to `Storage`. Can't you use regular `get`/`set`/`delete` with your magic key?

mrbbot (changes requested):
Hey! 👋 Thanks so much again for working on this. I've added quite a few comments. I hadn't realised how much was involved in this. 😅 Let me know if you have any questions.

packages/durable-objects/src/storage.ts
CraigglesO: Recent update:
https://community.cloudflare.com/t/2022-5-19-workers-runtime-release-notes/385028

packages/durable-objects/src/shadow.ts
CraigglesO: This clears a lot of things up. I will change to using get with '__MINIFLARE_ALARM__' key.

packages/durable-objects/test/storage.spec.ts
CraigglesO: marking twice
https://community.cloudflare.com/t/2022-5-19-workers-runtime-release-notes/385028
Would you prefer I did it in a different way?

packages/shared/src/storage.ts
CraigglesO: sure!

packages/durable-objects/test/storage.spec.ts
CraigglesO: After reading all comments I'm going to keep the implementation but use a `Symbol` as advised.

packages/durable-objects/src/plugin.ts
```diff
@@ -224,7 +268,8 @@ export class DurableObjectsPlugin
     this.#contextResolve();
   }
 
-  dispose(): void {
+  dispose(disposeAlarms = false): void {
+    if (disposeAlarms) this.#alarmStore.dispose();
```
mrbbot, line 272: I think you always want to `dispose` alarms here. When this plugin's options change, the `dispose()` method will be called on the existing instance of `DurableObjectPlugin`, a new instance will be created (setting up a new alarm store), and `setup()` will be called on that new instance.

packages/durable-objects/src/shadow.ts
```diff
@@ -162,4 +164,24 @@ export class ShadowStorage extends Storage {
...
+
+  setAlarm(scheduledTime: number | Date): Promise<void> {
+    if (typeof scheduledTime !== "number")
+      scheduledTime = scheduledTime.getTime();
+    this.alarm = scheduledTime;
+    return Promise.resolve();
+  }
+
+  deleteAlarm(): Promise<void> {
+    this.alarm = undefined;
```
mrbbot, line 184: I don't think this will work. The default value of `alarm` is `undefined`, so won't this cause the alarm to be deleted if the user doesn't call `setAlarm` in the transaction?

Later on, you check for `alarm === -1` to signal delete. I think you want to do `this.alarm = -1` here.

packages/durable-objects/src/shadow.ts
```diff
@@ -162,4 +164,24 @@ export class ShadowStorage extends Storage {
     }
     return { keys, cursor: "" /* unsupported */ };
   }
+
+  async getAlarm(): Promise<number | null> {
+    this.readSet?.add(ALARM_KEY);
```
mrbbot, line 169: This function doesn't have a `[kAlarmExists]` check.

packages/durable-objects/src/storage.ts
```diff
@@ -685,4 +751,51 @@ export class DurableObjectStorage implements DurableObjectOperator {
...
+      options?.allowConcurrency
+    );
+  }
+
+  // setAlarm accepts integer milliseconds since epoch or a js date
+  async setAlarm(
+    scheduledTime: number | Date,
+    options?: DurableObjectSetAlarmOptions
+  ): Promise<void> {
+    if (!this[kAlarmExists])
```
mrbbot, line 771: Could you add braces to this `if` statement?

packages/durable-objects/src/storage.ts
```diff
@@ -685,4 +751,51 @@ export class DurableObjectStorage implements DurableObjectOperator {
...
+      throw new Error(
+        "Your Durable Object class must have an alarm() handler in order to call setAlarm()"
+      );
+    return runWithGatesClosed(async () => {
+      await this.#mutex.runWithWrite(async () => {
+        await this.#alarmBridge?.setAlarm(scheduledTime);
```
mrbbot, line 777: You need to move this line to `#flush`, so it gets called when a transaction commits too.
```suggestion
```

packages/durable-objects/src/storage.ts
```diff
@@ -685,4 +751,51 @@ export class DurableObjectStorage implements DurableObjectOperator {
...
+      // Promise.resolve() allows other setAlarms/deleteAlarms (coalescing) before flush
+      await Promise.resolve();
+      return this.#mutex.runWithWrite(this.#flush);
+    }, options);
+  }
+
+  async deleteAlarm(options?: DurableObjectSetAlarmOptions): Promise<void> {
+    return runWithGatesClosed(async () => {
+      await this.#mutex.runWithWrite(async () => {
+        await this.#alarmBridge?.deleteAlarm();
```
mrbbot, line 791: Again, you need to move this line to `#flush` so it gets called when a transaction commits too.
```suggestion
```

packages/durable-objects/src/storage.ts
```diff
@@ -500,6 +555,17 @@ export class DurableObjectStorage implements DurableObjectOperator {
...
+    // flush alarm should it exist
+    if (typeof this.#shadow.alarm === "number") {
+      if (this.#shadow.alarm === -1) await this.#inner.delete(ALARM_KEY);
+      else
+        await this.#inner?.put(ALARM_KEY, {
+          metadata: { scheduledTime: this.#shadow.alarm },
+          value: new Uint8Array(),
+        });
+      this.#shadow.alarm = undefined;
+    }
```
mrbbot, line 567: I think you need to call alarm bridge methods here, so the alarm is stored in the main alarm store and scheduled. See comments below about moving calls.

packages/durable-objects/src/storage.ts
```diff
@@ -500,6 +555,17 @@ export class DurableObjectStorage implements DurableObjectOperator {
...
     // Must be called with #mutex's write lock held
 
+    // flush alarm should it exist
+    if (typeof this.#shadow.alarm === "number") {
+      if (this.#shadow.alarm === -1) await this.#inner.delete(ALARM_KEY);
+      else
+        await this.#inner?.put(ALARM_KEY, {
+          metadata: { scheduledTime: this.#shadow.alarm },
+          value: new Uint8Array(),
+        });
```
mrbbot, line 565: Could you add braces to this `if`/`else`? That 4 line `else` branch is a little too big to not be wrapped.

packages/storage-redis/src/index.ts
```diff
@@ -30,6 +30,7 @@ export class RedisStorage extends Storage {
   // returns one, and we can fetch metadata separately for listing
   readonly #key = (key: string): string => `${this.namespace}:value:${key}`;
   readonly #metaKey = (key: string): string => `${this.namespace}:meta:${key}`;
+  readonly #alarmKey = (): string => `${this.namespace}:alarm`;
```
mrbbot, line 33:
```suggestion
```

packages/shared-test/src/storage/recorder.ts
```diff
@@ -16,7 +16,10 @@ export type StorageEvent =
   | { type: "hasMany"; keys: string[] }
   | { type: "getMany"; keys: string[] }
   | { type: "putMany"; keys: string[] }
-  | { type: "deleteMany"; keys: string[] };
+  | { type: "deleteMany"; keys: string[] }
+  | { type: "getAlarm" }
+  | { type: "setAlarm" }
+  | { type: "deleteAlarm" };
```
mrbbot, line 22:
```suggestion
  | { type: "deleteMany"; keys: string[] };
```

packages/durable-objects/test/plugin.spec.ts
```diff
@@ -76,23 +76,27 @@ test("DurableObjectsPlugin: parses options from wrangler config", (t) => {
...
     "Durable Objects Persistence: true",
+    "Durable Object Alarms: true",
   ]);
 });
 
 test("DurableObjectsPlugin: getObject: waits for constructors and bindings", async (t) => {
   const factory = new MemoryStorageFactory();
   const plugin = new DurableObjectsPlugin(ctx, {
     durableObjects: { TEST: "TestObject" },
+    ignoreAlarms: true,
```
mrbbot, line 99: Why do you need to `ignoreAlarms` in all these tests? Surely it shouldn't matter, and would be good to check everything works correctly together?

mrbbot (changes requested):

packages/durable-objects/src/plugin.ts
CraigglesO: Ah, I was worried based upon test cases this does not happen. TY for clarification

docs/src/content/get-started/cli.md
```diff
@@ -279,6 +279,8 @@ Durable Objects Options:
  -o, --do                Durable Object to bind       [array:NAME=CLASS[@MOUNT]]
      --do-persist        Persist Durable Object data (to        [boolean/string]
                          optional path)
+    --(no-)do-alarms     Durable Objects will not monitor       [boolean]
```
mrbbot, line 282: This description needs updating to match the `@Option` annotation.

packages/durable-objects/src/plugin.ts
```diff
@@ -82,6 +86,17 @@ export class DurableObjectsPlugin
...
   })
   durableObjectsPersist?: boolean | string;
+
+  @Option({
+    type: OptionType.BOOLEAN,
+    name: "do-alarms",
+    description: "Enable Durable Object alarms (enabled by default)",
+    negatable: true,
+    logName: "Durable Object Alarms",
+    fromWrangler: ({ miniflare }) => miniflare?.do_alarms ?? true,
```
mrbbot, line 96: You should handle the default value once later on.
```suggestion
    fromWrangler: ({ miniflare }) => miniflare?.do_alarms,
```

packages/durable-objects/src/plugin.ts
```diff
@@ -165,17 +186,38 @@ export class DurableObjectsPlugin
...
     }
+    await this.#setupAlarms(storageFactory);
     return {
       bindings,
       requiresModuleExports: this.#processedObjects.length > 0,
     };
   }
 
+  async #setupAlarms(storage: StorageFactory): Promise<void> {
+    if (!this.durableObjectAlarms) return;
```
mrbbot, line 202: When `durableObjectAlarms` is `undefined` (default value when using the API), we still want to set up alarms.
```suggestion
    if (this.durableObjectAlarms === false) return;
```

packages/miniflare/README.md
```diff
@@ -131,6 +131,7 @@ KV Options:
 Durable Objects Options:
  -o, --do                Durable Object to bind                             [array:NAME=CLASS[@MOUNT]]
      --do-persist        Persist Durable Object data (to optional path)               [boolean/string]
+     --(no-)do-alarms    Durable Objects will not monitor or trigger alarms.          [boolean]
```
mrbbot, line 134: This description needs updating to match the `@Option` annotation.

packages/durable-objects/src/shadow.ts
```diff
@@ -162,4 +164,24 @@ export class ShadowStorage extends Storage {
...
+    if (typeof this.alarm === "number") return this.alarm;
```
mrbbot, line 170: If `this.alarm` is `null` (deleted in transaction), wouldn't you want to return `null` then too?

mrbbot (changes requested):
Getting really close, just a few minor things now.  😃

packages/durable-objects/src/storage.ts
```diff
@@ -500,6 +568,19 @@ export class DurableObjectStorage implements DurableObjectOperator {
   #flush = async (): Promise<void> => {
     // Must be called with #mutex's write lock held
 
+    // flush alarm should it exist
+    if (typeof this.#shadow.alarm === "number") {
+      await this.#inner?.put(ALARM_KEY, {
```
mrbbot, line 573:
```suggestion
      await this.#inner.put(ALARM_KEY, {
```

petebacondarwin:
🎉

Issue #256, opened 2022-05-07
TypeError: Value of "this" must be of type SubtleCrypto


lmcarreiro:
When I run this simple code, it throws an error when running from miniflare. The same code works fine running from Cloudflare (or `wrangler dev` without the `--local` flag).

Code:

```js
export default {
  async fetch(request, env, ctx) {
    const secret = new TextEncoder().encode('test')
    const key = await crypto.subtle.importKey('raw', secret, { name: 'HMAC', hash: { name: 'SHA-512' } }, false, [
      'sign',
      'verify',
    ])

    return new Response(JSON.stringify(key), { status: 200 })
  },
}
```

Error:

```
TypeError: Value of "this" must be of type SubtleCrypto
    at new NodeError (node:internal/errors:377:5)
    at Proxy.importKey (node:internal/crypto/webcrypto:517:30)
    at Object.fetch (/private/Users/leonardo/worker-test/src/index.ts:4:37)
    at /Users/leonardo/worker-test/node_modules/@miniflare/core/src/standards/event.ts:331:19
    at /Users/leonardo/worker-test/node_modules/@miniflare/shared/src/event.ts:29:9
    at ServiceWorkerGlobalScope.<anonymous> (/Users/leonardo/worker-test/node_modules/@miniflare/shared/src/event.ts:77:9)
    at ServiceWorkerGlobalScope.[nodejs.internal.kHybridDispatch] (node:internal/event_target:639:20)
    at ServiceWorkerGlobalScope.dispatchEvent (node:internal/event_target:581:26)
    at ServiceWorkerGlobalScope.dispatchEvent (/Users/leonardo/worker-test/node_modules/@miniflare/shared/src/event.ts:63:18)
    at ServiceWorkerGlobalScope.dispatchEvent (/Users/leonardo/worker-test/node_modules/@miniflare/shared/src/event.ts:87:26)
```

awwong1:
I observed this as well, downgrading to Node 16 locally appears to resolve this error.
Appears related to https://github.com/cloudflare/miniflare/pull/216

mrbbot:
Hey! 👋 Thanks for raising this. Will take a look at that PR soon. 👍

awwong1:
> Hey! 👋 Thanks for raising this. Will take a look at that PR soon. 👍

Thanks for the review and for merging it in! Grateful to have a contribution into this repo. I'll leave this issue open until a new release is deployed.

caspervonb:
[...]

PR #259, opened 2022-05-13
bind `this` within proxy for webcrypto & subtle

awwong1:
Extends PR https://github.com/cloudflare/miniflare/pull/216
Addresses Issue https://github.com/cloudflare/miniflare/issues/256

### DIFF
--- a/packages/core/src/standards/crypto.ts
+++ b/packages/core/src/standards/crypto.ts
@@ -86,12 +86,11 @@ export function createCrypto(blockGlobalRandom = false): typeof webcrypto {
 
   return new Proxy(webcrypto, {
     get(target, propertyKey, receiver) {
-      let result = undefined;
-      if (propertyKey === "getRandomValues") result = getRandomValues;
-      else if (propertyKey === "subtle") result = subtle;
-      else if (propertyKey === "DigestStream") result = DigestStream;
-      else result = Reflect.get(target, propertyKey, receiver);
+      if (propertyKey === "getRandomValues") return getRandomValues;
+      if (propertyKey === "subtle") return subtle;
+      if (propertyKey === "DigestStream") return DigestStream;
 
+      let result = Reflect.get(target, propertyKey, receiver);
       if (typeof result === "function") result = result.bind(webcrypto);
       return result;
     },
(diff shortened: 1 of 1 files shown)


packages/core/src/standards/crypto.ts
```diff
@@ -75,18 +75,26 @@ export function createCrypto(blockGlobalRandom = false): typeof webcrypto {
...
-      if (propertyKey === "digest") return digest;
-      if (propertyKey === "generateKey") return generateKey;
-      return Reflect.get(target, propertyKey, receiver);
+      let result = undefined;
+      if (propertyKey === "digest") result = digest;
+      else if (propertyKey === "generateKey") result = generateKey;
+      else result = Reflect.get(target, propertyKey, receiver);
+
+      if (typeof result === "function") result = result.bind(webcrypto.subtle);
+      return result;
```
mrbbot, line 84: `digest` doesn't need to be bound to `webcrypto.subtle`, and `generateKey` is already bound above.

```suggestion
      if (propertyKey === "digest") return digest;
      if (propertyKey === "generateKey") return generateKey;

      let result = Reflect.get(target, propertyKey, receiver);
      if (typeof result === "function") result = result.bind(webcrypto.subtle);
      return result;
```

packages/core/src/standards/crypto.ts
```diff
@@ -75,18 +75,26 @@ export function createCrypto(blockGlobalRandom = false): typeof webcrypto {
...
-      if (propertyKey === "DigestStream") return DigestStream;
-      return Reflect.get(target, propertyKey, receiver);
+      let result = undefined;
+      if (propertyKey === "getRandomValues") result = getRandomValues;
+      else if (propertyKey === "subtle") result = subtle;
+      else if (propertyKey === "DigestStream") result = DigestStream;
+      else result = Reflect.get(target, propertyKey, receiver);
+
+      if (typeof result === "function") result = result.bind(webcrypto);
+      return result;
```
mrbbot, line 97: `getRandomValues` is already bound above, and `DigestStream` shouldn't be bound.
I'm not actually sure the `Reflect.get` branch will ever be reached for valid `propertyKey`s, but no harm in keeping this in.
```suggestion
      if (propertyKey === "getRandomValues") return getRandomValues;
      if (propertyKey === "subtle") return subtle;
      if (propertyKey === "DigestStream") return DigestStream;

      let result = Reflect.get(target, propertyKey, receiver);
      if (typeof result === "function") result = result.bind(webcrypto);
      return result;
```

mrbbot (changes requested):
Hey! 👋 Thanks for the PR! 🙂 I've made a few suggestions. There are some things that are already bound, and some things that shouldn't be bound.

mrbbot (approved):
Awesome! Thank you! ✅ Will just wait for checks to pass.

PR #283, opened 2022-06-14
Feature/get range

CraigglesO:
First step towards R2.

This is a proposal to avoid a behavior mismatch, since large files and `storage-file` package with a range request would be performantly slow.

In order to properly handle range requests for large files, I came up with this idea to be as minimally invasive as possible.

With this method, you can check if `getRange` is a function in storage, and if not just do a normal full `get` + `slice` on the data. Otherwise, you can specify the start and length.

Note this method also ignores `expiration` as it specifically exists for R2.

### DIFF
--- a/packages/shared/src/storage.ts
+++ b/packages/shared/src/storage.ts
@@ -79,7 +79,7 @@ export abstract class Storage {
   ): Awaitable<StorageListResult<StoredKey>>;
 
   // storage-file specific methods
-  abstract getRange?<Meta = unknown>(
+  async getRange?<Meta = unknown>(
     key: string,
     start: number,
     length: number
--- a/packages/shared/src/storage.ts
+++ b/packages/shared/src/storage.ts
@@ -78,6 +78,13 @@ export abstract class Storage {
     skipMetadata: true
   ): Awaitable<StorageListResult<StoredKey>>;
 
+  // storage-file specific methods
+  abstract getRange?<Meta = unknown>(
+    key: string,
+    start: number,
+    length: number
+  ): Promise<StoredValueMeta<Meta> | undefined>;
+
   // Batch functions, default implementations may be overridden to optimise
 
   async hasMany(keys: string[]): Promise<number> {
--- a/packages/storage-file/src/helpers.ts
+++ b/packages/storage-file/src/helpers.ts
@@ -20,6 +20,25 @@ export function readFile(
   return onNotFound(fs.readFile(filePath, decode && "utf8"), undefined);
 }
 
+export async function read(
+  filePath: string,
+  start: number,
+  length: number
+): Promise<Buffer | undefined> {
+  let fd = null;
+  let res: Buffer;
+  try {
+    fd = await fs.open(filePath, "r");
+    res = Buffer.alloc(length);
+    await fd.read(res, 0, length, start);
+  } catch (e) {
+    throw e;
+  } finally {
+    await fd?.close();
+  }
+  return res;
+}
+
 export async function writeFile(
   filePath: string,
   data: Uint8Array | string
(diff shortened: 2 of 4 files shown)


CraigglesO:
Interestingly enough from my tests:
```js
      // We'll get this error if we try to get a namespaced key, where the
      // namespace itself is also a key (e.g. trying to get "key/sub-key" where
      // "key" is also a key). In this case, "key/sub-key" doesn't exist.
      if (e.code === "ENOTDIR") return;
```

does not work on windows. windows throws error code `ENOENT` even for directory failures.

packages/shared/src/storage.ts
```diff
@@ -78,6 +78,13 @@ export abstract class Storage {
     skipMetadata: true
   ): Awaitable<StorageListResult<StoredKey>>;
 
+  // storage-file specific method
+  async getRange?<Meta = unknown>(
```
mrbbot, line 82: Could we rename this to `getRangeMaybeExpired` to make it explicit to implementers/callers that this method may return expired data?

packages/shared/src/storage.ts
```diff
@@ -78,6 +78,13 @@ export abstract class Storage {
     skipMetadata: true
   ): Awaitable<StorageListResult<StoredKey>>;
 
+  // storage-file specific method
```
mrbbot, line 81: Nit: as this is in the abstract class, we shouldn't really be referring to specific implementations.
```suggestion
  // Implementations (e.g. storage-file) may override this for efficient range requests
```

packages/storage-file/src/helpers.ts
```diff
@@ -20,6 +20,25 @@ export function readFile(
...
+  length: number
+): Promise<Buffer | undefined> {
+  let fd = null;
+  let res: Buffer;
+  try {
+    fd = await fs.open(filePath, "r");
+    res = Buffer.alloc(length);
+    await fd.read(res, 0, length, start);
+  } catch (e) {
+    throw e;
```
mrbbot, line 35: To match the behaviour of `readFile` above, this should return `undefined` on `ENOENT`.
```suggestion
  } catch (e) {
    if (e.code === "ENOENT") return undefined;
    throw e;
```

packages/storage-file/src/index.ts
```diff
@@ -85,6 +85,32 @@ export class FileStorage extends LocalStorage {
...
+    if (!filePath) return;
+
+    try {
+      const value = await read(filePath, start, length);
+
+      if (value === undefined) return;
+      const meta = await this.meta<Meta>(filePath);
+      return {
+        value: viewToArray(value),
+        metadata: meta.metadata,
```
mrbbot, line 103: Even though R2 doesn't care about expiration, I think it's still useful to return it here for consistency with `getMaybeExpired` above. We can assert it's `undefined` in the R2 implementation.
```suggestion
        expiration: meta.expiration,
        metadata: meta.metadata,
```

packages/storage-file/test/index.spec.ts
```diff
@@ -67,6 +67,19 @@ test("FileStorage: list: returns original keys if sanitised", async (t) => {
...
+
+  const getFront = await storage.getRange?.("key", 0, 3);
+  t.is(utf8Decode(getFront?.value), "123");
+
+  const getBack = await storage.getRange?.("key", 6, 3);
+  t.is(utf8Decode(getBack?.value), "789");
+
+  const getMiddle = await storage.getRange?.("key", 3, 3);
+  t.is(utf8Decode(getMiddle?.value), "456");
+});
```
mrbbot, line 83: What happens if the range is out of bounds?

mrbbot (approved):
Awesome! Looks great! ✅

Issue #276, opened 2022-06-09
🪣 Support R2


mrbbot:
https://developers.cloudflare.com/r2/

PR #289, opened 2022-06-20
r2 implementation

CraigglesO:
Aims to close https://github.com/cloudflare/miniflare/issues/276

TODO:

- [x] Bucket testing.
- [x] Check that all error responses match.
- [x] Check that all implementations are the same.
- [x] Get windows tests working

PART 2:

- [x] Range object metadata
- [x] R2ObjectBody -> body readableStream not accepted by Response.
- [x] add startAfter option for list
- [x] tests for startAfter
- [x] tests for delimitedPrefixes
- [x] tests for ranges


vlovich:
Since I've sent you the ListObjects implementation, we've got a report of another issue & I have a couple of bug fixes pending. We should probably make sure to update the Miniflare implementation to match.

Also, we've added `range` in the returned object on `get` if a range is specified (the actual range encountered is returned). Range logic is *very* tricky and probably needs to be compared against what R2 is actually doing to make sure things are right (I haven't checked that part).

I'm also in the process of plumbing through startAfter as a list option - probably should be rolled into here (but again - double-check against the latest version of list in R2).

It's probably a good idea for whoever is reviewing this on the cloudflare side (not me - I don't have the cycles aside from providing on-demand consultation) to look at the real R2 implementation and compare/contrast:
* ListObjects implementation
* Conditional handling (conditional puts + general conditionals)
* Range handling

vlovich:
Also on-deck for being implemented (& might be worth prototyping the bindings for it in Miniflare):

1. Multipart
2. Copy object
3. Move object

I have some of the shape of the APIs sketched out already (multipart in particular) so please coordinate with me on Discord whenever someone takes that on.

packages/r2/src/r2Object.ts
```diff
@@ -0,0 +1,323 @@
...
+    return new Uint8Array(0);
+  }
+
+  async #getArrayBuffer(): Promise<ArrayBuffer> {
+    const body = await this.#getBody();
+    return body.buffer.slice(
+      body.byteOffset,
+      body.byteLength + body.byteOffset
+    );
+  }
```
mrbbot, line 302: Maybe remove these methods, and change the implementation of `R2ObjectBody#arrayBuffer` to call `arrayBuffer` from `stream/consumers`, after the `bodyUsed` check. Then `R2ObjectBody#text` and `R2ObjectBody#blob` can call `R2ObjectBody#arrayBuffer` instead.

packages/r2/test/bucket.spec.ts
```diff
@@ -0,0 +1,1847 @@
...
+const storageFactory = new FileStorageFactory();
+for (const macro of storageMacros) {
+  test(macro, storageFactory);
+}
+
+test.beforeEach(async (t) => {
+  const storage = await storageFactory.factory(t, {});
+  const r2 = new R2Bucket(storage);
+  t.context = { storage, r2 };
+});
```
mrbbot, line 99: I don't think you need this full setup. The `FileStorageFactory` is meant for testing `FileStorage`. In particular, you definitely don't want to do `for (const macro of storageMacros) {` as this will re-run all storage tests that are tested elsewhere. I'd do something like this:

```suggestion
test.beforeEach(async (t) => {
  const tmp = await useTmp(t);
  const storage = new FileStorage(tmp, true, testClock);
  const r2 = new R2Bucket(storage);
  t.context = { storage, r2 };
});
```

packages/r2/test/bucket.spec.ts
```diff
@@ -0,0 +1,1847 @@
...
+};
+
+test("with md5 as correct arrayBuffer", putMacro, {
+  key: "text",
+  value: "value",
+  expected: { value: utf8Encode("value") },
+  options: {
+    md5: md5ToBuffer(createHash(utf8Encode("value"))),
+  },
+});
```
mrbbot, line 717: You can do this with `viewToBuffer` from `@miniflare/shared`.
```suggestion
test("with md5 as correct arrayBuffer", putMacro, {
  key: "text",
  value: "value",
  expected: { value: utf8Encode("value") },
  options: {
    md5: viewToBuffer(Buffer.from(createHash(utf8Encode("value")), "hex")),
  },
});
```

packages/r2/test/bucket.spec.ts
```diff
@@ -0,0 +1,1847 @@
...
+    section2key1: "value21",
+  },
+  options: { prefix: "section1", limit: 2 },
+  objects: [
+    [{ key: "section1key1" }, { key: "section1key2" }],
+    [{ key: "section1key3" }],
+  ],
+});
+
+const testEqualityMacro = async (
```
mrbbot, line 1316: As this isn't really an AVA macro, maybe rename it to `assertEquality`?

packages/r2/test/r2Object.spec.ts
```diff
@@ -0,0 +1,508 @@
...
+  } else if (value instanceof Blob) {
+    stored = new Uint8Array(await value.arrayBuffer());
+  } else {
+    throw new TypeError(
+      "Accepts only nulls, strings, Blobs, ArrayBuffers, ArrayBufferViews, and ReadableStreams as values."
+    );
+  }
+
+  return new R2ObjectBody(metadata, stored);
+};
```
mrbbot, line 82: This is very similar to the implementation in `r2Object.ts`, maybe split that out into an exported function from `@miniflare/r2`? If you prefix the function name with `_` and mark it with `/** @internal */`, it will be omitted from public user-facing type definitions.

packages/r2/test/r2Object.spec.ts
```diff
@@ -0,0 +1,508 @@
...
+  t.is(headers.get("content-disposition"), "inline");
+  t.is(headers.get("content-encoding"), "gzip");
+  t.is(headers.get("cache-control"), "public, max-age=31536000");
+  t.is(
+    new Date(headers.get("cache-expiry") ?? 0).toUTCString(),
+    r2ObjectBody.httpMetadata?.cacheExpiry?.toUTCString()
+  );
+});
+
+test("R2Object: R2ObjectBody: Test bodyUsed", async (t) => {
```
mrbbot, line 143: Nit: `Test` shouldn't be capitalised to match the names used by other tests. (also below too)
```suggestion
test("R2Object: R2ObjectBody: test bodyUsed", async (t) => {
```

packages/storage-file/src/helpers.ts
```diff
@@ -24,22 +30,51 @@ export function readFile(
...
+  } catch (e: any) {
+    if (e.code === "ENOENT") return undefined;
+    throw e;
+  } finally {
+    await fd?.close();
+  }
+  return { value: res, offset, length };
+}
+
+export async function readFileSuffix(
```
mrbbot, line 56: Could you merge this method with `readFileRange` by allowing `offset` to be `undefined`, then setting it based on `length` as you do below?

packages/shared/src/storage.ts
```diff
@@ -79,11 +80,18 @@ export abstract class Storage {
...
+  ): Promise<StoredMeta<Meta> | FileMeta<Meta> | undefined>;
   async getRangeMaybeExpired?<Meta = unknown>(
     key: string,
-    start: number,
-    length: number
-  ): Promise<StoredValueMeta<Meta> | undefined>;
+    offset: number,
+    length?: number
+  ): Promise<RangeStoredValueMeta<Meta> | undefined>;
+  async getSuffixMaybeExpired?<Meta = unknown>(
```
mrbbot, line 91: Could you merge this method with `getRangeMaybeExpired` by allowing `offset` to be `undefined`?

packages/shared/src/storage.ts
```diff
@@ -79,11 +80,18 @@ export abstract class Storage {
...
+    key: string,
+    suffix: number
+  ): Promise<RangeStoredValueMeta<Meta> | undefined>;
```
mrbbot, line 94: I'm still a bit unsure about these methods being optional. Could you define `abstract` methods `head` and `getRange` (with `offset === undefined` for suffix) as above, that are implemented by `@miniflare/storage-memory`, `@miniflare/storage-file` (see https://github.com/cloudflare/miniflare/blob/master/packages/storage-memory/src/local.ts for both of these), and `@miniflare/storage-redis`?

mrbbot (changes requested):
This is great! An incredibly comprehensive test suite too! 😊

I've added my comments. One thing that I'm no longer sure about is the optional storage methods. I'd prefer if we defined `abstract` methods `head` and `getRange` in `Storage`, that were implemted by all implementations (memory, file, Redis) to maintain the abstraction. More details in that comment.

packages/r2/src/bucket.ts
```diff
@@ -0,0 +1,679 @@
...
+interface R2StoredValueMeta<Meta = unknown> extends StoredValueMeta<Meta> {
+  range?: undefined;
+}
+
+// For more information, refer to https://datatracker.ietf.org/doc/html/rfc7232
+export interface R2Conditional {
+  // Performs the operation if the object’s etag matches the given string.
+  etagMatches?: string | string[];
+  // Performs the operation if the object’s etag does not match the given string.
+  etagDoesNotMatch?: string | string[];
```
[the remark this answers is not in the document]
CraigglesO: After talking with vlovich, this was an oversight and should be updated to how I have it now. For example, header "If-Match" could be "a, b, c" meaning it should check against each ["a", "b", "c"].includes(x). Can you check with the team that they still plan on making this update?

packages/r2/src/bucket.ts
```diff
@@ -0,0 +1,679 @@
...
+  // If a delimiter has been specified, contains all prefixes between the specified
+  // prefix and the next occurence of the delimiter.
+  // For example, if no prefix is provided and the delimiter is ‘/’, foo/bar/baz
+  // would return foo as a delimited prefix. If foo/ was passed as a prefix
+  // with the same structure and delimiter, foo/bar would be returned as a delimited prefix.
+  delimitedPrefixes: string[];
+}
+
+const MAX_LIST_KEYS = 1_000;
+const MAX_KEY_SIZE = 1024;
```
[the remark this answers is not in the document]
CraigglesO: https://developers.cloudflare.com/r2/runtime-apis/#r2listoptions -> Defaults to 1000 with a max of 1000.

max_key_size is part of the code that was shared with me on checking key validity. If you would like to see it, ping me on discord.

packages/r2/src/bucket.ts
```diff
@@ -0,0 +1,679 @@
...
+  for (const [key, value] of Object.entries(httpMetadata)) {
+    if (key === "cacheExpiry") {
+      if (!(value instanceof Date) && value !== undefined) {
+        throwR2Error(
+          "PUT",
+          400,
+          "cacheExpiry's value must be a Date or undefined."
+        );
+      }
+    } else {
```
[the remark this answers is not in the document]
CraigglesO: Keen eye, but it is filtered here:
https://github.com/CraigglesO/miniflare/blob/feature/r2/packages/r2/src/r2Object.ts#L102

I checked with vlovich on this, and they prefer filtering over throwing an error.

packages/r2/src/bucket.ts
```diff
@@ -0,0 +1,679 @@
...
+
+    let stored:
+      | R2StoredValueMeta<R2ObjectMetadata>
+      | RangeStoredValueMeta<R2ObjectMetadata>
+      | undefined;
+
+    // get data dependent upon whether suffix or range exists
+    if (typeof range.suffix === "number") {
+      try {
+        stored = await this.#storage.getSuffixMaybeExpired?.<R2ObjectMetadata>(
```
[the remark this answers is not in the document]
CraigglesO: Right below these attempts to get stored we have:
```js
if (stored?.metadata === undefined) return null;
```
Since only file-storage can actually handle these function calls, I assumed we should just return null as if the data does not exist. But maybe we can throw an error instead to let users know they have to use persistance for R2 to work? Open to suggestions.

packages/r2/src/r2Object.ts
```diff
@@ -0,0 +1,323 @@
...
+    conditional;
+  // If the object doesn't exist
+  if (metadata === undefined) {
+    // the etagDoesNotMatch and uploadedBefore automatically pass
+    // etagMatches and uploadedAfter automatically fail if they exist
+    if (etagMatches !== undefined || uploadedAfter !== undefined) {
+      return false;
+    } else {
+      return true;
+    }
```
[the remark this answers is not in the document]
CraigglesO: Gorgeous.

packages/r2/src/r2Object.ts
```diff
@@ -0,0 +1,323 @@
...
+      "contentType",
+      "contentLanguage",
+      "contentDisposition",
+      "contentEncoding",
+      "cacheControl",
+      "cacheExpiry",
+    ];
+    for (const key of Object.keys(httpMetadata)) {
+      if (!httpMetadataList.includes(key)) {
+        delete httpMetadata[key as keyof R2HTTPMetadata];
```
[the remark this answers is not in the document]
CraigglesO: https://github.com/CraigglesO/miniflare/blob/feature/r2/packages/r2/src/bucket.ts#L471
httpMetadata is already pulled from the options and so we can edit it without effecting options.

for instance:
```js
function test (options) {
  let { x } = options
  x++
  return x
}

const options = { x: 0 }
test(options)
console.log(options.x) // 0
```

packages/r2/src/r2Object.ts
```diff
@@ -0,0 +1,323 @@
...
+    this.httpEtag = metadata.httpEtag;
+    this.uploaded = metadata.uploaded;
+    this.httpMetadata = metadata.httpMetadata;
+    this.customMetadata = metadata.customMetadata;
+    this.range = metadata.range;
+  }
+
+  // Retrieves the httpMetadata from the R2Object and applies their corresponding
+  // HTTP headers to the Headers input object. Refer to HTTP Metadata.
+  writeHttpMetadata(headers: Headers): void {
```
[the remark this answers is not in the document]
CraigglesO: No, `writeHttpMetadata` will only include headers found in `R2HTTPMetadata`. I tested this in a live bucket to verify.

packages/r2/src/r2Object.ts
```diff
@@ -0,0 +1,323 @@
...
+  if (typeof a === "string") return a === b;
+  else return a.includes(b);
+}
+
+// headers can be a list: e.g. ["if-match", "a, b, c"] -> "if-match: [a, b, c]"
+function parseHeaderArray(input?: string): undefined | string | string[] {
+  if (input === undefined) return;
+  if (typeof input !== "string") return;
+  const list = input.split(",");
+  if (list.length === 1) return list[0];
```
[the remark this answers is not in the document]
CraigglesO: I couldn't create a test case where this would be true.

The first case is verifying that the `,` didn't exist, so the string wouldn't be modified, in other words, if the user added a space at the end of their singular test, I think it shouldn't be modified.

The second case is trimming the spacing between the commas, so 'a, b, c' should remove the spaces after the commas.

If you can think of an example that trim would be needed feel free to push back.

On second thought, I'm goign to change the function to be more clear:
```ts
// headers can be a list: e.g. ["if-match", "a, b, c"] -> "if-match: [a, b, c]"
function parseHeaderArray(input?: string): undefined | string | string[] {
  if (typeof input !== "string") return;
  if (!input.includes(",")) return input;
  const list = input.split(",");
  return list.map((x) => x.trim());
}
```

packages/r2/src/bucket.ts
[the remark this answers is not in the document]
mrbbot: I don't want users to have to enable persistance to use R2. In general, tests are much easier to write without persistence, and persistence can't be enabled in `jest-environment-miniflare` because of how the isolated storage works. See my comment [here](https://github.com/cloudflare/miniflare/pull/289#discussion_r912348617) for a suggested alternative.

packages/r2/src/r2Object.ts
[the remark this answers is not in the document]
mrbbot: The problem is that you're using `delete` in `parseHttpMetadata`, so you end up with something like:
```js
function parseHttpMetadata(httpMetadata) {
  const httpMetadataList = [
    "contentType",
    "contentLanguage",
    "contentDisposition",
    "contentEncoding",
    "cacheControl",
    "cacheExpiry",
  ];
  for (const key of Object.keys(httpMetadata)) {
    if (!httpMetadataList.includes(key)) {
      delete httpMetadata[key];
    }
  }
  return httpMetadata;
}

function put(options) {
  let { httpMetadata } = options;
  parseHttpMetadata(httpMetadata);
}

let options = {
  httpMetadata: {
    contentType: "text/plain",
    extraKey: "value",
  },
};
console.log(options); // { httpMetadata: { contentType: "text/plain", extraKey: "value" } }
put(options);
console.log(options); // { httpMetadata: { contentType: "text/plain" } }
```

packages/shared/src/storage.ts
```diff
@@ -64,6 +81,20 @@ export abstract class Storage {
     key: string,
     skipMetadata: true
   ): Awaitable<StoredValue | undefined>;
+  abstract getRange<Meta = unknown>(
+    key: string,
+    offset?: number,
+    length?: number,
+    suffix?: number,
```
mrbbot, line 88: Maybe make this take an object, like `{ offset?: number, length?: number, suffix?: number }`? That way you don't have problems with `A required parameter cannot follow an optional parameter.`, and it's a bit clearer what's going on when you call the function?

packages/storage-file/src/index.ts
```diff
@@ -93,21 +105,29 @@ export class FileStorage extends LocalStorage {
 
   async getRangeMaybeExpired<Meta = unknown>(
     key: string,
-    start: number,
-    length: number
-  ): Promise<StoredValueMeta<Meta> | undefined> {
+    inputOffset?: number,
+    inputLength?: number,
+    inputSuffix?: number
```
mrbbot, line 110: Similarly here, this could take an `{ offset, length, suffix }` object.

mrbbot (commented):
This is looking so good! 😄 Thank you for making those storage changes. A few suggestions, then I'll merge. 🎉

ava.config.mjs
```diff
@@ -11,7 +11,7 @@ export default {
   nonSemVerExperiments: {
     nextGenConfig: true,
   },
-  files: ["packages/*/test/**/*.spec.ts"],
+  files: ["packages/r2/test/**/*.spec.ts"],
```
mrbbot: This should be `*`.

mrbbot (approved):
One tiny tiny AVA config change then I'll merge. 🎉 🎉 🎉

Issue #162, opened 2022-01-26
API mocking with MSW fails after upgrade to miniflare 2.2.0


fbarthez:
Hey @mrbbot, thanks for all the work on miniflare, it's been a game-changer!
With miniflare 1, I've been using [mock service worker](https://mswjs.io/) to mock API responses. When attempting to upgrade to miniflare 2.2.0, my tests failed. I've no idea where to start looking for a solution, so I've put up [a fairly minimal demo](https://github.com/fbarthez/miniflare-msw/), hoping that'll make it easier to pinpoint the cause. HEAD is using miniflare 2.2.0, while the [miniflare-1.4.1 branch](https://github.com/fbarthez/miniflare-msw/tree/miniflare-1.4.1) shows what it looks like when it's working. Running `npm test` should show failing tests for HEAD and passing for the latter.

mrbbot:
Hey! 👋 I think this is because instead of `node-fetch`, Miniflare 2 uses [`undici`](https://github.com/nodejs/undici) for its `fetch` implementation, which doesn't use the built-in `http` module, so bypasses [Mock Service Worker's monkeypatching](https://mswjs.io/docs/comparison#nock).

I'm not sure if this is you, but a similar question was recently asked on the Remix Discord server. The response there might be useful at least until Mock Service Worker is updated to work with `undici`: https://discord.com/channels/770287896669978684/777863813981536280/935974246339985438

fbarthez:
Thanks! That'll get me back on track, probably using the [built-in mocking feature of undici](https://nodejs.medium.com/introducing-undici-4-1e321243e007#b30b) for now.

SupremeTechnopriest:
Hey @fbarthez, 

Were you ever able to get this to work?  Im testing out the wrangler beta and Im not able to get the undici mocking to intercept fetch requests.

fbarthez:
Hey @SupremeTechnopriest, no it's still on my todo list and will probably stay there for at least another week. ;/

SupremeTechnopriest:
I kind of need to be able to mock the endpoints.  I also really like the MSW API.  I might bite the bullet and write the interceptor, but it might also be nice for `jest-environment-miniflare` to expose its own mocking framework.  Maybe a wrapper around the undici ones.  @mrbbot what do you think about starting a discussion around that?

SupremeTechnopriest:
[...]

PR #293, opened 2022-06-20
Add Support for Mocking Requests using Undici MockAgent 

william1616:
Hello,

I don't know what the contributor guidelines for miniflare are but it's a cool project and this is a simple workaround for #162 so hopefully this change can be accepted

We use cloudflare workers to write API gateways and we love it but testing has always been a nightmare; miniflare seams like a great way for us to do testing but it's missing one important thing - Fetch Mocks

This PR adds three new functions to Miniflare

`createMockAgent` which wraps undici `MockAgent` constructor - https://github.com/nodejs/undici/blob/main/docs/api/MockAgent.md#new-mockagentoptions
`getGlobalDispatcher` which wraps undici `getGlobalDispatcher` - https://github.com/nodejs/undici#undicigetglobaldispatcher
`setGlobalDispatcher` which wraps undici `setGlobalDispatcher` - https://github.com/nodejs/undici#undicisetglobaldispatcherdispatcher

This allows you to Mock upstream HTTP servers for testing - here's an example

```
const mf = useMiniflare(
  {},
  {
    modules: true,
    script: `export default {
    async fetch(request) {          
      return await fetch(\`https://miniflare.dev\`);
    }
  }`,
  }
);

// Create a Mock Agent and prevent internet access - https://github.com/nodejs/undici/blob/main/docs/api/MockAgent.md
const agent = await mf.createMockAgent();
agent.disableNetConnect();

const client = agent.get("https://miniflare.dev");
client.intercept({ path: "/", method: "GET" }).reply(200, "Hello World!");

const dispatcher = await mf.getGlobalDispatcher();
await mf.setGlobalDispatcher(agent);

const res = await mf.dispatchFetch("http://localhost/");
// Check we get the mocked response and not the actual miniflare.dev
t.is(await res.text(), "Hello World!");

// Reset the Dispatcher for Other Tests - it would be nice if there was a slightly cleaner way of handling this
mf.setGlobalDispatcher(dispatcher);
 ```

There may be a more user friendly ways of handling mocking - having to reset the global dispatcher at the end of the test is a bit annoying but this is hopefully a good and usable start 

Before submitting this PR I have
- run prettier on all files
- run all the unit tests
- written some new unit tests to specifically test this functionality including a test that uses the miniflare jest environment

As mentioned there doesn't seam to be a formal contributors guide so if I've missed anything please let me know and I'm happy to amend

### CODE REVIEW

**Comment by william1616** [comment]
@mrbbot thanks for the comments - I'm almost done addressing them; the one thing I can't get working is Jest

it looks like mf.setOptions isn't correctly setting the fetchMock - https://github.com/cloudflare/miniflare/pull/293/files#diff-030a0f6278d54bf8f41b941035a028c124f4b996e01f97b995fda35425cc668aR233 however I can't see why; mf.setOptions() will call #init() which should set fetchMock into the PluginContext (to test the context is actually being updated i've tried forcing ctxUpdate to true which doesn't seam to work) - hopefully you'll have an idea

once that's sorted it should be good to go

**Comment by andrevechina** [comment]
👍

File: packages/core/test/index.spec.ts [hunk 1]
```diff
@@ -49,6 +49,7 @@ import {
 } from "@miniflare/shared-test";
 import test, { Macro } from "ava";
 import { Request as BaseRequest, File, FormData } from "undici";
+import { createFetchMock } from "../src/standards/http";
```
mrbbot, line 52: You should re-export `createFetchMock` from `packages/core/standards/index.ts`, then import this from `@miniflare/core`.

packages/jest-environment-miniflare/src/index.ts
```diff
@@ -21,13 +21,17 @@ import { SitesPlugin } from "@miniflare/sites";
 import { WebSocketPlugin } from "@miniflare/web-sockets";
 import { ModuleMocker } from "jest-mock";
 import { installCommonGlobals } from "jest-util";
+import { createFetchMock } from "packages/core/src/standards/http";
```
mrbbot, line 24: Again, re-export createFetchMock from `packages/core/standards/index.ts`, then import this from `@miniflare/core` instead.

packages/jest-environment-miniflare/src/index.ts
```diff
@@ -221,6 +225,13 @@ export default class MiniflareEnvironment implements JestEnvironment<Timer> {
       const state = await plugin.getObject(storage, id);
       return state.storage;
     };
+    global.getMiniflareFetchMock = () => {
+      const agent = createFetchMock();
+      return agent;
+    };
+    global.setMiniflareFetchMock = async (agent: MockAgent) => {
+      await mf.setOptions({ fetchMock: agent });
+    };
```
mrbbot, line 234: I was thinking we create a single fetch mock, pass it when we construct the `MiniflareCore` instance above, then only provide a `getMiniflareFetchMock` function to get it. As far as I know, `MockAgent` allows all network requests by default.

packages/jest-environment-miniflare/test/fixtures/core.mock.spec.js
```diff
@@ -0,0 +1,11 @@
...
+
+test("handles requests", async () => {
+  const mockAgent = getMiniflareFetchMock();
+  mockAgent.disableNetConnect();
+  const client = mockAgent.get("https://random.mf");
+  client.intercept({ path: "/" }).reply(200, "Hello World!");
+  await setMiniflareFetchMock(mockAgent);
+  const res = await worker.fetch(new Request("https://random.mf"));
+  expect(await res.text()).toBe("Hello World!");
+});
```
mrbbot, line 11: I'd stick this test in `core.module.spec.js`, then you can remove the extra `test.serial` in `packages/jest-environment-miniflare/test/index.spec.ts`.

mrbbot (commented):
Hey! 👋 Apologies for the delayed response. I've recently returned from a long holiday and am catching up on issues now. Thanks for making those changes. 🙂 I've added some suggestions.

william1616:
@mrbbot all changes made - ready for a re-review!

mrbbot (approved):
Amazing! 😃 Thank you! I've just pushed a couple minor linting fixes, but LGTM! ✅

william1616:
Great! Would love to start using this ASAP - does miniflare work to a planned release schedule or are new versions pushed more ad hoc?

Issue #297, opened 2022-06-26
Serialisation error with storage.list() when an alarm is set


evanderkoogh:
Steps to reproduce:

1) `this.state.storage.setAlarm(Date.now() + 60 * 1000)`
2) `this.state.storage.list()`
3) Profit!

It says that the serialisation format has changed from miniflare 1 to 2, but I am not storing anything on disk, let alone something from miniflare 1. Both projects this happened on were initialised with wrangler2 and have only ever included miniflare 2.

PR #298, opened 2022-06-26
adjust for alarm storage

CraigglesO:
Fixes https://github.com/cloudflare/miniflare/issues/297.

### DIFF
--- a/packages/durable-objects/src/storage.ts
+++ b/packages/durable-objects/src/storage.ts
@@ -183,6 +183,7 @@ async function list<Value = unknown>(
       "list() cannot be called with both start and startAfter values."
     );
   }
+  options = { ...options };
   const originalLimit = options.limit;
   // Since alarms now exist in storage, add 1 to the limit to account for
   // the alarm key.
@@ -192,7 +193,7 @@ async function list<Value = unknown>(
     // Then if `startAfter` does exist as a key, we can remove it later.
     // To ensure we still return `limit` keys in this case, add 1 to the limit
     // if one is set.
-    options = { ...options, start: options.startAfter };
+    options.start = options.startAfter;
     if (options.limit !== undefined) options.limit++;
   }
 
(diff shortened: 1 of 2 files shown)


packages/durable-objects/src/storage.ts
```diff
@@ -184,6 +184,9 @@ async function list<Value = unknown>(
     );
   }
   const originalLimit = options.limit;
+  // Since alarms now exist in storage, add 1 to the limit to account for
+  // the alarm key.
+  if (options.limit !== undefined) options.limit++;
```
mrbbot, line 189: This has the side effect of mutating `limit` for the caller because `options` is passed by reference. If they called this method again with the same `options` object, they'd get different results. I'd do something like this instead:
```suggestion
  // Clone options so we can mutate it
  options = { ...options };
  // Since alarms now exist in storage, add 1 to the limit to account for
  // the alarm key.
  if (options.limit !== undefined) options.limit++;
```
...then also replace `options = { ...options, start: options.startAfter };` further down with `options.start = startAfter;` as there's no need for the additional clone.

packages/durable-objects/src/storage.ts
```diff
@@ -194,17 +197,18 @@ async function list<Value = unknown>(
   }
 
   const { keys } = await storage.list(options);
-  let keyNames = keys.map(({ name }) => name);
+  let keyNames = keys
+    .map(({ name }) => name)
+    .filter((name) => name !== ALARM_KEY);
 
   if (options.startAfter !== undefined) {
     if (keyNames[0] === options.startAfter) {
```
mrbbot, line 205: These `if` statements can be collapsed now.

mrbbot (changes requested):
Hey! 👋 Thanks for fixing this. One tiny mutability concern then I'll merge. 🙂

CraigglesO:
Should be ready now. 😄

mrbbot (approved):

PR #309, opened 2022-07-12
preserve file path in FormData objects

yusefnapora:
Hi, I'm working on a project that pulls `File` objects out of multipart `FormData` uploads and needs the full file path in order to correctly calculate some hashes and so on.

In a real CF worker, `File` objects from multipart uploads include the full path in their `name` property, e.g. `directory/foo.txt`, but Miniflare just sets the basename (`foo.txt`).

This PR just sets `preservePath: true` when constructing a `Busboy` instance, to make the behavior line up with the real worker env.

Let me know if there's any changes you'd want, or if you'd like me to file an issue to track the bug.

Cheers :)


mrbbot (approved):
Hey! 👋 Thanks for the PR! Looks great! ✅ Apologies for the delayed response. I've recently returned from a long holiday.

PR #311, opened 2022-07-12
fix NotSupportedError when using NODE-ED25519

### LINKED ISSUE #310: NotSupportedError when using NODE-ED25519 keys
When calling `crypto.subtle.importKey` or `crypto.subtle.generateKey` with an algorithm of `{ name: 'NODE-ED25519', namedCurve: 'NODE-ED25519' }`, Miniflare fails with a `DOMException` with `name == "NotSupportedError"`.

I've got a branch locally that fixes this that I'll PR in a bit :)

yusefnapora:
This closes #310 by adding a helper function that converts `NODE-ED25519` to `Ed25519`.

I didn't add a check for just the string form (without the `namedCurve`), since I wasn't sure if that's supported in the real CF worker env. But it would be super easy to add it to the helper.


packages/core/src/standards/crypto.ts
```diff
@@ -63,20 +63,35 @@ function digest(
   return webcrypto.subtle.digest(algorithm, data);
 }
 
+function validAlgorithm(algorithm: AlgorithmIdentifier): AlgorithmIdentifier {
+  if (typeof algorithm === 'object' && algorithm.name === 'NODE-ED25519') {
```
mrbbot, line 67: Cloudflare requires `namedCurved` be set to `NODE-ED25519` too (see footnotes [here](https://developers.cloudflare.com/workers/runtime-apis/web-crypto/#supported-algorithms)), so could you add an additional check for `algorithm.namedCurved === "NODE-ED25519"`. Also note Miniflare has a lint rule for using double quotes.

packages/core/src/standards/crypto.ts
```diff
@@ -63,20 +63,35 @@ function digest(
...
+    return 'Ed25519';
```
mrbbot, line 68: Node will log an additional experimental warning if `"Ed25519"` is used instead of `{ name: "Ed25519", namedCurve: "Ed25519" }`. Let's return that instead.

packages/core/src/standards/crypto.ts
mrbbot, line 67: We should only make this replacement in recent versions of Node.js. Miniflare supports all Node versions `>= 16.7.0`. I'd put the following feature detect at the top level of this file, then `await` the result in this function before replacing.

```js
const usesModernEd25519 = (async () => {
  try {
    // Modern versions of Node.js expect `Ed25519` instead of `NODE-ED25519`.
    // This will throw a `DOMException` if `Ed25519` should be used instead.
    // See https://github.com/nodejs/node/pull/42507.
    await webcrypto.subtle.generateKey(
      { name: "NODE-ED25519", namedCurve: "NODE-ED25519" },
      false,
      ["sign", "verify"]
    );
    return false;
  } catch {
    return true;
  }
})();
```

```suggestion
  if (typeof algorithm === "object" && algorithm.name === "NODE-ED25519" && algorithm.namedCurve === "NODE-ED25519" && await usesModernEd25519) {
```

packages/core/src/standards/crypto.ts
```diff
@@ -63,20 +63,35 @@ function digest(
...
+      // @ts-expect-error importKey is missing from TS declaration for SubtleCrypto
+      webcrypto.subtle.importKey(type, key, validAlgorithm(alg), extractable, usages),
     blockGlobalRandom
   );
 
   const subtle = new Proxy(webcrypto.subtle, {
     get(target, propertyKey, receiver) {
       if (propertyKey === "digest") return digest;
       if (propertyKey === "generateKey") return generateKey;
+      if (propertyKey === "importKey") return importKey;
```
mrbbot, line 94: We'll also need to wrap the `crypto.subtle.sign` and `crypto.subtle.verify` methods in a similar way.

packages/core/src/standards/crypto.ts
```diff
@@ -63,20 +63,35 @@ function digest(
...
   const generateKey = assertsInRequest(
-    webcrypto.subtle.generateKey.bind(webcrypto.subtle),
+    async (alg, extractable, usages) => 
+      webcrypto.subtle.generateKey(validAlgorithm(alg), extractable, usages),
+    blockGlobalRandom
+  );
+  const importKey = assertsInRequest(
+    async (type, key, alg, extractable, usages) => 
+      // @ts-expect-error importKey is missing from TS declaration for SubtleCrypto
+      webcrypto.subtle.importKey(type, key, validAlgorithm(alg), extractable, usages),
```
mrbbot, line 86: Cloudflare Workers only allow you to import `raw` public `Ed25519` keys. If `type` is `raw` and `alg` is `Ed25519`, could we add `public: true` to the `alg` too?

mrbbot (commented):
Hey! 👋 Apologies for the delayed response. I've recently returned from a long holiday and am just starting to catch up on issues and PRs now.

I've added some comments. Let me know if you have any questions. 🙂

Issue #318, opened 2022-07-20
storage-file -> readFileRange does not respect symbolic links


CraigglesO:
https://github.com/cloudflare/miniflare/blob/master/packages/storage-file/src/helpers.ts#L40

If a file is a symbolic link, the size of the symbolic link is returned rather than the actual file size.

This can be fixed by using [realpath](https://nodejs.org/api/fs.html#fspromisesrealpathpath-options)

I can fix it this weekend if a PR hasn't been created by then.

PR #319, opened 2022-07-21
add realpath

CraigglesO:
Fixes https://github.com/cloudflare/miniflare/issues/318

While `BUCKET.put` will never create this problem, I found adding large data manually to r2-cache using symbolic links failed. The size reported would be the size of the symbolic link itself.


mrbbot (commented):
Hey! 👋 Apologies for the delayed response. I've recently returned from a long holiday and am just starting to catch up on issues and PRs now. Would you be able to add a test? then happy to merge. 🙂

CraigglesO:
Absolutely no worries.
You're right, I should have added a test case.
Should be good now ☘️

mrbbot (approved):

PR #320, opened 2022-07-21
Upgrade undici to `v5.8.0`

yusukebe:
Hi! Upgraded *undici* to `v5.8.0`. This release fixed the vulnerabilities.

https://github.com/nodejs/undici/releases/tag/v5.8.0

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7607,9 +7607,9 @@
       }
     },
     "node_modules/undici": {
-      "version": "5.5.1",
-      "resolved": "https://registry.npmjs.org/undici/-/undici-5.5.1.tgz",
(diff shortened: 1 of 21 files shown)


yusukebe:
This `undici` updating has breaking changes. We can't adapt as it is.
But, I think we should fix the vulnerabilities. Hmm...

yusukebe:
Hi @mrbbot ! Thank you for checking this PR.

> Will check we can bump the minimum supported Node version (in Wrangler too)

Please!

> but these security issues are unlikely to affect Miniflare users, as it's only intended as a local development and testing tool.

Ah, you're definitely right!

PR #328, opened 2022-08-03
[WIP] Use the class name as the identifier for Durable Objects

GregBrimble:
IDs should be scoped to the class, rather than to whatever they're bound as. This unlocks us communicating between two different services which bind to the same Durable Object.


GregBrimble:
Closed because we don't want to make a breaking change. Context here: https://github.com/cloudflare/wrangler2/pull/1611.

PR #329, opened 2022-08-04
D1 beta support

geelen:
This adds `@miniflare/d1` package:

* `Statement` class, acts similarly to https://github.com/WiseLibs/better-sqlite3/blob/master/docs/api.md#class-statement but is entirely lazy, only truly gets prepared & bound when `.all()`, `.first()`, `.run()` or `.raw()` are called.
* `BetaDatabase` class, acts as a D1 binding within Miniflare
* `createSQLiteDB` returns a `better-sqlite3` DB instance for manipulating an in-memory or on-disk Sqlite3 database. Used in Wrangler for running commands against local DBs.
* `getSQLiteNativeBindingLocation` returns the location of the `better_sqlite3.node` file, which may not be installed in the current project.

Note this uses `npx-import` (https://npx-import.pages.dev) to lazily load better-sqlite3, otherwise every installation of miniflare would add ~30s to installation (including `npx wrangler ...`)

> Aside: You can run `time npx -y -p better-sqlite3 echo` to see how long it takes on your machine:

```
npx -y -p better-sqlite3@~7.6.2 echo  21.77s user 0.69s system 92% cpu 24.276 total
```

On a M1 Pro ^


packages/d1/README.md
```diff
@@ -0,0 +1,44 @@
...
+## Example
+
+```js
+import { BetaDatabase } from "@miniflare/d1";
+import { MemoryStorage } from "@miniflare/storage-memory";
+const db = new BetaDatabase(new MemoryStorage());
+
+// BetaDatabase only supports .fetch(), once D1 is out of beta the full API will be available here:
+await db.fetch("/execute", {
+  method: 'POST',
```
mrbbot, line 16: Nit: this, and the method below, should use double-quotes 🙃

packages/d1/src/database.ts
```diff
@@ -0,0 +1,198 @@
...
+    this.#db = db;
+    this.#query = query;
+    this.#bindings = bindings;
+  }
+
+  // Lazily accumulate binding instructions, because ".bind" in better-sqlite3
+  // is a real action that means the query must be valid when it's written,
+  // not when it's about to be executed (i.e. in a batch).
+  bind(...params: BindParams) {
+    return new Statement(this.#db, this.#query, params);
```
mrbbot, line 37: Not sure what the intended behaviour is here, but should `params` be merged with `#bindings`?

packages/d1/src/database.ts
```diff
@@ -0,0 +1,198 @@
...
+  }
+  private prepareAndBind() {
```
mrbbot, line 39: As `Statement` gets exposed to users' workers, I'd recommend using JavaScript private class features to hide this method (and all other `private (static)`s below) at runtime.
```suggestion
  #prepareAndBind() {
```

packages/d1/src/database.ts
```diff
@@ -0,0 +1,198 @@
...
+        lastRowId: null,
+        changes: null,
+        success: true,
+        served_by: "x-miniflare.db3",
+      };
+    } catch (e) {
+      throw errorWithCause("D1_ALL_ERROR", e);
+    }
+  }
+  private static _all(statementWithBindings: SqliteStatement) {
```
mrbbot, line 74:
```suggestion
  #all(statementWithBindings: SqliteStatement) {
```

packages/d1/src/database.ts
```diff
@@ -0,0 +1,198 @@
...
+  async first(col?: string) {
+    const statementWithBindings = this.prepareAndBind();
+    try {
+      const data = Statement._first(statementWithBindings);
+      return typeof col === "string" ? data[col] : data;
+    } catch (e) {
+      throw errorWithCause("D1_FIRST_ERROR", e);
+    }
+  }
+  private static _first(statementWithBindings: SqliteStatement) {
```
mrbbot, line 100:
```suggestion
  #first(statementWithBindings: SqliteStatement) {
```
...although could this just be inlined?

packages/d1/src/database.ts
```diff
@@ -0,0 +1,198 @@
...
+        lastRowId: lastInsertRowid,
+        changes,
+        success: true,
+        served_by: "x-miniflare.db3",
+      };
+    } catch (e) {
+      throw errorWithCause("D1_RUN_ERROR", e);
+    }
+  }
+  private static _run(statementWithBindings: SqliteStatement) {
```
mrbbot, line 123:
```suggestion
  #run(statementWithBindings: SqliteStatement) {
```
...although could this just be inlined?

packages/d1/src/database.ts
```diff
@@ -0,0 +1,198 @@
...
+    return statementWithBindings.run();
+  }
+
+  async raw() {
+    const statementWithBindings = this.prepareAndBind();
+    return Statement._raw(statementWithBindings);
+  }
+  private static _raw(statementWithBindings: SqliteStatement) {
```
mrbbot, line 131:
```suggestion
  #raw(statementWithBindings: SqliteStatement) {
```
...although could this just be inlined?

packages/d1/src/database.ts
```diff
@@ -0,0 +1,198 @@
...
+
+export class BetaDatabase {
+  readonly #storage: Storage;
+  #db?: SqliteDB;
+
+  constructor(storage: Storage) {
+    this.#storage = storage;
+  }
+
+  async init() {
```
mrbbot, line 149: Guessing this shouldn't be exposed to users? Could you use a `Symbol` here instead so users couldn't call this:
```ts
export const kInit = Symbol("kInit");
```
```suggestion
  async [kInit]() {
```

packages/d1/src/plugin.ts
```diff
@@ -0,0 +1,71 @@
...
+    this.assignOptions(options);
+    this.#persist = resolveStoragePersist(ctx.rootPath, this.d1Persist);
+  }
+
+  async getBetaDatabase(
+    storage: StorageFactory,
+    dbName: string
+  ): Promise<BetaDatabase> {
+    const db = new BetaDatabase(await storage.storage(dbName, this.#persist));
+    await db.init();
```
mrbbot, line 50:
```ts
import { kInit } from "./database";
```
```suggestion
    await db[kInit]();
```

packages/storage-memory/src/local.ts
```diff
@@ -116,4 +116,8 @@ export abstract class LocalStorage extends Storage {
     await Promise.all(deletePromises);
     return res;
   }
+
+  getSqliteDatabasePath(): string {
+    return ":memory:";
+  }
```
mrbbot, line 122: Have you tested what happens if you use an in-memory database, then update the script? Do the database contents stick around? Guessing they do until you edit D1 options in `wrangler.toml`. This is slightly different behaviour to KV/DO/cache that will persist in memory until you restart Miniflare, but as long as they don't reset on script reloads that's good. 🙂

packages/miniflare/src/api.ts
```diff
@@ -49,6 +50,7 @@ export const PLUGINS = {
 
   // Storage
   KVPlugin,
+  D1Plugin,
```
mrbbot, line 53: You'll need to add `D1Plugin` here (https://github.com/cloudflare/miniflare/blob/master/packages/shared-test-environment/src/plugins.ts) too so it works in the Jest/Vitest environments.

mrbbot (commented):
Nice! 🎉 Added a few comments, mostly around hiding internal methods from users' workers.

CraigglesO:
You may have already fixed this. Browsing the code I couldn't tell. I made a duplicate of this branch for myself just for testing purposes because if you use memory (and possibly persist) in testing:

1) beforeEach function adds data to DB
2) normal test fetches a worker that looks up said data

What happens is the data is wiped because the plugin asks for the DB again... but an in memory DB would just create a new one, and I think sometimes there are issues even with a persist db.

My patch works but isn't great just to show you [how I fixed it](https://github.com/CraigglesO/miniflare/blob/d1-beta-next/packages/d1/src/plugin.ts#L49). But I figured it would be something good to know before this is merged.

geelen:
That's a good point, @CraigglesO... Each test _should_ start from a clean slate (if you're using in-memory DBs, at least), but `beforeEach` is really part of a test, so it shouldn't get wiped between.

Does the same problem occur with KV or R2? I wonder if it's something boneheaded I've done with the D1 stuff or a problem more generally...

CraigglesO:
So KV and R2 pull data in from the [storage class](https://github.com/cloudflare/miniflare/blob/master/packages/r2/src/plugin.ts#L49).
For instance memory storage just keeps all key-values [in a map](https://github.com/cloudflare/miniflare/blob/master/packages/storage-memory/src/memory.ts#L8). BUT, once a storage is created, it's stored in a `StorageFactory` as a [top level plugin in Miniflare](https://github.com/cloudflare/miniflare/blob/master/packages/core/src/storage.ts#L4), so it won't be dumped as long as the miniflare instance exists (or disposed).

Currently in the case of memory (and I'd argue for persistence as well), it is creating a brand new `better-sqlite` instance everytime it is fetched.

If you cache the instance for later fetches that should work fine. You can look to DurableObjects for how to cache and [cleanup on each reload](https://github.com/cloudflare/miniflare/blob/master/packages/durable-objects/src/plugin.ts#L289), [and on dispose](https://github.com/cloudflare/miniflare/blob/master/packages/durable-objects/src/plugin.ts#L330).

No worries, this project has become somewhat monolithic. I had a terrible first PR but this one looks great otherwise.

@mrbbot - correct me if I'm wrong here please.

packages/d1/README.md
```diff
@@ -0,0 +1,44 @@
...
+Workers D1 module for [Miniflare](https://github.com/cloudflare/miniflare): a
+fun, full-featured, fully-local simulator for Cloudflare Workers. See
+[📦 D1](https://miniflare.dev/storage/d1) for more details.
+
+## Example
+
+```js
+import { BetaDatabase } from "@miniflare/d1";
+import { MemoryStorage } from "@miniflare/storage-memory";
+const db = new BetaDatabase(new MemoryStorage());
```
mrbbot, line 12: `BetaDatabase` doesn't take a `Storage` anymore. 🙁

packages/d1/README.md
```diff
@@ -0,0 +1,44 @@
...
+  body: JSON.stringify({
+    sql: `CREATE TABLE my_table (cid INTEGER PRIMARY KEY, name TEXT NOT NULL);`,
+  }),
+});
+const response = await db.fetch("/query", {
+  method: "POST",
+  body: JSON.stringify({
+    sql: `SELECT * FROM sqlite_schema`,
+  }),
+});
```
mrbbot, line 26: `BetaDatabase` doesn't actually support these functions anymore right?

mrbbot (commented):
Looks good! ✅ Couple tiny things then let's get this released 🎉 I'm going to start writing the changelog/docs for `2.9.0` now. 👍

PR #333, opened 2022-08-15
Bump undici version and minimum node version

cameron-robey:
To target security vulnerabilities in undici, we update to the latest version.

From undici v5.7.0, we require a node >=16.8.0 - here we update the minimum to 16.13.0 (the first Node 16 LTS release)


JacobMGEvans (approved):
As long as @mrbbot is good with these changes, I think it is good to go.

cameron-robey:
Just worth noting that this will require some people to bump their version of node - I assume we have some precedent for a release like this

mrbbot (approved):
Nice, thank you! 🙂 Added a couple tiny commits to fix tests, but otherwise... ✅

threepointone:
Would you be ok using 16.8.0 as the minimum version instead? Makes the breaking change radius a little smaller.

threepointone:
That's fair. Let's land this?

threepointone:
wait it was landed already ha

Issue #202, opened 2022-03-03
FetchEvent#waitUntil in Jest context


jamesarosen:
### Problem
The promises passed to `FetchEvent.prototype.waitUntil` cannot be retrieved via `Response.prototype.waitUntil()`. There doesn't seem to be any other way to access them either, but if there is, the [Writing and Running Tests docs](https://miniflare.dev/testing/jest#writing-and-running-tests) don't mention it.

### Configuration
```js
// jest.config.js
module.exports = {
  testEnvironment: 'miniflare',
}
```

### Example Code
```js
describe('waitUntil', () => {
  it('caches things', async () => {
    async function myHandler(fetchEvent) {
      const { url } = fetchEvent.request
      fetchEvent.waitUntil(caches.default.put(url, new Response('written to cache')))
      return new Response('returned from worker')
    }

    // build a FetchEvent:
    const url = 'https://example.com/cache-something'
    const request = new Request(url)
    const fetchEvent = new FetchEvent('fetch', { request })

    // get a response for it:
    const response = await myHandler(fetchEvent)

    // the response should have the Promises from `fetchEvent.waitUntil`:
    const awaited = await response.waitUntil()
    expect(awaited[0]).toBeTruthy() // this fails; awaited is []
    expect(await awaited[0].text()).toBe('written to cache')

    // and the cache-write should have completed:
    const cachedResponse = await caches.default.match(url)
    expect(cachedResponse).toBeTruthy()
    expect(await cachedResponse.text()).toBe('written to cache')
  })
})
```

jamesarosen:
The implementation of `FetchEvent.prototype.waitUntil` stores the promises, but they're locked away behind a Symbol: https://github.com/cloudflare/miniflare/blob/dc55621c0767e462b05635f8dd6ba79f4f9445f7/packages/core/src/standards/event.ts#L102-L107

mrbbot:
Hey! 👋 Thanks for raising this. Will try get a `getWaitUntil(event)` method added and exposed to tests soon. 🙂

jamesarosen:
@mrbbot any news on this? Can I help?

mrbbot:
Hey! 👋 Apologies for the delayed response. I've recently returned from a long holiday and am catching up on issues now.

@jamesarosen I'd be very happy to accept a pull-request for this, otherwise I can do it. You'll need to add a `getMiniflareWaitUntil` method here...

[...]

PR #345, opened 2022-08-22
Feature: getMiniflareWaitUntil

CraigglesO:
Resolves https://github.com/cloudflare/miniflare/issues/202

Adds `getMiniflareWaitUntil(event: FetchEvent | ScheduledEvent): Promise<unknown>[];` to return an array of promises stored via `event.waitUntill(...)`

Example usage for both `FetchEvent` and `ScheduledEvent` can be found at `packages/jest-environment-miniflare/test/fixtures/waitUntil.worker.spec.js`


packages/core/src/standards/event.ts
```diff
@@ -34,7 +34,7 @@ const SUGGEST_GLOBAL_BINDING_MODULES =
 
 const kResponse = Symbol("kResponse");
 const kPassThrough = Symbol("kPassThrough");
-const kWaitUntil = Symbol("kWaitUntil");
+export const kWaitUntil = Symbol("kWaitUntil");
```
mrbbot, line 37: Instead of exporting this (since we don't want people to be able to edit it), I'd define a `getWaitUntil(FetchEvent | ScheduledEvent)` function in this file. Moreover, I think instead of just returning the `[kWaitUntil]` `Promise<unknown>[]`, I think we should always call `Promise.all()` first and return a `Promise<unknown[]>`, as I see very few circumstances where people wouldn't immediately need to do that afterwards anyway. Taking this further still, the function should be generic in `WaitUntil` like `Response`: https://github.com/cloudflare/miniflare/blob/6075e19b07538128d2de7c7a31f6fb2213b24044/packages/core/src/standards/http.ts#L501

packages/jest-environment-miniflare/src/index.ts
```diff
@@ -239,6 +245,11 @@ export default class MiniflareEnvironment implements JestEnvironment<Timer> {
       return state.storage;
     };
     global.getMiniflareFetchMock = () => this.mockAgent;
+    global.getMiniflareWaitUntil = (
+      event: FetchEvent | ScheduledEvent
+    ): Promise<unknown>[] => {
+      return event[kWaitUntil];
+    };
```
mrbbot, line 252: With the above change, you could replace this with a reference to the `getWaitUntil` function.

mrbbot (commented):
Hey! 👋 Thanks for the PR! I've added some suggestions. I think we should implicitly `Promise.all()` for users, since they're very likely to do this themselves anyway.

packages/jest-environment-miniflare/src/index.ts
```diff
@@ -30,12 +33,30 @@ import { installCommonGlobals } from "jest-util";
...
+  [kWaitUntil]: Promise<unknown>[] = [];
+
+  passThroughOnException(): void {}
+
+  waitUntil(promise: Promise<any>): void {
+    this[kWaitUntil].push(promise);
+  }
+
+  flush<WaitUntil extends any[] = unknown[]>(): Promise<WaitUntil> {
+    return Promise.all(this[kWaitUntil]) as Promise<WaitUntil>;
```
CraigglesO, line 46: Apologies, I might be doing this wrong, but I get an error if I don't explain that any doesn't fall outside unknown.

mrbbot (commented):
Hey! 👋 Added a couple more comments, but otherwise, looks good. 🙂

packages/jest-environment-miniflare/src/index.ts
```diff
@@ -111,6 +127,7 @@ export default class MiniflareEnvironment implements JestEnvironment<Timer> {
     global.clearTimeout = clearTimeout;
     global.setInterval = setInterval;
     global.setTimeout = setTimeout;
+    global.ExecutionContext = ExecutionContext;
```
mrbbot, line 130: You can remove this line. We only specify these here because `installCommonGlobals` does some magic stuff to `setTimeout`, etc.
```suggestion
```

mrbbot (commented):
One tiny thing, then I'll merge! ✅

packages/jest-environment-miniflare/src/index.ts
```diff
@@ -195,6 +212,7 @@ export default class MiniflareEnvironment implements JestEnvironment<Timer> {
           setInterval: global.setInterval,
           clearTimeout: global.clearTimeout,
           clearInterval: global.clearInterval,
+          ExecutionContext: global.ExecutionContext,
```
mrbbot, line 215:
```suggestion
          ExecutionContext,
```

CraigglesO:
My fault for not finishing. Got really busy this week again. Just happy to see this go through!

PR #355, opened 2022-09-01
WIP R2 API

penalosa:
Working (probably), but still writing tests

### DIFF
--- a/packages/tre/src/plugins/r2/gateway.ts
+++ b/packages/tre/src/plugins/r2/gateway.ts
@@ -1,38 +1,20 @@
-import crypto from "crypto";
-import { Blob } from "buffer";
-import { arrayBuffer } from "stream/consumers";
-import { ReadableStream } from "stream/web";
-import { TextEncoder } from "util";
+import { RangeStoredValueMeta, Storage } from "@miniflare/shared";
+import { InvalidRange, NoSuchKey } from "./errors";
 import {
-  RangeStoredValueMeta,
-  RequestContext,
-  Storage,
-  assertInRequest,
-  getRequestContext,
-  viewToArray,
-  waitForOpenInputGate,
-  waitForOpenOutputGate,
-  Clock,
-} from "@miniflare/shared";
-import { Headers } from "undici";
-import {
+  R2HTTPMetadata,
   R2Object,
   R2ObjectBody,
-  createHash,
+  R2ObjectMetadata,
   createVersion,
-  parseHttpMetadata,
-  parseOnlyIf,
-  parseR2ObjectMetadata,
-  testR2Conditional,
 } from "./r2Object";
-import { R2HTTPMetadata, R2ObjectMetadata } from "./r2Object";
+import { Validator } from "./validator";
 
 // For more information, refer to https://datatracker.ietf.org/doc/html/rfc7232
 export interface R2Conditional {
   // Performs the operation if the object’s etag matches the given string.
-  etagMatches?: string | string[];
+  etagMatches?: string;
   // Performs the operation if the object’s etag does not match the given string.
-  etagDoesNotMatch?: string | string[];
+  etagDoesNotMatch?: string;
   // Performs the operation if the object was uploaded before the given date.
   uploadedBefore?: number;
   // Performs the operation if the object was uploaded after the given date.
@@ -48,31 +30,21 @@ export interface R2Range {
 export interface R2GetOptions {
   // Specifies that the object should only be returned given satisfaction of
   // certain conditions in the R2Conditional. Refer to R2Conditional above.
-  onlyIf?: R2Conditional | Headers;
+  onlyIf?: R2Conditional;
   // Specifies that only a specific length (from an optional offset) or suffix
   // of bytes from the object should be returned. Refer to
   // https://developers.cloudflare.com/r2/runtime-apis/#ranged-reads.
   range?: R2Range;
 }
 
-export type R2PutValueType =
-  | ReadableStream
-  | ArrayBuffer
-  | ArrayBufferView
-  | string
-  | null
-  | Blob;
 export interface R2PutOptions {
-  // Specifies that the object should only be stored given satisfaction of
-  // certain conditions in the R2Conditional. Refer to R2Conditional above.
-  onlyIf?: R2Conditional | Headers;
   // Various HTTP headers associated with the object. Refer to
   // https://developers.cloudflare.com/r2/runtime-apis/#http-metadata.
-  httpMetadata?: R2HTTPMetadata | Headers;
+  httpMetadata: R2HTTPMetadata;
   // A map of custom, user-defined metadata that will be stored with the object.
-  customMetadata?: Record<string, string>;
+  customMetadata: Record<string, string>;
   // A md5 hash to use to check the recieved object’s integrity.
-  md5?: ArrayBuffer | string;
+  md5?: string;
 }
 
 export type R2ListOptionsInclude = ("httpMetadata" | "customMetadata")[];
@@ -85,11 +57,9 @@ export interface R2ListOptions {
   // An opaque token that indicates where to continue listing objects from.
   // A cursor can be retrieved from a previous list operation.
   cursor?: string;
-  // Key after which the list results should start, exclusive.
-  startAfter?: string;
   // The character to use when grouping keys.
   delimiter?: string;
-  // Can include httpMetadata and/or customMetadata. If included, items returned by
+  // Can include httpFields and/or customFields. If included, items returned by
   // the list will include the specified metadata. Note that there is a limit on the
   // total amount of data that a single list operation can return.
   // If you request data, you may recieve fewer than limit results in your response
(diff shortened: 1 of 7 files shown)


packages/tre/src/plugins/r2/gateway.ts
```diff
@@ -1,24 +1,537 @@
...
+  } else if (value instanceof Blob) {
+    return new Uint8Array(await value.arrayBuffer());
+  } else {
+    throw new TypeError(
+      "R2 put() accepts only nulls, strings, Blobs, ArrayBuffers, ArrayBufferViews, and ReadableStreams as values."
+    );
+  }
+}
 
 export class R2Gateway {
```
mrbbot, line 340: My thinking with `R2Gateway` is that it contains shared code between `R2Router`, and a future `R2Bucket` class that we might implement (when we start thinking about integration testing for users' workers). R2 bucket bindings are instances of `R2Bucket` in the runtime.

This means that we probably don't need to return `R2Object(Body)`s from methods in `R2Gateway`, as this would be the responsibility of `R2Bucket`. At this stage, I don't actually think we need the `R2Object`/`R2ObjectBody` classes. Similarly, `R2Gateway#put` should probably just take a `Uint8Array` value instead of an `R2PutValueType`. We're always going to pass a `Uint8Array` from `R2Router`, and `R2PutValueType` normalisation should go in `R2Bucket`.

Essentially, everything that's defined in the runtime's C++ files will go in the future `R2Bucket` class, and everything that's in the R2 gateway worker should get split between `R2Router` and `R2Gateway` classes. 🙂

packages/tre/src/plugins/r2/router.ts
```diff
@@ -16,20 +17,30 @@ export interface R2Params {
...
     return new Response();
   };
 
-  @PUT("/:bucket/:key")
-  put: RouteHandler<R2Params> = async (req, params) => {
+  @PUT("/:bucket")
+  put: RouteHandler<Omit<R2Params, "key">> = async (req, params) => {
+    const bytes = await req.arrayBuffer();
+
+    const metadataSize = Number(req.headers.get("cf-r2-metadata-size"));
```
mrbbot, line 32: Nit: could you extract `CF-R2-Metadata-Size` out into a constant?

packages/tre/src/plugins/r2/router.ts
```diff
@@ -44,7 +55,16 @@ export class R2Router extends Router<R2Gateway> {
   list: RouteHandler<Omit<R2Params, "key">> = async (req, params) => {
     const persist = decodePersist(req.headers);
     const gateway = this.gatewayFactory.get(params.bucket, persist);
-    await gateway.list();
-    return new Response();
+
+    const json = JSON.stringify(await gateway.list());
+    const utf8 = new TextEncoder().encode(json);
+    const byteSize = utf8.length;
```
mrbbot, line 61: Nit:
```suggestion
    const byteSize = Buffer.byteLength(json);
```

packages/tre/src/plugins/r2/router.ts
```diff
@@ -44,7 +55,16 @@ export class R2Router extends Router<R2Gateway> {
...
+
+    return new Response(json, {
+      headers: {
+        "CF-R2-Metadata-Size": `${byteSize}`,
```
mrbbot, line 65: Nit: use the extracted `CF-R2-Metadata-Size` constant here. 👍

packages/tre/src/plugins/r2/router.ts
```diff
@@ -16,20 +17,30 @@ export interface R2Params {
...
+  put: RouteHandler<Omit<R2Params, "key">> = async (req, params) => {
+    const bytes = await req.arrayBuffer();
+
+    const metadataSize = Number(req.headers.get("cf-r2-metadata-size"));
+
+    const [metadataBytes, value] = [
+      bytes.slice(0, metadataSize),
+      bytes.slice(metadataSize),
+    ];
+    const metadata = JSON.parse(new TextDecoder().decode(metadataBytes));
```
mrbbot, line 38: Nit: could you lift `new TextDecoder()` out of this function so we're not creating a new one on each request?

packages/tre/src/plugins/r2/router.ts
```diff
@@ -16,20 +17,30 @@ export interface R2Params {
 export class R2Router extends Router<R2Gateway> {
   @GET("/:bucket/:key")
   get: RouteHandler<R2Params> = async (req, params) => {
-    // console.log(await req.json());
+    console.log("GET KEY");
 
     const persist = decodePersist(req.headers);
     const gateway = this.gatewayFactory.get(params.bucket, persist);
     await gateway.get(params.key);
     return new Response();
```
mrbbot, line 25: Shouldn't the result of `gateway.get()` be returned in the response?

mrbbot (commented):
Hey! 👋 Thanks again for getting to this so quickly. 🙂 I've added a few comments, including some clarification on what I think should be in `R2Gateway`. Let me know if you have any questions! 🙂

penalosa:
Apologies @mrbbot! This wasn't quite ready for review—it should be now. I'll go through and have a check of your comments and fix any that are still relevant with the new code

packages/tre/src/plugins/r2/gateway.ts
```diff
@@ -1,24 +1,537 @@
...
+const UNPAIRED_SURROGATE_PAIR_REGEX =
+  /^(?:[^\uD800-\uDBFF]|^)[\uDC00-\uDFFF]|[\uD800-\uDBFF](?![\uDC00-\uDFFF])$/;
+
+const encoder = new TextEncoder();
+
+type Method = "HEAD" | "GET" | "PUT" | "LIST" | "DELETE";
+
+function throwR2Error(method: Method, status: number, message: string): void {
+  throw new Error(`R2 ${method} failed: (${status}) ${message}`);
+}
```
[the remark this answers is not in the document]
penalosa: Fixed (kinda) in the latest push. I have a subclass of `Error` `export class R2Error extends Error`, which encodes the status code & CF error code (I didn't actually think of subclassing HttpError, primarily because each error needs explicitly caught and converted to a response, as the runtime needs each error formatted fairly specifically)

packages/tre/src/plugins/r2/router.ts
```diff
@@ -1,50 +1,188 @@
...
+const decoder = new TextDecoder();
 
-    const persist = decodePersist(req.headers);
-    const gateway = this.gatewayFactory.get(params.bucket, persist);
-    await gateway.get(params.key);
-    return new Response();
+const decodeMetadata = async (req: Request) => {
+  const bytes = await req.arrayBuffer();
+
+  const metadataSize = Number(req.headers.get(CfHeader.MetadataSize));
```
mrbbot, line 26: Should this be returning a `400` response if the header isn't a number?

packages/tre/src/plugins/r2/router.ts
```diff
@@ -1,50 +1,188 @@
...
+
+  const [metadataBytes, value] = [
+    bytes.slice(0, metadataSize),
+    bytes.slice(metadataSize),
+  ];
+  const metadata = JSON.parse(decoder.decode(metadataBytes));
+  return { metadata, value: new Uint8Array(value) };
+};
+const decodeHeaderMetadata = (req: Request) => {
+  const metadata = JSON.parse(req.headers.get(CfHeader.Request) as string);
```
mrbbot, line 36: Similarly, should this be returning a `400` response if the header isn't defined?

packages/tre/src/plugins/r2/r2Object.ts
```diff
@@ -0,0 +1,140 @@
...
+  name: string;
+  // Various HTTP headers associated with the object. Refer to HTTP Metadata.
+  httpFields: R2HTTPMetadata;
+  // A map of custom, user-defined metadata associated with the object.
+  customFields: { k: string; v: string }[];
+}
+
+export function createVersion(): string {
+  const size = 32;
+  return crypto.randomBytes(size).toString("base64").slice(0, size);
```
mrbbot, line 49:
```suggestion
  return crypto.randomBytes(24).toString("base64");
```

packages/tre/src/plugins/r2/r2Object.ts
```diff
@@ -0,0 +1,140 @@
...
+      name: this.key,
+      httpFields: this.httpMetadata,
+      customFields: Object.entries(this.customMetadata).map(([k, v]) => ({
+        k,
+        v,
+      })),
+    };
+  }
+
+  encode(): { metadataSize: number; value: Uint8Array } {
```
mrbbot, line 103: Maybe extract this return type into a small `interface`, since you use it in `encodeMultiple` and `R2ObjectBody` too?

packages/tre/src/plugins/r2/router.ts
```diff
@@ -1,50 +1,188 @@
...
-  @GET("/:bucket/:key")
-  get: RouteHandler<R2Params> = async (req, params) => {
-    // console.log(await req.json());
+const decoder = new TextDecoder();
 
-    const persist = decodePersist(req.headers);
-    const gateway = this.gatewayFactory.get(params.bucket, persist);
-    await gateway.get(params.key);
-    return new Response();
+const decodeMetadata = async (req: Request) => {
```
mrbbot, line 23: Nit: for consistency with the `function`s below, could you use `function` instead of `const ... = () => {}` syntax?

packages/tre/src/plugins/r2/validator.ts
```diff
@@ -0,0 +1,267 @@
...
+  R2Conditional,
+  R2GetOptions,
+  R2ListOptions,
+  R2PutOptions,
+} from "./gateway";
+
+import { R2HTTPMetadata, R2Object, R2ObjectMetadata } from "./r2Object";
+const MAX_LIST_KEYS = 1_000;
+
+const MAX_KEY_SIZE = 1024;
```
mrbbot, line 21: Nit:
```suggestion
import { R2HTTPMetadata, R2Object, R2ObjectMetadata } from "./r2Object";

const MAX_LIST_KEYS = 1_000;
const MAX_KEY_SIZE = 1024;
```

packages/tre/src/plugins/r2/validator.ts
```diff
@@ -0,0 +1,267 @@
...
+
+const UNPAIRED_SURROGATE_PAIR_REGEX =
+  /^(?:[^\uD800-\uDBFF]|^)[\uDC00-\uDFFF]|[\uD800-\uDBFF](?![\uDC00-\uDFFF])$/;
+const MAX_VALUE_SIZE = 5 * 1_000 * 1_000 * 1_000 - 5 * 1_000 * 1_000;
+
+// false -> the condition testing "failed"
+export function testR2Conditional(
```
mrbbot, line 28: Should this be `export`ed? Or only accessible via `Validator`?

packages/tre/src/plugins/r2/validator.ts
```diff
@@ -0,0 +1,267 @@
...
+      throw new BadDigest();
+    }
+    return md5Hash;
+  }
+  condition(meta: R2Object, onlyIf?: R2Conditional) {
+    // test conditional should it exist
+    if (!testR2Conditional(onlyIf, meta) || meta?.size === 0) {
+      throw new PreconditionFailed().attach(meta);
+    }
+  }
```
mrbbot, line 86: Should these 2 functions return `this` for consistency?

mrbbot (commented):
This looks great! 😃 Thanks again for doing this. I've added a few small comments. 👍

packages/tre/src/plugins/r2/router.ts
penalosa: I've added this as a safety check. The reason I didn't add it initially is because the runtime should unconditionally add this header with valid data (and if not, it's probably a bug in the runtime, rather than a code path that a user could trigger)

packages/tre/src/plugins/r2/router.ts
penalosa: As above

packages/tre/src/plugins/r2/validator.ts
penalosa: I've added it to `condition()`. The reason `md5()` returns a string is purely for convenience, as the return value is useful elsewhere (as the stored etag).

mrbbot: That makes sense, let's keep it like that. 👍

mrbbot (approved):

PR #363, opened 2022-09-08
Add Request.cf support, as well as splitting out `startServer` and `dispatchFetch`

penalosa:


packages/tre/src/index.ts
```diff
@@ -161,8 +174,136 @@ export class Miniflare {
...
+      console.log(
+        bold(
+          red(`Unable to fetch the \`Request.cf\` object! Falling back to a default placeholder...
+${dim(e.cause ? e.cause.stack : e.stack)}`)
+        )
+      );
+    }
+  }
+  // Initialise a proxy server that adds the `CF-Blob` header to runtime requests
+  #createServer(port: number): http.Server {
```
mrbbot, line 215: I think the `Miniflare` class is starting to get pretty big. 😅 Could you move all the proxy server and `cf` fetching stuff (including the new methods and constants in `helpers.ts`)  into a separate `ProxyServer` class in `src/proxy.ts` or something? We may want to move the loopback stuff into a separate file too, but we can do that later. 👍

packages/tre/src/index.ts
```diff
@@ -161,8 +174,136 @@ export class Miniflare {
...
+    const proxyWebSocketServer = new WebSocketServer({ noServer: true });
+    proxyWebSocketServer.on("connection", (client, request) => {
+      // Filter out browser WebSocket headers, since `ws` injects some. Leaving browser ones in causes key mismatches, especially with `Sec-WebSocket-Accept`
+      const safeHeaders = filterWebSocketHeaders(request.headers);
```
mrbbot, line 219: Rather than creating a new headers object here, I would probably just mutate the incoming `request.headers` with `delete request.headers["sec-websocket-version"]` etc, since we now have ownership of `request`. Similarly, I would add the `CF-Blob` header onto `request.headers` and pass that directly to `new WebSocket()`.

packages/tre/src/index.ts
```diff
@@ -161,8 +174,136 @@ export class Miniflare {
...
+
+      const runtime = new WebSocket(`ws://127.0.0.1:${port}${request.url}`, {
```
mrbbot, line 221: Suggestion for using `this.#runtimeEntryURL` directly:

Outside this closure:
```js
await this.#initPromise;
const wsRuntimeEntryURL = new URL(this.#runtimeEntryURL!); 
wsRuntimeEntryURL.protocol = "ws";
```
...then (assuming `new WebSocket` takes a `URL`):
```suggestion
      const runtime = new WebSocket(new URL(request.url, wsRuntimeEntryURL), {
```

packages/tre/src/index.ts
```diff
@@ -161,8 +174,136 @@ export class Miniflare {
...
+      runtime.addEventListener("message", (e: MessageEvent) => {
+        client.send(e.data);
+      });
+      runtime.addEventListener("close", () => client.close());
+    });
+    const server = http.createServer((originalRequest, originalResponse) => {
+      const proxyToRuntime = http.request(
+        {
+          hostname: "127.0.0.1",
+          port,
```
mrbbot, line 240: Suggestion for using this.#runtimeEntryURL directly:

```suggestion
          hostname: this.#runtimeEntryURL!.hostname,
          port: Number(this.#runtimeEntryURL!.port),
```

packages/tre/src/index.ts
```diff
@@ -161,8 +174,136 @@ export class Miniflare {
...
+      );
+      client.addEventListener("close", () => runtime.close());
+
+      runtime.addEventListener("message", (e: MessageEvent) => {
+        client.send(e.data);
+      });
+      runtime.addEventListener("close", () => client.close());
+    });
+    const server = http.createServer((originalRequest, originalResponse) => {
+      const proxyToRuntime = http.request(
```
mrbbot, line 237: Nit: for consistency with above parameters:
```suggestion
      const runtimeRequest = http.request(
```

packages/tre/src/index.ts
```diff
@@ -161,8 +174,136 @@ export class Miniflare {
...
+        {
+          hostname: "127.0.0.1",
+          port,
+          path: originalRequest.url,
+          method: originalRequest.method,
+          headers: injectCfHeaders(originalRequest.headers, this.#cf),
+        },
+        (runtime) => {
```
mrbbot, line 245: Nit: for consistency with above parameters:
```suggestion
        (runtimeResponse) => {
```

packages/tre/src/index.ts
```diff
@@ -161,8 +174,136 @@ export class Miniflare {
...
+      // Proxy messages to and from the runtime
+      client.addEventListener("message", (e: MessageEvent) =>
+        runtime.send(e.data)
+      );
+      client.addEventListener("close", () => runtime.close());
+
+      runtime.addEventListener("message", (e: MessageEvent) => {
+        client.send(e.data);
+      });
+      runtime.addEventListener("close", () => client.close());
```
mrbbot, line 234: You'll want to pass on the [WebSocket close code](https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent/code)? See here (ignoring `ws.readyState` check): https://github.com/cloudflare/miniflare/blob/ddc45f3385516df1cf7c09fcf2c18682192e6733/packages/web-sockets/src/couple.ts#L48-L52

packages/tre/src/index.ts
```diff
@@ -161,8 +174,136 @@ export class Miniflare {
...
+        headers: injectCfHeaders(init?.headers ?? {}, this.#cf),
+      });
+    } else {
+      url = new URL(input.url);
+      url.host = (this.#runtimeEntryURL as URL).host;
+      forward = new Request(url, {
+        ...(input as RequestInit),
+        headers: injectCfHeaders(input?.headers ?? {}, this.#cf),
+      });
+    }
```
mrbbot, line 302: How about something simpler like this?
```suggestion
    const request = new Request(input, init);
    request.headers.set(CfHeader.Blob, JSON.stringify(this.#cf));
    return fetch(request);
```

packages/tre/src/index.ts
```diff
@@ -161,8 +174,136 @@ export class Miniflare {
...
+    return await fetch(url, forward as RequestInit);
+  }
   async #init() {
+    await this.#setupCf();
```
mrbbot, line 306: Could we call this function here, but move the `await` of the return `Promise` to just before `await this.#waitForRuntime()` at the end of this function? That way we can download the `cf.json` object in the background whilst the runtime is starting up.

mrbbot (commented):
Looking good! 🙂 Added some comments to simplify the implementation a little. If you could add support for the `cfFetch` option as well that be great. 👍

packages/tre/src/index.ts
```diff
@@ -161,8 +174,136 @@ export class Miniflare {
...
+      });
+    } else {
+      url = new URL(input.url);
+      url.host = (this.#runtimeEntryURL as URL).host;
+      forward = new Request(url, {
+        ...(input as RequestInit),
+        headers: injectCfHeaders(input?.headers ?? {}, this.#cf),
+      });
+    }
+    return await fetch(url, forward as RequestInit);
```
mrbbot, line 303: I think there are a few cases this doesn't handle (e.g. `input` is a `Request` and `init` is also specified). I remember having some problems with `undici` and spreading `Request` objects in the past too, so maybe do something like this? (`_urlFromRequestInput` is from `@miniflare/core`)

```suggestion
    await this.#initPromise;
    const url = _urlFromRequestInput(input);
    url.host = this.#runtimeEntryURL!.host;
    if (init !== undefined) {
      init = new Request(input, init)   
    } else if (input instanceof Request) {
      init = input; 
    }
    const request = new Request(url, init);
    request.headers.set(CfHeader.Blob, JSON.stringify(this.#cf));
    return fetch(request);
```

packages/tre/src/index.ts
penalosa: Refactored into `proxy.ts`, which exports a class that Miniflare initialises and delegates to for relevant methods

packages/tre/src/index.ts
penalosa: What about:
```ts
const url = new URL(request.url!);
      url.host = this.runtimeURL!.host;
      const runtime = new WebSocket(url, {
        headers: request.headers,
      });
```

mrbbot: I think `request.url` is a relative path, so `new URL()` will complain it isn't an absolute fully-qualified URL. 🙁 `url` won't have the `ws://` protocol either.

packages/tre/src/index.ts
penalosa: I wrote it like that primarily to try and avoid the overhead of constructing two `Request` objects. Having thought about it a bit more though, I think there's a simpler way:
```ts
const forward = new Request(input, init);
forward.headers.set(CfHeader.Blob, JSON.stringify(this.#cf));
const url = new URL(forward.url);
url.host = this.runtimeURL!.host;
return await fetch(url, forward as RequestInit);
```

mrbbot: Oooo that's really nice! 😄 Do you need the `as RequestInit`?

package.json
```diff
@@ -54,5 +54,8 @@
   },
   "volta": {
     "node": "18.2.0"
+  },
+  "dependencies": {
+    "undici": "^5.10.0"
   }
```
mrbbot, line 60: I'm not sure `undici` should be included here. This `package.json` is only used for shared scripts for repository management.

packages/tre/src/helpers.ts
```diff
@@ -1,4 +1,13 @@
+import assert from "assert";
+import { IncomingHttpHeaders } from "http";
+import path from "path";
+import {
+  Headers,
+  HeadersInit,
+  IncomingRequestCfProperties,
+} from "@miniflare/core";
 import { z } from "zod";
+import { CfHeader } from "./plugins/shared/constants";
```
mrbbot, line 10: Guessing these imports can be removed now that you're not adding anything to this file?

packages/tre/src/index.ts
```diff
@@ -122,6 +124,8 @@ export class Miniflare {
 
   #updatePromise?: Promise<void>;
 
+  #proxyServer?: ProxyServer;
```
mrbbot, line 127: Does `#proxyServer` need to be optional? You're defining it in the constructing, and making it required would remove the need to non-null assertions later on.

packages/tre/src/proxy.ts
```diff
@@ -0,0 +1,212 @@
...
+  // Initialise a proxy server that adds the `CF-Blob` header to runtime requests
+  createServer() {
+    const proxyWebSocketServer = new WebSocketServer({ noServer: true });
+    proxyWebSocketServer.on("connection", (client, request) => {
+      // Filter out browser WebSocket headers, since `ws` injects some. Leaving browser ones in causes key mismatches, especially with `Sec-WebSocket-Accept`
+      delete request.headers["sec-websocket-version"];
+      delete request.headers["sec-websocket-key"];
+      delete request.headers["sec-websocket-extensions"];
+      request.headers[CfHeader.Blob] = JSON.stringify(this.#cf);
+      console.log(request.url);
```
mrbbot, line 120:
```suggestion
```

packages/tre/src/proxy.ts
```diff
@@ -0,0 +1,212 @@
...
+  async dispatchFetch(
+    input: RequestInfo,
+    init?: RequestInit
+  ): Promise<Response> {
+    const forward = new Request(input, init);
+    forward.headers.set(CfHeader.Blob, JSON.stringify(this.#cf));
+    const url = new URL(forward.url);
+    url.host = this.runtimeURL!.host;
+    return await fetch(url, forward as RequestInit);
+  }
```
mrbbot, line 202: Nit: this function doesn't need to be marked `async`, and you can remove the `await` before `fetch`. This will prevent a few unnecessary `Promise` allocations. 🙂

packages/tre/src/proxy.ts
```diff
@@ -0,0 +1,212 @@
...
+  async dispatchFetch(
+    input: RequestInfo,
+    init?: RequestInit
+  ): Promise<Response> {
+    const forward = new Request(input, init);
+    forward.headers.set(CfHeader.Blob, JSON.stringify(this.#cf));
+    const url = new URL(forward.url);
+    url.host = this.runtimeURL!.host;
+    return await fetch(url, forward as RequestInit);
```
mrbbot, line 201: Nit: can you remove `as RequestInit` here. Doesn't `Request` satisfy that `interface` anyways?

mrbbot (commented):
This is looking really clean! A few minor comments, then let's get this merged! 🙂

packages/tre/src/proxy.ts
penalosa: It does, which is why I'm okay with the casting with no checks. Weirdly though, it's not typed as satisfying that interface:
Screenshot 2022-09-12 at 10 30 33

I don't know if this is a weirdness with `undici` versioning or something, but `RequestInit` has a type for `referrerPolicy` of:
```ts
export type ReferrerPolicy =
  | ''
  | 'no-referrer'
  | 'no-referrer-when-downgrade'
  | 'origin'
  | 'origin-when-cross-origin'
  | 'same-origin'
  | 'strict-origin'
  | 'strict-origin-when-cross-origin'
  | 'unsafe-url';
```
while `Request` has a type of `string`, and so they're not quite compatible

packages/tre/src/proxy.ts
```diff
@@ -0,0 +1,207 @@
...
+      );
+
+      originalRequest.pipe(runtimeRequest);
+    });
+    // Handle websocket requests
+    server.on("upgrade", (request, socket, head) => {
+      proxyWebSocketServer.handleUpgrade(request, socket, head, (ws) => {
+        proxyWebSocketServer.emit("connection", ws, request);
+      });
+    });
```
mrbbot, line 176: I've recently been improving WebSocket compatibility in Miniflare 2 and realised there are a few cases where this doesn't work:
- If the worker includes extra headers in the WebSocket upgrade response, these aren't passed onto the client
- If the client sends a `Sec-WebSocket-Protocol: protocol1, protocol2, ...` header, and the worker selects a protocol (e.g. `protocol2`), `protocol1` will always be sent back, and then the server will crash. 🙁 
- If the client sends a message as soon as it receives the `open` event, it's possible the runtime socket may not be fully open yet, so attempting to send a message crashes the server.

Rather than trying to fix all these issues using `ws`, I think it might be better if we try to do as little proxying work as possible, and passthrough most of the request/response untouched.

I had a look at what the `http-proxy` module did here, and it's actually not too complicated. Here's a stripped down version that seems to work (you may want to lift `writeHead` out into the global scope). We probably don't need to depend on `ws` anymore either. 👍 
```suggestion
    server.on("upgrade", (originalRequest, originalSocket, originalHead) => {
      // Based on `http-proxy` (MIT Licensed):
      // https://github.com/http-party/node-http-proxy/blob/9b96cd725127a024dabebec6c7ea8c807272223d/lib/http-proxy/passes/ws-incoming.js#L79

      if (originalHead?.byteLength) originalSocket.unshift(originalHead);

      function writeHead(
        socket: stream.Writable,
        version: string,
        statusCode: number | undefined,
        statusMessage: string | undefined,
        headers: http.OutgoingHttpHeaders
      ) {
        socket.write(`HTTP/${version} ${statusCode} ${statusMessage}\r\n`);
        for (const [key, values] of Object.entries(headers)) {
          if (Array.isArray(values)) {
            for (const value of values) socket.write(`${key}: ${value}\r\n`);
          } else {
            socket.write(`${key}: ${values}\r\n`);
          }
        }
        socket.write("\r\n");
      }

      const runtimeRequest = http.request(
        {
          hostname: this.runtimeURL!.hostname,
          port: this.runtimeURL!.port,
          path: originalRequest.url,
          method: originalRequest.method,
          headers: originalRequest.headers,
        },
        (runtimeResponse) => {
          if (!(runtimeResponse as { upgrade?: boolean }).upgrade) {
            writeHead(
              originalSocket,
              runtimeResponse.httpVersion,
              runtimeResponse.statusCode,
              runtimeResponse.statusMessage,
              runtimeResponse.headers
            );
            runtimeResponse.pipe(originalSocket);
          }
        }
      );
      runtimeRequest.on(
        "upgrade",
        (runtimeResponse, runtimeSocket, runtimeHead) => {
          if (runtimeHead?.byteLength) runtimeSocket.unshift(runtimeHead);
          writeHead(
            originalSocket,
            "1.1",
            101,
            "Switching Protocols",
            runtimeResponse.headers
          );
          runtimeSocket.pipe(originalSocket).pipe(runtimeSocket);
        }
      );

      originalRequest.pipe(runtimeRequest);
    });
```

penalosa: Ah, that's a good point—proxying less is definitely a good idea. I had been testing with a browser websocket, which doesn't expose much in the way of complexity (no access to headers etc...), so I hadn't run across these edge cases, but I suppose we should support arbitrary clients rather than just browser ones.

I can definitely switch over to using the `http-proxy` approach, but alternatively (and to get around any further edge cases that we might run across), what about stepping down the protocol stack slightly, and proxying the TCP socket instead? All we really need to change about the request is the `cf-blob` header, and we might be better off injecting that directly rather than going through Node's http stack?

```ts
function injectCfHeader(cf: IncomingRequestCfProperties) {
      return async function* (source: AsyncIterable<Buffer>) {
        for await (const chunk of source) {
          if (chunk.includes("\r\n\r\n")) {
            const index = chunk.indexOf("\r\n\r\n");
            const pre = Buffer.from(chunk.slice(0, index));
            const post = Buffer.from(chunk.slice(index));
            const header = Buffer.from(
              `\r\n${CfHeader.Blob}: ${JSON.stringify(cf)}`
            );
            yield Buffer.concat([pre, header, post]);
          } else {
            yield chunk;
          }
        }
      };
    }
    const tcpServer = net.createServer((conn) => {
      const runtime = net.createConnection(
        Number(this.runtimeURL!.port),
        this.runtimeURL!.hostname
      );
      conn.pipe(compose(injectCfHeader(this.#cf))).pipe(runtime);
      conn.on("end", () => runtime.end());
      conn.on("error", () => runtime.end());
      runtime.on("error", () => conn.end());
      runtime.on("end", () => conn.end());
      runtime.pipe(conn);
    });
```

mrbbot: 😅 I _really_ like this idea! However, I'm a bit concerned about this method for "end-of-headers" detection. Is it technically _possible_ for the `\r\n\r\n` to be split over multiple chunks? Also, more likely, could `\r\n\r\n` ever appear in the body too? It feels like to handle all these cases properly, we'd end up needing to reimplement a streaming HTTP parser, which is basically what we're getting from Node's `http` module. There may also be cases in the future where we want to rewrite responses in Miniflare (e.g. pretty errors, live reload), so leaving that option open with the `http` approach would be good (although we should probably move most of this to wrangler).

mrbbot: I'm not too concerned with the differing Node `http` and runtime semantics. For the somewhat silly example, the user is performing an action with undefined behaviour so they probably shouldn't be depending on an exact response. Also, most users will be accessing Miniflare 3 through Wrangler, which will definitely be following Node's `http` semantics, so us following them here shouldn't be a problem.

penalosa:
This has been drastically simplified following @kentonv's suggestions to mutate `request.cf` from the core entry worker rather than a proxy server

packages/tre/src/plugins/core/index.ts
```diff
@@ -69,7 +70,14 @@ const BINDING_TEXT_CUSTOM_SERVICE = "MINIFLARE_CUSTOM_SERVICE";
 // TODO: is there a way of capturing the full stack trace somehow?
 // Using `>=` for version check to handle multiple `setOptions` calls before
 // reload complete.
-export const SCRIPT_ENTRY = `addEventListener("fetch", (event) => {
+export const getScriptEntry = (
+  cf: IncomingRequestCfProperties
+) => `addEventListener("fetch", (event) => {
+  const request = new Request(event.request)
+  const cf = ${JSON.stringify(cf)}
```
mrbbot, line 77: Rather than using a function here, could you set `cf` as a JSON-type binding?

packages/tre/src/plugins/core/index.ts
```diff
@@ -69,7 +70,14 @@ const BINDING_TEXT_CUSTOM_SERVICE = "MINIFLARE_CUSTOM_SERVICE";
...
+  Object.entries(cf).forEach(([k,v]) => {
+    request.cf[k] = v
+  })
```
mrbbot, line 80: Could you pass the `cf` object in the `RequestInit` when constructing a `Request` instead? Something like `new Request(event.request, { cf: ${BINDING_JSON_CF_BLOB} })`.

mrbbot (commented):
Nice! That's way nicer! 😃 Added a couple suggestions. 👍

packages/tre/src/cf.ts
```diff
@@ -0,0 +1,106 @@
...
+        )
+      );
+    }
+  }
+  get ready() {
+    return this.#initPromise;
+  }
+  config() {
+    return this.#cf;
+  }
```
mrbbot, line 105: Given `#assembleConfig` is `async`, could we simplify this to `get config(): Promise<any>`, where `config` returns a `Promise` that resolves with the fetched config?

packages/tre/src/cf.ts
```diff
@@ -0,0 +1,106 @@
...
+export const CF_DAYS = 30;
+
+type CoreOptions = OptionalZodTypeOf<Plugins["core"]["sharedOptions"]>;
+export class CfFetcher {
+  #options: CoreOptions;
+  #cf = fallbackCf;
+
+  readonly #initPromise: Promise<void>;
+
+  constructor(options: CoreOptions) {
```
mrbbot, line 53: Given this is only using the `cfFetch` option, could we just take that instead of the full `CoreOptions`?

packages/tre/src/plugins/core/index.ts
```diff
@@ -46,6 +45,8 @@ export const CoreOptionsSchema = z.object({
 export const CoreSharedOptionsSchema = z.object({
   host: z.string().optional(),
   port: z.number().optional(),
+  cfFetch: z.union([z.boolean(), z.string()]).optional(),
+  cf: z.object({}).optional(),
```
mrbbot, line 49: Could we merge these options, so that if `cfFetch` is an object, we just use that instead?

packages/tre/src/index.ts
```diff
@@ -146,6 +146,7 @@ export class Miniflare {
     );
 
     this.#disposeController = new AbortController();
+    this.#cfFetcher = new CfFetcher(sharedOpts.core);
```
mrbbot, line 149: Currently, I think this will fetch from `workers.cloudflare.com/cf.json`, even if the user has set the `cf` option. We probably don't want this. You could only construct this if `cf` wasn't set, but you'd have to be careful of `setOptions` potentially unsetting this value.

packages/tre/src/cf.ts
mrbbot: Actually, how about making the `CfFetcher` class an `async function` instead that returns the config? Then call it and store the `Promise` only if `cfFetch` is `undefined` or `true`?

packages/tre/src/plugins/core/index.ts
penalosa: Unified to `cf: boolean | string | object`

packages/tre/src/index.ts
penalosa: I simplified the logic here somewhat. In essence:
cf: true | undefined=fetch from the remote, falling back to fallback CF
cf: false=use fallback CF
cf: string=use cf at path, falling back to remote, falling back to fallback CF
cf: object=use as CF object

packages/tre/src/plugins/core/index.ts
```diff
@@ -42,11 +41,59 @@ export const CoreOptionsSchema = z.object({
...
+  metroCode: z.string().optional(),
+  postalCode: z.string().optional(),
+  region: z.string().optional(),
+  regionCode: z.string().optional(),
+  requestPriority: z.string(),
+  timezone: z.string().optional(),
+  tlsVersion: z.string(),
+  tlsCipher: z.string(),
+  tlsClientAuth: IncomingRequestCfPropertiesTLSClientAuthSchema,
+});
```
mrbbot, line 85: I'm hesitant to validate these options. These options seem to change semi-regularly, and for testing, users may only want to specify the subset of them they actually use. Maybe let's validate them, but make them all optional?

packages/tre/src/plugins/core/index.ts
```diff
@@ -158,20 +202,20 @@ export const CORE_PLUGIN: Plugin<
     workerIndex,
     sharedOptions,
   }) {
+    console.log(sharedOptions.cf);
```
mrbbot, line 205:
```suggestion
```

packages/tre/src/plugins/core/index.ts
penalosa: I've made them all optional. Also note `IncomingRequestCfPropertiesSchema.passthrough()`, which lets any extra properties be specified.

PR #376, opened 2022-09-14
Reuse Docker container between reloads, closes #369

### LINKED ISSUE #369: Reuse Docker container between reloads

mrbbot:
Instead of creating a new container whenever we change the config, we now start a single container with a custom entrypoint script. On receiving `SIGUSR1`, this script restarts the runtime process. This change reduces reload time from ~1.5s to ~400ms on my machine.

### DIFF
--- a/packages/tre/src/index.ts
+++ b/packages/tre/src/index.ts
@@ -1,6 +1,7 @@
 import assert from "assert";
 import http from "http";
 import path from "path";
+import exitHook from "exit-hook";
 import getPort from "get-port";
 import { bold, green, grey } from "kleur/colors";
 import stoppable from "stoppable";
@@ -99,6 +100,7 @@ type PluginRouters = {
   [Key in keyof Plugins]: OptionalInstanceType<Plugins[Key]["router"]>;
 };
 
+// `__dirname` relative to bundled output `dist/src/index.js`
 const RUNTIME_PATH = path.resolve(__dirname, "..", "..", "lib", "cfwrkr");
 
 type StoppableServer = http.Server & stoppable.WithStop;
@@ -114,6 +116,7 @@ export class Miniflare {
 
   readonly #runtimeConstructor: RuntimeConstructor;
   #runtime?: Runtime;
+  #removeRuntimeExitHook?: () => void;
   #runtimeEntryURL?: URL;
 
   readonly #disposeController: AbortController;
@@ -181,6 +184,8 @@ export class Miniflare {
       entryPort,
       loopbackPort
     );
+    this.#removeRuntimeExitHook = exitHook(() => void this.#runtime?.dispose());
+
     this.#runtimeEntryURL = new URL(`http://127.0.0.1:${entryPort}`);
 
     const config = await this.#initialConfigPromise;
@@ -410,6 +415,7 @@ export class Miniflare {
     this.#disposeController.abort();
     await this.#initPromise;
     await this.#updatePromise;
+    this.#removeRuntimeExitHook?.();
     this.#runtime?.dispose();
     await this.#stopLoopbackServer();
   }
--- a/packages/tre/src/runtime/index.ts
+++ b/packages/tre/src/runtime/index.ts
@@ -102,7 +102,7 @@ class NativeRuntime extends Runtime {
     runtimeProcess.stdin.end();
   }
 
-  dispose() {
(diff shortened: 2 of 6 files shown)


packages/tre/src/index.ts
```diff
@@ -181,6 +184,8 @@ export class Miniflare {
       entryPort,
       loopbackPort
     );
+    this.#removeRuntimeExitHook = exitHook(() => void this.#runtime?.dispose());
```
penalosa, line 187: What about calling `this.dispose()`?

packages/tre/src/runtime/index.ts
```diff
@@ -21,14 +31,16 @@ export interface RuntimeConstructor {
 }
 
 const COMMON_RUNTIME_ARGS = ["serve", "--binary", "--verbose"];
```
penalosa, line 33: Unrelated to this PR, but should we expose `--verbose` as a miniflare option to control this?

packages/tre/src/runtime/index.ts
```diff
@@ -40,13 +52,14 @@ class NativeRuntime implements Runtime {
...
 
   constructor(
-    protected readonly runtimeBinaryPath: string,
-    protected readonly entryPort: number,
-    protected readonly loopbackPort: number
+    runtimeBinaryPath: string,
+    entryPort: number,
+    loopbackPort: number
   ) {
+    super(runtimeBinaryPath, entryPort, loopbackPort);
```
penalosa, line 62: Is this super call needed? `Runtime` doesn't do anything with its arguments, right?

packages/tre/src/runtime/index.ts
```diff
@@ -86,11 +99,12 @@ class NativeRuntime implements Runtime {
 
     // 3. Write config
     runtimeProcess.stdin.write(configBuffer);
+    runtimeProcess.stdin.end();
   }
 
-  async dispose() {
+  dispose() {
```
penalosa, line 105: I know the `async` annotation isn't needed because you're directly returning a promise. It does make it clearer at a glance though that the function is async, and so I wonder explicitly annotating the return type here could serve that same function from a readability perspective?

packages/tre/src/runtime/index.ts
```diff
@@ -21,14 +31,16 @@ export interface RuntimeConstructor {
 }
 
 const COMMON_RUNTIME_ARGS = ["serve", "--binary", "--verbose"];
+// `__dirname` relative to bundled output `dist/src/index.js`
+const RESTART_PATH = path.resolve(__dirname, "..", "..", "lib", "restart.sh");
 
-function waitForExit(process: childProcess.ChildProcess): Promise<number> {
+function waitForExit(process: childProcess.ChildProcess): Promise<void> {
```
penalosa, line 37: Why the change from `void` to `number`?

penalosa (changes requested):

packages/tre/src/index.ts
mrbbot: Hmmm, a potential issue with that is that `exitHook`s need to be synchronous, as there's no guarantee Node will wait for async tasks to finish. The only thing that really needs to be tidied up is shutting down the runtime, all other proxy/loopback servers will be shutdown by Node automatically. To be fair though, the runtime should be attached as a child process, and should be terminated when Node terminates. There have just been a few times where I've observed it doesn't.

packages/tre/src/runtime/index.ts
mrbbot: For sure, eventually I'd like to remove this flag by default. Currently, it's the only way afaik to see `console.log`s though.

packages/tre/src/runtime/index.ts
mrbbot: Constructors for subclasses always need a `super` call, and the `Runtime` base constructor includes `protected readonly` modifiers on each the parameters meaning they'll be added as fields. Calling `super` is required to initialise them.

packages/tre/src/runtime/index.ts
mrbbot: This allows us to return `#processExitPromise` directly from `dispose()` without an `await`. We weren't actually using the exit code anywhere, though perhaps we should be logging it if `verbose` is enabled?

penalosa (approved):

PR #377, opened 2022-09-14
"Fix" capnproto serialisation of `undefined` values, closes #371

### LINKED ISSUE #371: Fix config serialisation errors with missing properties

mrbbot:
Previously passing `compatibilityDate: undefined` or `compatibilityFlags: undefined` would throw.

This is because the `undefined`s would make it to the `encodeCapnpStruct` function:
https://github.com/cloudflare/miniflare/blob/b0e1d06d32e1f6ff8f809d3b8180e5f17494732a/packages/tre/src/plugins/core/index.ts#L172-L173

`encodeCapnpStruct` distinguishes between `{ compatibilityDate: undefined }` and `{}`. Calling `tre/src/runtime/config/sserve-conf.capnp.js/Worker#setCompatibilityDate(value)` with an `undefined` value throws.

The problem is that `undefined` may be needed in some cases to signal that a function should be called, e.g. `sserve-conf.capnp.js/Worker_Binding_Type#setR2Bucket()`.

This "fix" checks if the function actually accepts a `value` before calling it. Because we don't have runtime type information, this is done by looking at the function's source code dynamically. 🤮 

Writing out this description has given me an idea for an alternative. We could replace the `undefined`s in `sserve-conf.ts/Worker_Binding_Type` with a special `Symbol` to signal that we wanted to call the function with no arguments. Then just ignore all `undefined` values in `encodeCapnpStruct`? What do you think?

### DIFF
--- a/packages/tre/src/runtime/config/index.ts
+++ b/packages/tre/src/runtime/config/index.ts
@@ -1,5 +1,5 @@
 import { Data, List, Message, Struct } from "capnp-ts";
-import { Config } from "./sserve-conf";
+import { Config, kVoid } from "./sserve-conf";
 import { Config as CapnpConfig } from "./sserve-conf.capnp.js";
 
 function capitalize<S extends string>(str: S): Capitalize<S> {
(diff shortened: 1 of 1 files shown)


penalosa:
> Rethinking this today, I not sure I want to ship this fix at the moment. Moving this PR to draft until I've got a better solution. 👍

There's definitely something about parsing JS strings that makes me uneasy...

penalosa (approved):
Looks good! My one concern is the typing of `obj: any`—is there any reason that's not able to be fully typed?

PR #383, opened 2022-09-20
Support `siteInclude`/`siteExclude` options and populate `__STATIC_CONTENT_MANIFEST`

mrbbot:
This PR adds support for the `siteInclude`/`siteExclude` options by adding additional logic to the `__STATIC_CONTENT` KV namespace service. It also populates `__STATIC_CONTENT_MANIFEST` using the same process recently added to Miniflare 2. We don't want to cache Workers Sites files, as when developing, the latest local file should always be returned.

### DIFF
--- a/packages/tre/src/plugins/kv/sites.ts
+++ b/packages/tre/src/plugins/kv/sites.ts
@@ -1,0 +1,211 @@
+import assert from "assert";
+import { pathToFileURL } from "url";
+import { FileStorage } from "@miniflare/storage-file";
+import {
+  MatcherRegExps,
+  deserialiseRegExps,
+  globsToRegExps,
+  serialiseRegExps,
+  testRegExps,
+} from "../../helpers";
+import { Service, Worker_Binding } from "../../runtime";
+import { SERVICE_LOOPBACK } from "../core";
+import {
+  BINDING_SERVICE_LOOPBACK,
+  BINDING_TEXT_PERSIST,
+  HEADER_PERSIST,
+  PARAM_FILE_UNSANITISE,
+} from "../shared";
+import { HEADER_SITES, KV_PLUGIN_NAME, PARAM_URL_ENCODED } from "./constants";
+
+export interface SitesOptions {
+  sitePath: string;
+  siteInclude?: string[];
+  siteExclude?: string[];
+}
+export interface SiteMatcherRegExps {
+  include?: MatcherRegExps;
+  exclude?: MatcherRegExps;
+}
+// Cache glob RegExps between `getBindings` and `getServices` calls
+const sitesRegExpsCache = new WeakMap<SitesOptions, SiteMatcherRegExps>();
+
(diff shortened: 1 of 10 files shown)


packages/tre/src/plugins/shared/gateway.ts
```diff
@@ -25,21 +30,39 @@ export class GatewayFactory<Gateway> {
...
+      if (url.protocol === "file:") {
+        const root = path.join(fileURLToPath(url), sanitisedNamespace);
+        const unsanitise =
+          url.searchParams.get(PARAM_FILE_UNSANITISE) === "true";
+        return new FileStorage(root, !unsanitise);
+      }
+      // TODO: support Redis/SQLite storages?
+      throw new MiniflareCoreError(
+        "ERR_PERSIST_UNSUPPORTED",
+        `Unsupported "${url.protocol}" persistence protocol for storage`
```
penalosa, line 56: Maybe give the full URL to aid debugging?
```suggestion
        `Unsupported "${url.protocol}" persistence protocol for storage (${url.href})`
```

packages/tre/src/plugins/shared/gateway.ts
```diff
@@ -25,21 +30,39 @@ export class GatewayFactory<Gateway> {
...
       this.#memoryStorages.set(namespace, (storage = new MemoryStorage()));
       return storage;
     }
 
-    const sanitised = sanitisePath(namespace);
+    // Sanitise namespace to make it file-system safe
+    const sanitisedNamespace = sanitisePath(namespace);
+
+    // If `persist` looks like a URL, parse it
+    if (persist !== true && URL_REGEXP.test(persist)) {
```
penalosa, line 45: Is the `URL_REGEXP` necessary? What about just catching the `TypeError` from `new URL()` to handle invalid URLs?

packages/tre/src/plugins/kv/sites.ts
```diff
@@ -0,0 +1,211 @@
...
+const siteRegExps = {
+  include: serialisedSiteRegExps.include && deserialiseRegExps(serialisedSiteRegExps.include),
+  exclude: serialisedSiteRegExps.exclude && deserialiseRegExps(serialisedSiteRegExps.exclude),
+};
+
+async function handleRequest(request) {
+  // Only permit reads
+  if (request.method !== "GET") {
+    const message = \`Cannot \${request.method.toLowerCase()}() with read-only Workers Sites namespace\`;
+    return new Response(message, { status: 400, statusText: message });
```
penalosa, line 88:
```suggestion
    return new Response(message, { status: 405, statusText: message });
```

penalosa (approved):
A couple tiny tiny things, but otherwise looks good. I'm not fully clear on what's happening end to end, and from a naive reading it feels like some of the regexp logic could be simplified (there seem to be a few functions that do similar things), but I'm probably wrong

PR #385, opened 2022-09-20
Feat: allow access to the Durable Object's instance from inside a test

cdrx:
cdaa62b added the ability to directly construct a Durable Object, which is great for testing a DO in isolation but doesn't allow for easily testing Durable Objects that communicate with other Durable Objects.

Without access to the internal instance that `stub.fetch(...)` uses, it isn't easy to test request that span across multiple Durable Objects.

For example, given this example of an object that makes a sub request to another object:

```
class TestObject extends DurableObject {
  fetch(request) {
    // make a subrequest to OtherObject
    const { OTHER_OBJECT } = getMiniflareBindings();
    const id = OTHER_OBJECT.idFromName("test");
    const stub = OTHER_OBJECT.get(id);
    return stub.fetch("https://object/");
  }
}

class OtherObject extends DurableObject {
  testFunction() {
    return "Hello world"
  }
  
  fetch(request) {
    return new Response(this.testFunction());
  }
}
```

The change in this PR allows writing a test that can introspect the behaviour of `OtherObject`, even though the initial request goes to `TestObject`, like this:

```
test("", async () => {
  const { TEST_OBJECT, OTHER_OBJECT } = getMiniflareBindings();
  
  const otherId = OTHER_OBJECT.idFromName("test");
  const otherInstance = getMiniflareDurableObjectInstance(otherId);
  const otherTestFunction = vi.spyOn(otherInstance, "testFunction");
  
  const testId = TEST_OBJECT.idFromName("test");
  await TEST_OBJECT.get(testId).fetch("https://object/");
  
  expect(otherTestFunction).toHaveBeenCalledOnce();
});
```


cdrx:
@mrbbot I'm curious about how the architecture will look, when using Miniflare with the OSS runtime.

When running tests, will it be the case that the Durable Object are run inside the OSS runtime (V8), but Miniflare still runs Jest inside Node's JS context?

packages/jest-environment-miniflare/test/fixtures/durableobjects.module.spec.js
```diff
@@ -43,3 +45,16 @@ test("Durable Objects direct", async () => {
...
 });
+
+test("Access to Durable Object instance", async () => {
+  const env = getMiniflareBindings();
+  const id = env.TEST_OBJECT.idFromName("test");
+  const stub = env.TEST_OBJECT.get(id);
+  const instance = await getMiniflareDurableObjectInstance(id);
+  const fetch = jest.spyOn(instance, "fetch");
+
+  await stub.fetch(new Request("https://object/"));
```
mrbbot, line 56: The problem here, is that `fetch` won't run under the input/output gates, so concurrent calls to `fetch` may produce invalid results. I'm not too sure what the best solution here is:

1. We could ask users to also do `const state = getMiniflareDurableObjectState()`, and use `runWithMiniflareDurableObjectGates()` as above, but reflecting on that API now, it feels a little too verbose.
2. We could wrap use a `Proxy` or something on `getMiniflareDurableObjectInstance()` to intercept `fetch()` calls and apply input/output gates to them, but this might break mocks/spies.

Maybe option 1. is fine for now. What do you think?

mrbbot (commented):
Hey! 👋 Thanks for raising this use case, and opening the PR! 🙂 I've added a comment about input/output gating. Interested to hear your thoughts, but otherwise happy to merge this. ✅

mrbbot (approved):

PR #391, opened 2022-09-26
In-Memory Durable Objects

### LINKED ISSUE #366: Add support for Durable Objects

mrbbot:
This PR adds support for non-durable Durable Objects. The open-source runtime does not yet support persisted objects. Data is lost when calling `setOptions()` or restarting. Closes #366.

### DIFF
--- a/packages/tre/src/index.ts
+++ b/packages/tre/src/index.ts
@@ -96,10 +96,12 @@ function getDurableObjectClassNames(
   const serviceClassNames = new Map<string, string[]>();
   for (const workerOpts of allWorkerOpts) {
     const workerServiceName = getUserServiceName(workerOpts.core.name);
-    for (const klass of Object.values(workerOpts.do.durableObjects ?? {})) {
+    for (const designator of Object.values(
+      workerOpts.do.durableObjects ?? {}
+    )) {
       // Fallback to current worker service if name not defined
       const [className, serviceName = workerServiceName] =
-        normaliseDurableObject(klass);
+        normaliseDurableObject(designator);
       let classNames = serviceClassNames.get(serviceName);
       if (classNames === undefined) {
         serviceClassNames.set(serviceName, (classNames = []));
--- a/packages/tre/src/index.ts
+++ b/packages/tre/src/index.ts
@@ -23,8 +23,9 @@ import {
   Plugins,
   SERVICE_ENTRY,
   SOCKET_ENTRY,
+  normaliseDurableObject,
 } from "./plugins";
-import { HEADER_CUSTOM_SERVICE } from "./plugins/core";
+import { HEADER_CUSTOM_SERVICE, getUserServiceName } from "./plugins/core";
 import {
   Config,
   Runtime,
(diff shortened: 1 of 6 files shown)


packages/tre/src/index.ts
```diff
@@ -23,8 +23,9 @@ import {
   Plugins,
   SERVICE_ENTRY,
   SOCKET_ENTRY,
+  normaliseDurableObject,
```
penalosa, line 26: Ssh, don't tell the Americans...

packages/tre/src/index.ts
```diff
@@ -84,6 +85,31 @@ function validateOptions(
...
+// (which would have to be recursive because of `export * from ...`), we collect
+// all Durable Object bindings, noting that bindings may be defined for objects
+// in other services.
+function getDurableObjectClassNames(
+  allWorkerOpts: PluginWorkerOptions[]
+): Map<string, string[]> {
+  const serviceClassNames = new Map<string, string[]>();
+  for (const workerOpts of allWorkerOpts) {
+    const workerServiceName = getUserServiceName(workerOpts.core.name);
+    for (const klass of Object.values(workerOpts.do.durableObjects ?? {})) {
```
penalosa, line 99: Maybe something more descriptive than `klass`? I know you can't use `class`, but maybe `durableObject`?

packages/tre/src/plugins/core/index.ts
```diff
@@ -89,12 +93,7 @@ export const SCRIPT_CUSTOM_SERVICE = `addEventListener("fetch", (event) => {
   event.respondWith(${BINDING_SERVICE_LOOPBACK}.fetch(request));
 })`;
 
-const now = new Date();
-const fallbackCompatibilityDate = [
-  now.getFullYear(),
-  (now.getMonth() + 1).toString().padStart(2, "0"),
-  now.getDate().toString().padStart(2, "0"),
-].join("-");
+const FALLBACK_COMPATIBILITY_DATE = "2000-01-01";
```
penalosa, line 96: Why are we shifting from today to a compat date of 2000? Surely keeping it up to date would be better?

packages/tre/src/plugins/do/index.ts
```diff
@@ -20,6 +26,20 @@ export const DurableObjectsSharedOptionsSchema = z.object({
   durableObjectsPersist: PersistenceSchema,
 });
 
+export function normaliseDurableObject(
+  klass: NonNullable<
```
penalosa, line 30: Change the name of `klass`

packages/tre/src/plugins/kv/index.ts
```diff
@@ -75,7 +75,6 @@ export const KV_PLUGIN: Plugin<
               service: { name: SERVICE_LOOPBACK },
             },
           ],
-          compatibilityDate: "2022-09-01",
         },
```
penalosa, line 78: Why remove this compat date?

penalosa (changes requested):
Couple changes, but otherwise looks good

packages/tre/src/index.ts
mrbbot: 😆

packages/tre/src/plugins/core/index.ts
mrbbot: The runtime has a maximum compatibility date it supports, which may lag behind the current date by a few days. If we specify a date after this, it will throw an error. I think the default compatibility date (at least for Miniflare) should be fixed in the past too, and shouldn't depend on the current time.

packages/tre/src/plugins/kv/index.ts
mrbbot: Because there's another one identical above. Must've been some incorrect rebasing... 😅

packages/tre/src/plugins/core/index.ts
penalosa: Sure, that makes sense. Do we not want something slightly more recent than 2000 though?

penalosa: At least after the last behaviour change, which looks like it was `2022-08-04` https://developers.cloudflare.com/workers/platform/compatibility-dates/

penalosa (approved):

PR #392, opened 2022-09-26
Remove obsolete Miniflare packages 😢 and depend on `workerd` from npm

mrbbot:
This PR removes all the obsolete Miniflare 2 code that isn't needed now we're using the open-source runtime. Storage classes and shared utilities have been moved into the `@miniflare/tre` package, which will eventually be renamed back to `miniflare`.


penalosa (approved):
Could probably do with some README changes, but lgtm!

penalosa (approved):

PR #394, opened 2022-09-27
fix: Add missing `@miniflare/r2` dependency to `shared-test-environment` package

askoufis:
I believe `@miniflare/r2` should be an explicit dependency of `shared-test-environment`. My guess is that it was accidentally _not_ added because NPM's hoisting resulted in no errors occurring.


mrbbot (approved):
Thank you! ✅

PR #395, opened 2022-09-27
Fallback to `workerd` compatibility date

mrbbot:
Previously, if the user set a compatibility date greater than the latest supported compatibility date of `workerd` but not in the future, `workerd` would error. Now, we log a warning, and fallback to the latest supported date instead.

Notably, `wrangler` sets the current date as the default compatibility date, so this was causing issues with the default setup.

### DIFF
--- a/packages/tre/src/plugins/core/index.ts
+++ b/packages/tre/src/plugins/core/index.ts
@@ -1,9 +1,16 @@
 import { readFileSync } from "fs";
 import fs from "fs/promises";
+import { bold, yellow } from "kleur/colors";
 import { Request, Response } from "undici";
 import { z } from "zod";
-import { Service, Worker_Binding, Worker_Module, kVoid } from "../../runtime";
-import { Awaitable, JsonSchema } from "../../shared";
+import {
+  Service,
+  Worker_Binding,
+  Worker_Module,
+  kVoid,
+  supportedCompatibilityDate,
+} from "../../runtime";
+import { Awaitable, JsonSchema, MiniflareCoreError } from "../../shared";
 import { BINDING_SERVICE_LOOPBACK, Plugin } from "../shared";
 import {
   ModuleDefinitionSchema,
@@ -12,6 +19,8 @@ import {
   STRING_SCRIPT_PATH,
   convertModuleDefinition,
 } from "./modules";
+
+const numericCompare = new Intl.Collator(undefined, { numeric: true }).compare;
 
 // (request: Request) => Awaitable<Response>
 export const ServiceFetch = z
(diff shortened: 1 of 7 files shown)


packages/tre/src/runtime/index.ts
```diff
@@ -1,14 +1,36 @@
...
+  const version = workerdPkg.version.split(".");
+  assert(version.length >= 3); // <major>.<max_compatibility_date>.<patch>
+
+  const compatDate = version[1];
+  const length = compatDate.length;
+  assert(length >= 6); // YYYYMMDD
+  const year = compatDate.substring(0, length - 4);
+  const month = compatDate.substring(length - 4, length - 2);
+  const date = compatDate.substring(length - 2, length);
+  return (workerdVersion = `${year}-${month}-${date}`);
```
penalosa, line 29: I wonder if it would be better to run `workerd` here? `workerd --version` should output `workerd YYYY-MM-DD`, which might be a better way of getting this?

packages/tre/src/plugins/core/index.ts
```diff
@@ -99,8 +108,45 @@ export const SCRIPT_CUSTOM_SERVICE = `addEventListener("fetch", (event) => {
...
+    );
+  } else if (numericCompare(compatibilityDate, workerdVersion) > 0) {
+    // If this compatibility date is greater than the maximum supported
+    // compatibility date of the runtime, but not in the future, warn,
+    // and use the maximum supported date instead
+    console.warn(
+      yellow(
+        [
+          "The newest compatibility date supported by the installed version of the Cloudflare Workers Runtime is ",
+          bold(`"${workerdVersion}"`),
```
penalosa, line 136: What about checking @GregBrimble's new compatibility date API here, to see if they're running a version of workerd that is out of date, and suggest updating wrangler if so?

packages/tre/src/runtime/index.ts
mrbbot: Yeah, I was thinking of that. But for the `DockerRuntime`, this would require us starting a container. We could do this, but given we're `import workerdPath from "workerd";`ing, so we're always getting the installed `workerd` version, (not another binary the user might have on their path), I think this is OK. Might be nice to have a named export in the `workerd` package though, since presumably you have the version when you're generating them?

`import workerdPath, { workerdVersion } from "workerd";`

packages/tre/src/plugins/core/index.ts
mrbbot: Yep, we could do that to check if an update was actually needed. Alternatively, we check the npm registry for an update to the `workerd` package as we're doing for Miniflare 2 at the moment. Or, we could move the `npx-import`ing from `wrangler` to `miniflare` to ensure we're always using the latest `workerd` version?

packages/tre/src/runtime/index.ts
mrbbot: That would be great! Thanks! :)

penalosa: `import path, { version, compatibilityDate } from 'workerd';`

packages/tre/src/plugins/core/index.ts
mrbbot: Actually, given `wrangler` already has an update check, I don't think we'll want a separate one for `workerd`. We should probably try setup something to automatically bump `workerd` in `miniflare` and `miniflare` in `wrangler` on a new `workerd` release.

penalosa (approved):

PR #396, opened 2022-09-28
Return runtime entry URL from `Miniflare#ready`

mrbbot:
This PR resolves the `Miniflare#ready` `Promise` with the runtime entry URL. This allows users to get the port that Miniflare actually allocated for the runtime (needed by `wrangler`'s `unstable_dev`), or make WebSocket requests using `ws` for instance (until we figure out a proper solution for this).

I've also removed `startServer()`, since it was just an alias for `ready`, doesn't actually start a server, and was going to be a breaking change anyway (as we no longer return an `http.Server` instance).

### DIFF
--- a/packages/tre/src/index.ts
+++ b/packages/tre/src/index.ts
@@ -438,8 +438,11 @@ export class Miniflare {
     return { services, sockets };
   }
 
-  get ready() {
-    return this.#initPromise;
+  get ready(): Promise<URL> {
+    // Cannot use async/await with getters.
+    // Safety: `#runtimeEntryURL` is assigned in `#init()`. `#initPromise`
+    // doesn't resolve until `#init()` returns.
+    return this.#initPromise.then(() => this.#runtimeEntryURL!);
   }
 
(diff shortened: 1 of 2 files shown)


packages/tre/src/index.ts
```diff
@@ -438,8 +441,11 @@ export class Miniflare {
     return { services, sockets };
   }
 
-  get ready() {
-    return this.#initPromise;
+  get ready(): Promise<URL> {
+    // Cannot use async/await with getters.
```
penalosa, line 445: Does this have to be a getter? What about `async ready()...`?

mrbbot: I guess not, though given we've already defined and released this as a getter, might be good to stick with it. If we were going to make it a method, maybe we should rename it to `waitForRuntime()`?

penalosa: Fair enough, yeah. In that case LGTM

penalosa (approved):

Issue #402, opened 2022-10-03
R2: when parsing etags, quotes should be stripped


vincentbernat:
Hey!

At https://github.com/cloudflare/miniflare/blob/932b537f8bce011b125cc1edf7a8b1d9254405ba/packages/r2/src/r2Object.ts#L173, `parseOnlyIf` should strip quotes from provided etags. HTTP etags are quoted, but S3/R2 etags are not.

PR #403, opened 2022-10-03
R2 bug fix: remove quotes in onlyIf headers

CraigglesO:
resolves https://github.com/cloudflare/miniflare/issues/402

### DIFF
--- a/packages/r2/src/r2Object.ts
+++ b/packages/r2/src/r2Object.ts
@@ -164,7 +164,7 @@ function matchStrings(a: string | string[], b: string): boolean {
 }
 
 // headers can be a list: e.g. ["if-match", "a, b, c"] -> "if-match: [a, b, c]"
-function parseHeaderArray(input: string): undefined | string | string[] {
+function parseHeaderArray(input: string): string | string[] {
   // split if comma found, otherwise return input
   if (!input.includes(",")) return _stripQuotes(input);
   return input.split(",").map((x) => _stripQuotes(x.trim()));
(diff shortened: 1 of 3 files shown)


packages/r2/src/r2Object.ts
```diff
@@ -164,10 +164,14 @@ function matchStrings(a: string | string[], b: string): boolean {
...
-  if (typeof input !== "string") return;
-  if (!input.includes(",")) return input;
-  return input.split(",").map((x) => x.trim());
+function parseHeaderArray(input: string): string | string[] {
+  // split if comma found, otherwise return input
+  if (!input.includes(",")) return _stripQuotes(input);
+  return input.split(",").map((x) => _stripQuotes(x.trim()));
+}
+
+function _stripQuotes(input: string): string {
```
mrbbot, line 173: Nit: as this function is not `@internal`, it doesn't need to start with an `_`:
```suggestion
function stripQuotes(input: string): string {
```

packages/r2/src/r2Object.ts
```diff
@@ -164,10 +164,14 @@ function matchStrings(a: string | string[], b: string): boolean {
...
+  return input.replaceAll('"', "");
```
mrbbot, line 174: Rather than removing all `"`s, could you just remove the first and/or last characters if needed? Given this would require `trim()`ing first, maybe move the call to `trim()` into this function too?

packages/r2/test/r2Object.spec.ts
```diff
@@ -427,6 +427,20 @@ test("R2Object: parseOnlyIf: each parameter is parsed correctly as an R2Conditio
   t.deepEqual(parsed.uploadedAfter, new Date(0));
 });
 
+test("R2Object: parseOnlyIf with quotes: each parameter is parsed correctly as an R2Conditional object", (t) => {
+  const r2conditional = {
+    etagMatches: '"*"',
+    etagDoesNotMatch: ['"123"', '"456"'],
+    uploadedBefore: new Date(0),
```
mrbbot, line 434: Shouldn't this also be a `Date` string to test the removal of quotes behaviour?

mrbbot (commented):
Hey! 👋 Thanks for fixing this! 🙂 Added a few comments.

PR #405, opened 2022-10-09
Correct Content-Length for range responses from the cache

notorca:
Responses from the cache for range requests have the original "Content-Length". Additionally this breaks `response.arrayBuffer()` it returns buffer with invalid size.


mrbbot (approved):
Awesome! Thanks for fixing this! 🙂

PR #411, opened 2022-10-12
Add support for dead letter queues

jbw1991:
* Queue broker now respects the configured number of max retries for a consumer
* Queue broker now respects the configured "dead letter queue". After reaching the max number of attempts, "failed" messages are placed on this queue.

### DIFF
--- a/packages/queues/src/broker.ts
+++ b/packages/queues/src/broker.ts
@@ -199,7 +199,7 @@ export class Queue<Body = unknown> implements QueueInterface<Body> {
       return;
     }
     const maxAttempts = this.#consumer.maxRetries + 1;
-    const deadLetterQueue = this.#consumer.deadLetterQueue;
+    const deadLetterQueueName = this.#consumer.deadLetterQueue;
 
     // Create a batch and execute the queue event handler
     const batch = new MessageBatch<Body>(this.#queueName, [...this.#messages]);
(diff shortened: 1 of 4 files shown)


packages/queues/src/broker.ts
```diff
@@ -216,12 +214,41 @@ export class Queue<Body = unknown> implements QueueInterface<Body> {
...
+      } else {
+        this.#log?.warn(
+          `Dropped message "${msg.id}" after ${maxAttempts} failed attempts!`
+        );
+      }
+    });
+
+    if (toRetry.length) {
+      this.#messages.push(...toRetry);
+      if (this.#messages.length > 0) {
```
mrbbot, line 243: I think this check is redundant, if `toRetry.length` is truthy, it must be `> 0`, so messages will have been added to `#messages` and therefore it's length must be `> 0`.

packages/queues/src/broker.ts
```diff
@@ -216,12 +214,41 @@ export class Queue<Body = unknown> implements QueueInterface<Body> {
...
+        this.#ensurePendingFlush();
+      }
+    }
+
+    if (deadLetterQueue) {
+      toDLQ.forEach((msg) => {
+        this.#broker.getOrCreateQueue(deadLetterQueue).send(msg.body);
```
mrbbot, line 250: Nit: could you lift `this.#broker.getOrCreateQueue(deadLetterQueue)` out of the `forEach`?

mrbbot (approved):
Awesome! ✅ Added a couple tiny suggestions but this looks great! 🙂

packages/queues/src/broker.ts
jbw1991: 👍 Done

a-robinson:
Hi @mrbbot , any chance we could release this change? Is there much work involved in doing so?

a-robinson:
Thanks! 😃

PR #413, opened 2022-10-15
Add `getMiniflareDurableObjectIds` global test function

### LINKED ISSUE #384: `getMiniflareDurableObjectIds(class: string)` helper function
Durable Object IDs have no runtime binding to be listed, however the Cloudflare API has an endpoint to retrieve all created Durable Objects, with the additional information whether there is storage attached.

A helper function like `getMiniflareDurableObjectIds(class: string)` could return an array like the API, probably without the pagination, or the limits.

This could be useful in tests to verify if in certain cases no DO is created, or that no storage is used.
A possible extension could be to also provide a `_stored` key which shows how many bytes the Durable Object is using.

DaniFoldi:
As discussed, it also includes a fix to `vitest-environment-miniflare` versioning.

Resolves #384.

### DIFF
--- a/packages/durable-objects/src/plugin.ts
+++ b/packages/durable-objects/src/plugin.ts
@@ -267,6 +267,15 @@ export class DurableObjectsPlugin
     }).runWith(() => state[kAlarm]());
   }
 
+  getObjects(
+    storageFactory: StorageFactory,
+    namespace: string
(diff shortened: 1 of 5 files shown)


mrbbot (approved):
Awesome! ✅ Thanks for the PR! 😃

PR #416, opened 2022-10-31
Add `KVGateway` tests, fix `cacheTtl` validation and last `cursor`

mrbbot:
- Upgrade to `ava@5`
- Port `KVNamespace` tests from Miniflare 2: https://github.com/cloudflare/miniflare/blob/master/packages/kv/test/namespace.spec.ts
- Only throw if `cacheTtl` is invalid, not just specified
- Return `cursor: undefined` on the last page to match the behaviour of the Worker's runtime

### DIFF
--- a/ava.config.mjs
+++ b/ava.config.mjs
@@ -8,9 +8,6 @@ const rewritePaths = Object.fromEntries(
 );
 
 export default {
-  nonSemVerExperiments: {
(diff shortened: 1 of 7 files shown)


JacobMGEvans (approved):
Hoping to get this in so I can use the updated AVA, I can also just cut from this branch and treat it like the main testing branch.

PR #419, opened 2022-11-02
Support Sqlite as a storage backend

penalosa:
I thought this would be relatively simple, so here's a quick first attempt. It's only lightly tested, but supporting sqlite should actually be pretty simple. It might be worth thinking about using this as a default in the future, as it resolves a whole class of problems around keys.


packages/tre/src/storage/sqlite.ts
```diff
@@ -0,0 +1,183 @@
...
+    this.database.exec(SCHEMA);
+    this.sql = (
+      parts: TemplateStringsArray,
+      ...args: (string | number | Buffer)[]
+    ) => {
+      const query = parts.join(" ? ");
+      const stmt = this.database.prepare(query);
+      if (
+        query.toLowerCase().startsWith("insert") ||
+        query.toLowerCase().startsWith("delete")
```
mrbbot, line 80: Could you reverse this to `query.toLowerCase().startsWith("select")`?

packages/tre/src/storage/sqlite.ts
```diff
@@ -0,0 +1,183 @@
...
+  }
+
+  async hasMaybeExpired(key: string): Promise<StoredMeta | undefined> {
+    const data = this.sql<{
+      meta: string;
+    }>`SELECT meta from storage where key=${key}`;
+    if (data === undefined || data.length === 0) return;
+    return {
+      metadata: JSON.parse(data[0].meta).metadata,
+      expiration: JSON.parse(data[0].meta).expiration,
```
mrbbot, line 96: Could you lift the parsed `meta` out into a variable so you only parse it once?

packages/tre/src/storage/sqlite.ts
```diff
@@ -0,0 +1,183 @@
...
+  async headMaybeExpired<Meta>(
+    key: string
+  ): Promise<FileMeta<Meta> | undefined> {
+    const data = this.sql<{
+      meta: string;
+    }>`SELECT meta from storage where key=${key}`;
+    if (data === undefined || data.length === 0) return;
+    return {
+      metadata: JSON.parse(data[0].meta).metadata,
+      expiration: JSON.parse(data[0].meta).expiration,
```
mrbbot, line 109: Could you lift the parsed `meta` out into a variable so you only parse it once?

packages/tre/src/storage/sqlite.ts
```diff
@@ -0,0 +1,183 @@
...
+    const data = this.sql<{
+      key: string;
+      data: Buffer;
+      meta: string;
+    }>`SELECT * from storage where key=${key}`;
+    if (data === undefined || data.length === 0) return;
+    return {
+      value: new Uint8Array(data[0].data),
+      metadata: JSON.parse(data[0].meta).metadata,
+      expiration: JSON.parse(data[0].meta).expiration,
```
mrbbot, line 125: Could you lift the parsed `meta` out into a variable so you only parse it once?

packages/tre/src/storage/sqlite.ts
```diff
@@ -0,0 +1,183 @@
...
+    const data = this.sql<{
+      key: string;
+      data: Buffer;
+      meta: string;
+    }>`SELECT * from storage where key=${key}`;
+    if (data === undefined || data.length === 0) return;
+    const entry: StoredValueMeta<Meta> = {
+      value: new Uint8Array(data[0].data),
+      metadata: JSON.parse(data[0].meta).metadata,
+      expiration: JSON.parse(data[0].meta).expiration,
```
mrbbot, line 142: Could you lift the parsed `meta` out into a variable so you only parse it once?

packages/tre/src/storage/sqlite.ts
```diff
@@ -0,0 +1,183 @@
...
+    };
+    const res = await getRange(entry.value, _offset, _length, suffix);
+    if (res === undefined) return;
+
+    const { value, offset, length } = res;
+    return {
+      ...entry,
+      range: { offset, length },
+      value: value,
```
mrbbot, line 151: Nit:
```suggestion
      value,
```

packages/tre/src/storage/sqlite.ts
```diff
@@ -0,0 +1,183 @@
...
+  ): Promise<void> {
+    this
+      .sql`INSERT OR REPLACE INTO storage (key, data, meta) VALUES (${key}, ${Buffer.from(
+      value
+    )}, ${JSON.stringify({ key, expiration, metadata })})`;
+  }
+
+  async deleteMaybeExpired(key: string): Promise<boolean> {
+    this.sql`DELETE FROM storage WHERE key=${key}`;
+    return true;
```
mrbbot, line 167: This should only return `true` when the key was deleted. Are you able to detect this somehow?

packages/tre/src/storage/sqlite.ts
```diff
@@ -0,0 +1,183 @@
...
+    for (const { key, meta } of this.sql<{
+      key: string;
+      meta: string;
+    }>`SELECT key, meta from storage`!) {
+      keys.push({
+        name: key,
+        ...JSON.parse(meta),
+      });
+    }
+    return keys;
```
mrbbot, line 181: Could this be a `.map()` instead?

packages/tre/src/storage/sqlite.ts
```diff
@@ -0,0 +1,183 @@
...
+  private sql: <Row>(
+    parts: TemplateStringsArray,
+    ...args: (string | number | Buffer)[]
+  ) => Row[] | undefined;
+
+  constructor(database: URL, clock = defaultClock) {
+    super(clock);
+    this.database = new Database(path.resolve(database.pathname));
+    this.database.exec(SCHEMA);
+    this.sql = (
```
mrbbot, line 72: I was about to write a comment about SQL injection when I saw the below code 😅 , but this is really cool! Is there any reason it's defined in the constructor though, and not as a regular class method?

packages/tre/src/storage/sqlite.ts
```diff
@@ -0,0 +1,183 @@
...
+}
+
+export interface FileMeta<Meta = unknown> extends StoredMeta<Meta> {
+  key?: string;
+}
+const SCHEMA = `
+CREATE TABLE IF NOT EXISTS storage (
+    key text NOT NULL PRIMARY KEY,
+    data blob,
+    meta text
```
mrbbot, line 57: Could we rename `data` to `value` to match Miniflare's `StoredValue` interface? Also, this is super-nit-picky so I apologise 😅, could we use a consistent capitalisation on SQL keywords, so `TEXT`, `BLOB`, etc. Also `FROM` and `WHERE` in the queries below.

packages/tre/src/storage/sqlite.ts
```diff
@@ -0,0 +1,183 @@
...
+  }
+
+  async put<Meta = unknown>(
+    key: string,
+    { value, expiration, metadata }: StoredValueMeta<Meta>
+  ): Promise<void> {
+    this
+      .sql`INSERT OR REPLACE INTO storage (key, data, meta) VALUES (${key}, ${Buffer.from(
+      value
+    )}, ${JSON.stringify({ key, expiration, metadata })})`;
```
mrbbot, line 162: Could we extract `Buffer.from(...)` and `JSON.stringify({ ... })}` into variables? Also, does SQLite not accept `Uint8Array`s for binary data?

packages/tre/src/storage/sqlite.ts
```diff
@@ -0,0 +1,183 @@
...
+  StoredValueMeta,
+} from "./storage";
+
+export interface FileRange {
+  value: Uint8Array;
+  offset: number;
+  length: number;
+}
+
+export async function getRange(
```
mrbbot, line 19: This function doesn't need to be `async`.

packages/tre/src/storage/sqlite.ts
```diff
@@ -0,0 +1,183 @@
...
+        query.toLowerCase().startsWith("delete")
+      ) {
+        stmt.run(...args);
+      } else {
+        return stmt.all(...args);
+      }
+    };
+  }
+
+  async hasMaybeExpired(key: string): Promise<StoredMeta | undefined> {
```
mrbbot, line 89: This function, and all the others in this class, don't need to be `async`. (assuming `getRange` is updated)

package.json
```diff
@@ -57,5 +58,8 @@
   },
   "volta": {
     "node": "18.2.0"
+  },
+  "dependencies": {
+    "better-sqlite3": "^7.6.2"
```
mrbbot, line 63: This and `@types/better-sqlite3` should be dependencies of `@miniflare/tre` instead.

mrbbot (commented):
Nice! We'll need to think about how this integrates with D1, but it looks like that code isn't included in the `tre` branch atm, so don't worry about that yet.

packages/tre/src/plugins/shared/gateway.ts
```diff
@@ -57,6 +62,8 @@ export class GatewayFactory<Gateway> {
         const unsanitise =
           url.searchParams.get(PARAM_FILE_UNSANITISE) === "true";
         return new FileStorage(root, !unsanitise);
+      } else if (url.protocol === "sqlite:") {
+        return new SqliteStorage(url);
```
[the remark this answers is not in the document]
penalosa: Multiple `new Database`s should be fine. I'd be hesitant about using `sanitisedNamespace` as the table name, since you parameterise a table name in SQL query, so that would have to be added directly into the string (definitely opening up SQL injection). SQL injection may not be a problem _right now_ since it's all local data and developers can access the database directly anyway, but it's still a bit of a concern. Storing everything in one table with a `namespace` field would seem to be okay?

packages/tre/src/storage/sqlite.ts
penalosa: I've made extra helper methods instead

packages/tre/src/storage/sqlite.ts
penalosa: Nope, I've moved it

packages/tre/src/storage/sqlite.ts
penalosa: Seems like it does

packages/tre/src/storage/sqlite.ts
penalosa: They don't, but I don't want consumers of this class to rely on that (the specific SQLite driver I'm using happens to be sync, but most are async)

mrbbot: That's a good point 👍 I don't expect we'll ever be using this class directly, rather it should always just be an implementation of `Storage` which requires you to `await` all results anyways. Let's keep them `async`. 🙂

packages/tre/src/plugins/shared/gateway.ts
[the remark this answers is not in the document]
mrbbot: SQL injection is definitely a concern. If we did use `namespace` as the table name, we'd want to make sure we were escaping it for sure. The reason I think it might be nicer, is SQL GUIs, including those integrated into IDEs, will show tables like directories, making it much easier to find/edit data in tables (like in the dashboard) without writing SQL queries/filters.



Using different table names also means you should be able to reuse the same SQLite database file for KV, DO and R2 (assuming the namespace IDs for each are different). This again means you wouldn't have to have multiple databases open in your IDE, improving the developer experience.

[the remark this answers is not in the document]
penalosa: I agree that SQL injection is a concern, which is why I put the namespace as a field. The IDE experience being poor is valid though (although regardless of the solution here, there is only going to be one database file), and should be solved.

My main concern is that we're not able to safely escape the table name, since SQL (whether sqlite or any other engine) doesn't support that. I don't want us to start thinking about doing SQL sanitisation on the namespace ourselves, because that's too easy to accidentally get wrong. Another approach is to run a very strict allowlist regex over the namespace, and use that as the table name, along with a unique suffix (either a GUID or an incrementing number, which could be omitted if the allowlisted name was already unique)? (I'm thinking `replace(/[^a-zA-Z0-9_]+/g, '-'). This namespace-tablename mapping could be stored in a separate table, and wouldn't introduce much overhead, since it would just need to be read/written at database instantiation. The table names wouldn't _exactly_ correspond to namespaces, but they'd be close enough to match by eye in an IDE (`TEST_BUCKET`->`TEST_BUCKET`, `TEST BUCKET`->`TEST-BUCKET`)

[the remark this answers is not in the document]
mrbbot: I like the allowlist idea, maybe we could concatenate a small hash of the full-namespace afterwards instead to avoid adding an additional table?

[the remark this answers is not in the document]
penalosa: A hash makes much more sense than a lookup table :joy:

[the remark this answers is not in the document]
penalosa: Right, that's implemented with a 2 byte hash, which should be good

packages/tre/package.json
```diff
@@ -45,12 +45,14 @@
     "stoppable": "^1.1.0",
     "undici": "^5.10.0",
     "workerd": "^1.20220926.3",
-    "zod": "^3.18.0"
+    "zod": "^3.18.0",
+    "better-sqlite3": "^7.6.2"
```
mrbbot, line 49: This should be ordered alphabetically :)

mrbbot (commented):
Nice! A few more minor comments, but looking good! 😃

packages/tre/src/storage/sqlite.ts
```diff
@@ -0,0 +1,228 @@
...
+  StoredKeyMeta,
+  StoredMeta,
+  StoredValueMeta,
+} from "./storage";
+
+export interface FileRange {
+  value: Uint8Array;
+  offset: number;
+  length: number;
+}
```
mrbbot, line 19: I think this interface is unused too.

packages/tre/src/storage/sqlite.ts
```diff
@@ -0,0 +1,228 @@
...
+    const stmt = this.database.prepare(query.template);
+    return this.parse(stmt.get(...query.parameters));
+  }
+
+  constructor(database: string, namespace: string, clock = defaultClock) {
+    super(clock);
+    this.database = new Database(database);
+
+    const hash = crypto
+      .createHash("shake256", { outputLength: 2 })
```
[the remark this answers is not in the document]
penalosa: I'm not actually sure... The docs don't seem to have a list, but if it works in v16 that's probably fine?

packages/tre/package.json
mrbbot: ...within the rest of the list :)

mrbbot (approved):
Nice! 😃 One tiny thing with the `package.json` dependency order, and looks like there's a conflict there now too, but other than that... ✅

Issue #421, opened 2022-11-03
Fetching rocket-loader.min.js always 404s


jstevans:
It looks like [there's logic](https://github.com/cloudflare/miniflare/blob/master/packages/http-server/src/index.ts#L275-L296) specifically to 404 whenever a request is made against `/cdn-cgi/*`, with the exception of Scheduled Tasks. 

For folks with a Cloudflare Worker set-up in front of their Cloudflare Pages apps, and who have Rocket Loader enabled, this breaks their local development flow. `/cdn-cgi/scripts/[...]/rocket-loader.min.js` appears as a script element src in HTML, and 404s cause the page to fail to load.

mrbbot:
Closed by https://github.com/cloudflare/miniflare/pull/422.

PR #422, opened 2022-11-03
Forward cdn-cgi/scripts paths to Cloudflare

jstevans:
This addresses issue #421.

Right now, Miniflare 404s on attempts to download `rocket-loader.min.js`. For folks who've set up a Worker in front of Cloudflare Pages (e.g. to handle API requests on the same domain), this breaks their web app during local development: it renders as a white screen.

This PR adds a small amount of logic to simply forward those requests onto Cloudflare.

### DIFF
--- a/packages/http-server/src/index.ts
+++ b/packages/http-server/src/index.ts
@@ -5,6 +5,7 @@ import http from "http";
 import https from "https";
 import net from "net";
 import { Transform, Writable } from "stream";
+import consumers from "stream/consumers";
 import { ReadableStream } from "stream/web";
 import { URL } from "url";
 import zlib from "zlib";
@@ -288,6 +289,15 @@ export function createRequestListener<Plugins extends HTTPPluginSignatures>(
           url
         );
         status = 200;
+      } else if (pathname.startsWith("/cdn-cgi/scripts/") {
+        response = await fetch(new URL(pathname, "https://cloudflare.com"));
+        status = response.status;
+        if (res) {
+          const bytes = consumers.buffer(response.body);
+          const body = String.fromCharCode.apply(null, bytes);
+          response = new Response(body);
+          await writeResponse(response, res, HTTPPlugin2.liveReload, mf.log);
+        }
       } else {
         status = 404;
       }


packages/http-server/src/index.ts
```diff
@@ -288,6 +289,15 @@ export function createRequestListener<Plugins extends HTTPPluginSignatures>(
...
         );
         status = 200;
+      } else if (pathname.startsWith("/cdn-cgi/scripts/") {
+        response = await fetch(new URL(pathname, "https://cloudflare.com"));
+        status = response.status;
+        if (res) {
+          const bytes = consumers.buffer(response.body);
+          const body = String.fromCharCode.apply(null, bytes);
+          response = new Response(body);
+          await writeResponse(response, res, HTTPPlugin2.liveReload, mf.log);
```
mrbbot, line 299: You shouldn't need to buffer and recreate the response here. I'd just write it directly.
```suggestion
          await writeResponse(response, res, HTTPPlugin.liveReload, mf.log);
```

packages/http-server/src/index.ts
```diff
@@ -288,6 +289,15 @@ export function createRequestListener<Plugins extends HTTPPluginSignatures>(
           url
         );
         status = 200;
```
mrbbot, line 291: Because we don't want to write a response twice, you'll need to move the `res?.writeHead(status, { "Content-Type": "text/plain; charset=UTF-8" });` and `res?.end();` lines below underneath this line, and also underneath the `status = 404;` line below.

mrbbot (commented):
Hey! 👋 Thanks for the PR! Added a couple comments. 👍

jstevans:
Thanks for the feedback! I've done as suggested on both :)

packages/http-server/src/index.ts
```diff
@@ -288,11 +289,19 @@ export function createRequestListener<Plugins extends HTTPPluginSignatures>(
...
+        res?.writeHead(status, { "Content-Type": "text/plain; charset=UTF-8" });
+        res?.end();
+      } else if (pathname.startsWith("/cdn-cgi/scripts/") {
```
mrbbot, line 294:
```suggestion
      } else if (pathname.startsWith("/cdn-cgi/scripts/")) {
```

packages/http-server/src/index.ts
```diff
@@ -288,11 +289,19 @@ export function createRequestListener<Plugins extends HTTPPluginSignatures>(
...
+        response = await fetch(new URL(pathname, "https://cloudflare.com"));
```
mrbbot, line 295: You'll need to import `fetch` from `@miniflare/core` at the top of this file

mrbbot (commented):
Nice! Thanks for making those fixes, I've added a couple more, then we can merge this. 👍

jstevans:
@mrbbot Done!

mrbbot (approved):

PR #423, opened 2022-11-06
Fix incorrect request body when intercepting fetch with getMiniflareFetchMock

robertcepa:
Hey there!

I am using `miniflare` with `jest` to test a Workers library. I need to write a simple unit test that intercepts a request body and returns it in response.

Example ([full code here](https://github.com/robertcepa/toucan-js/blob/undici-issue-1756/test/index.spec.ts)):
```
fetchMock.get('...')
  .intercept({
    method: () => true,
    path: () => true,
  })
  .reply(200, (opts) => {
    return opts;
  });
```
The issue I'm encountering is that `opts.body` doesn't contain request body string, as expected per https://undici.nodejs.org/#/docs/best-practices/mocking-request?id=reply-with-data-based-on-request , but `AsyncGenerator`.

I first thought that this was an issue in `undici` but after discussion with their team (https://github.com/nodejs/undici/issues/1756) it seems that the solution is to expose `isMockActive` property on `Dispatcher` for test utilities on `MockAgent` to work properly.

This PR adds a new `MiniflareDispatcherMock` class that extends the base `MiniflareDispatcher` with `isMockActive` getter. I decided to not add the getter directly to `MiniflareDispatcher`, because it's a testing function that doesn't exist on real dispatchers.

This class is used on `fetch` polyfill to instantiate dispatcher when bound with `MockAgent`.

### DIFF
--- a/packages/core/src/standards/http.ts
+++ b/packages/core/src/standards/http.ts
@@ -736,6 +736,15 @@ class MiniflareDispatcher extends Dispatcher {
     // @ts-expect-error just want to pass through to global dispatcher here
     return this.inner.destroy(...args);
   }
+
+  /**
+   * Required for some testing utilities.
+   * For context see: https://github.com/nodejs/undici/issues/1756#issuecomment-1304711596
+   */
+  get isMockActive(): boolean {
+    // @ts-expect-error Missing type on MockAgent, but exists at runtime
+    return this.inner.isMockActive ?? false;
+  }
 }
 
 export async function fetch(
--- a/packages/jest-environment-miniflare/test/fixtures/core.module.spec.js
+++ b/packages/jest-environment-miniflare/test/fixtures/core.module.spec.js
@@ -13,3 +13,31 @@ test("handles requests with mocked upstream", async () => {
   const res = await fetch(new Request("https://random.mf"));
   expect(await res.text()).toBe("Hello World!");
 });
+
+test("reply callback options has expected data", async () => {
+  const mockAgent = getMiniflareFetchMock();
+  mockAgent.disableNetConnect();
+  const client = mockAgent.get("https://random.mf");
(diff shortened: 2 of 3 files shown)


packages/core/src/standards/http.ts
```diff
@@ -738,6 +738,26 @@ class MiniflareDispatcher extends Dispatcher {
...
+    // EventEmitterOptions isn't exported by the "events" module
+    options?: ConstructorParameters<typeof EventEmitter>[0]
+  ) {
+    super(inner, removeHeaders, options);
+  }
+
+  get isMockActive(): boolean {
+    // @ts-expect-error Missing type on MockAgent, but exists at runtime
+    return this.inner.isMockActive ?? false;
+  }
```
mrbbot, line 758: Would it be possible to just add this getter to `MiniflareDispatcher` without creating a new `MiniflareDispatcherMock` class?

mrbbot (commented):
Hey! 👋 Thanks for the PR! Added a small comment. 👍

mrbbot (approved):

sjblurton:
Well done @robertcepa this is something we have been trying to work around here. Thank you for the PR.

Hey @mrbbot, we have been trying to get this working here since mid-last year and have just noticed someone has already fixed it. Do you have any idea when this will be released? Thank you.

ashleyoldershaw:
Hi! I'm also invested in seeing this released :) do we have a timeline?

sjblurton:
@mrbbot was talking about this PR today with the team. Is there a plan for a wide release? The automated testing we are trying to implement needs this.

sjblurton:
Can we get any updates on this one, please? We really want to be able to test our API. 😁

CameronB15:
Just stumbled across this PR when trying to figure out how to unit test my worker. Do you know when this will be added to a release @mrbbot?

PR #424, opened 2022-11-07
Actually respect `port` and `host` option

mrbbot:
Previously, `port` was being passed through `get-port` before being used. This meant that if `Miniflare` was disposed and then immediately created again (as opposed to using `setOptions`), it couldn't use the same `port`, as `get-port` deliberately doesn't allocate recently allocated ports.

Notably, Wrangler does exactly this: re-create `Miniflare` instances. Ideally, we should be using `setOptions` there, but this at least means the port stays the same between reloads.

This PR also passes the `host` option to `workerd`, now that we're not putting an HTTP proxy in front of `workerd` in Miniflare.

### DIFF
--- a/packages/tre/src/index.ts
+++ b/packages/tre/src/index.ts
@@ -216,13 +216,11 @@ export class Miniflare {
     const loopbackPort = address.port;
 
     // Start runtime
-    const entryPort = await getPort({ port: this.#sharedOpts.core.port });
-
-    // TODO: respect entry `host` option
-    this.#runtime = new this.#runtimeConstructor(entryPort, loopbackPort);
+    const host = this.#sharedOpts.core.host ?? "127.0.0.1";
+    const port = this.#sharedOpts.core.port ?? (await getPort({ port: 8787 }));
+    this.#runtime = new this.#runtimeConstructor(host, port, loopbackPort);
     this.#removeRuntimeExitHook = exitHook(() => void this.#runtime?.dispose());
-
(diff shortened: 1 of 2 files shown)


packages/tre/src/index.ts
```diff
@@ -216,13 +216,11 @@ export class Miniflare {
     const loopbackPort = address.port;
 
     // Start runtime
-    const entryPort = await getPort({ port: this.#sharedOpts.core.port });
-
-    // TODO: respect entry `host` option
-    this.#runtime = new this.#runtimeConstructor(entryPort, loopbackPort);
+    const host = this.#sharedOpts.core.host ?? "127.0.0.1";
+    const port = this.#sharedOpts.core.port ?? (await getPort({ port: 8787 }));
```
penalosa, line 220: This will now fail if `core.port` is in use, rather than falling back to 8787

penalosa (changes requested):
This takes the port selection out of getPort, which means if the core specified port is in use, Miniflare will now fail rather than recovering with a different port. Is there some way to get around that? (perhaps catching the requisite error from workerd?)

packages/tre/src/index.ts
mrbbot: I actually think this is nicer behaviour. If the user is explicitly setting the `port` option, we should fail if we can't bind to that port. Previously, this option was acting more like `preferredPort` which I don't think is right. If users want this behaviour, they could call `getPort` themselves before calling `new Miniflare()`.

penalosa (approved):

PR #429, opened 2022-11-09
Add support for live reload

mrbbot:
This PR brings over Miniflare 2's live reload feature to Miniflare 3. This automatically reloads pages whenever options are changed or the server is restarted.

Note that instead of terminating WebSocket's in `workerd`, we terminate them in `Miniflare`. This allows us to instantly send messages to all connected clients once options are applied, reducing the reload latency. :zap:

### DIFF
--- a/packages/tre/src/index.ts
+++ b/packages/tre/src/index.ts
@@ -1,5 +1,7 @@
 import assert from "assert";
 import http from "http";
+import net from "net";
+import { Duplex } from "stream";
 import exitHook from "exit-hook";
 import getPort from "get-port";
 import { bold, green, grey } from "kleur/colors";
@@ -12,9 +14,9 @@ import {
   Response,
   fetch,
 } from "undici";
+import { WebSocketServer } from "ws";
 import { z } from "zod";
 import { setupCf } from "./cf";
-
 import {
   GatewayConstructor,
   GatewayFactory,
@@ -178,6 +180,8 @@ export class Miniflare {
   // Aborted when dispose() is called
   readonly #disposeController: AbortController;
   #loopbackServer?: StoppableServer;
+  #loopbackPort?: number;
+  readonly #liveReloadServer: WebSocketServer;
 
   constructor(opts: MiniflareOptions) {
     // Initialise plugin gateway factories and routers
@@ -201,6 +205,7 @@ export class Miniflare {
     );
 
     this.#disposeController = new AbortController();
+    this.#liveReloadServer = new WebSocketServer({ noServer: true });
     this.#runtimeMutex = new Mutex();
     this.#initPromise = this.#runtimeMutex.runWith(() => this.#init());
   }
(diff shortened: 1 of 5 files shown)


JacobMGEvans (approved):

packages/tre/src/index.ts
```diff
@@ -201,6 +205,7 @@ export class Miniflare {
     );
 
     this.#disposeController = new AbortController();
+    this.#liveReloadServer = new WebSocketServer({ noServer: true });
```
penalosa, line 208: Why is this `noServer`? Doesn't it add a lot of complexity?

mrbbot: We'd like to reuse the loopback server, rather than creating a separate server that binds to a different port.

packages/tre/src/plugins/core/index.ts
```diff
@@ -86,27 +90,84 @@ const BINDING_JSON_VERSION = "MINIFLARE_VERSION";
...
+  function reload() { location.reload(); }
+  function connect(reconnected) {
+    var ws = new WebSocket(url);
+    if (reconnected) ws.onopen = reload;
+    ws.onclose = function(e) {
+      e.code === 1012 ? reload() : e.code === 1000 || e.code === 1001 || setTimeout(connect, 1000, true);
+    }
+  }
+  connect();
+})();
```
penalosa, line 111: I'm slightly uneasy about this code—it's using quite old style JS, and seems a bit fragile. It would be nice to see an exponential backoff reconnection rather than a 1000ms one, as well? If possible, could it be made a function that interpolated the `LIVE_RELOAD_SCRIPT_PORT_PLACEHOLDER` variable as a function argument, rather than a constant?

penalosa: Or a variable binding

packages/tre/src/plugins/core/index.ts
```diff
@@ -86,27 +90,84 @@ const BINDING_JSON_VERSION = "MINIFLARE_VERSION";
...
+    const liveReloadScript = globalThis.${BINDING_DATA_LIVE_RELOAD_SCRIPT};
+    if (
+      liveReloadScript !== undefined &&
+      response.headers.get("content-type")?.toLowerCase().includes("text/html")
+    ) {
+      const headers = new Headers(response.headers);
+      const contentLength = parseInt(headers.get("content-length"));
+      if (!isNaN(contentLength)) {
+        headers.set("content-length", contentLength + liveReloadScript.byteLength);
+      }
```
penalosa, line 144: Very nice!

packages/tre/src/plugins/core/index.ts
```diff
@@ -86,27 +90,84 @@ const BINDING_JSON_VERSION = "MINIFLARE_VERSION";
...
+      
+      const { readable, writable } = new IdentityTransformStream();
+      event.waitUntil((async () => {
+        await response.body?.pipeTo(writable, { preventClose: true });
+        const writer = writable.getWriter();
+        await writer.write(liveReloadScript);
```
penalosa, line 150: Will this... work? A `<script/>` element after the closing `</html>`?

penalosa (changes requested):
This looks great overall! Couple small things

packages/tre/src/plugins/core/index.ts
mrbbot: Yep, a function is a much nicer solution here. 👍

packages/tre/src/plugins/core/index.ts
mrbbot: Yep, https://github.com/cloudflare/miniflare/pull/79 🙂

penalosa (approved):
Lgtm 🎉

PR #430, opened 2022-11-10
Support remote KV storage

mrbbot:
This PR adds experimental support for reading/writing from real KV namespaces. The user's computer is treated like another "colo", and values are cached there as they are in KV, respecting `cacheTtl`s passed to `get`.

> :warning: We're still exploring how this feature will work, so it's likely it'll change in the future.

### DIFF
--- a/packages/tre/src/plugins/kv/remote.ts
+++ b/packages/tre/src/plugins/kv/remote.ts
@@ -1,0 +1,272 @@
+import assert from "assert";
+import { Blob } from "buffer";
+import { BodyInit, FormData, Response } from "undici";
+import { z } from "zod";
+import { Clock, millisToSeconds } from "../../shared";
+import {
+  RemoteStorage,
+  StorageListOptions,
+  StorageListResult,
+  StoredValueMeta,
+} from "../../storage";
+import { KVError } from "./gateway";
+
+interface RemoteCacheMetadata {
+  // UNIX timestamp in seconds when this key was last modified. KV allows you
+  // to reduce the cache TTL if it was previously set too high
+  // (https://developers.cloudflare.com/workers/runtime-apis/kv/#cache-ttl).
+  // This is used to check if the data should be revalidated from the remote.
+  storedAt: number;
+  // Whether this key represents a deleted value and should be treated as
+  // `undefined`.
+  tombstone?: true;
+  // The actual user-specified expiration of this key, if any. We want to return
+  // this to users instead of our cache expiration.
+  actualExpiration?: number;
+  // The actual user-specified metadata of this key, if any. We want to return
+  // this to users instead of this object.
+  actualMetadata?: unknown;
+}
+
+const APIEnvelopeSchema = z.object({
+  success: z.boolean(),
+  errors: z.array(z.object({ code: z.number(), message: z.string() })),
+  messages: z.array(z.object({ code: z.number(), message: z.string() })),
+});
+
+const KVGetMetadataResponseSchema = z.intersection(
+  APIEnvelopeSchema,
+  z.object({ result: z.unknown() })
+);
+
+const KVListResponseSchema = z.intersection(
+  APIEnvelopeSchema,
+  z.object({
+    result: z.array(
+      z.object({
+        name: z.string(),
+        expiration: z.onumber(),
+        metadata: z.unknown(),
+      })
+    ),
+    result_info: z.optional(
+      z.object({
+        count: z.onumber(),
+        cursor: z.ostring(),
+      })
+    ),
+  })
+);
+
+async function assertSuccessfulResponse(response: Response) {
+  if (response.ok) return;
+
+  // If this wasn't a successful response, throw a KVError
+  const contentType = response.headers.get("Content-Type");
+  if (contentType?.toLowerCase().includes("application/json")) {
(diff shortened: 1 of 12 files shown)


packages/tre/src/plugins/kv/remote.ts
```diff
@@ -0,0 +1,225 @@
...
+  RemoteStorage,
+  StorageListOptions,
+  StorageListResult,
+  StoredValueMeta,
+} from "../../storage";
+import { KVError } from "./gateway";
+
+type TombstoneMetadata =
+  | undefined
+  | { $MINIFLARE_INTERNAL_REMOTE_TOMBSTONE?: true };
```
penalosa, line 15: Can this be a `Symbol` instead?

mrbbot: Unfortunately not, since it needs to be JSON-serialisable in case the file system is being used as the cache. 😞

penalosa: Ah, fair

packages/tre/src/plugins/kv/remote.ts
```diff
@@ -0,0 +1,273 @@
...
+    skipMetadata?: boolean,
+    cacheTtl = DEFAULT_CACHE_TTL
+  ): Promise<StoredValueMeta | undefined> {
+    // If this key is cached, return it
+    const cachedValue = await this.cache.get<RemoteCacheMetadata>(key);
+    // cacheTtl may have changed between the original get call that cached
+    // this value and now, so check the cache is still fresh with the new TTL
+    if (
+      cachedValue?.metadata !== undefined &&
+      !cacheNeedsRevalidation(this.clock, cacheTtl, cachedValue)
```
penalosa, line 117: I'm not certain, but from reading the docs I think we'd also need to _store_ this new `cacheTtl`

packages/tre/src/plugins/kv/remote.ts
```diff
@@ -0,0 +1,273 @@
...
+    }
+
+    // Otherwise, fetch the key...
+    const encodedKey = encodeURIComponent(key);
+    const valueResource = `storage/kv/namespaces/${this.namespace}/values/${encodedKey}`;
+    const metadataResource = `storage/kv/namespaces/${this.namespace}/metadata/${encodedKey}`;
+    const [valueResponse, metadataResponse] = await Promise.all([
+      this.cloudflareFetch(valueResource),
+      this.cloudflareFetch(metadataResource),
+    ]);
```
penalosa, line 136: I wish there was a unified endpoint for this

packages/tre/src/plugins/kv/remote.ts
```diff
@@ -0,0 +1,273 @@
...
+    // This handles the 404 case when a key isn't found
```
penalosa, line 137: It throws an error—is that interpreted as `undefined` somewhere else down the line?

packages/tre/src/plugins/kv/remote.ts
```diff
@@ -0,0 +1,273 @@
...
+    const expirationHeader = valueResponse.headers.get("Expiration");
+    let expiration: number | undefined;
+    if (expirationHeader !== null) {
+      const maybeExpiration = parseInt(expirationHeader);
+      if (!isNaN(maybeExpiration)) expiration = maybeExpiration;
+    }
+
+    // ...and cache it for the specified TTL, then return it
+    const result: StoredValueMeta = { value, expiration, metadata };
+    await this.cache.put<RemoteCacheMetadata>(key, {
```
penalosa, line 159: Does this need to be `await`ed?

packages/tre/src/storage/sqlite.ts
```diff
@@ -79,7 +79,6 @@ export class SqliteStorage extends LocalStorage {
   }
 
   private run(query: ReturnType<typeof sql>): { changes: number } {
-    console.log(query);
```
penalosa, line 82: I need to stop leaving `console.log` in things :joy:

JacobMGEvans (approved):

packages/tre/src/plugins/kv/remote.ts
mrbbot: Yep, I think you're right 👍

packages/tre/src/plugins/kv/remote.ts
mrbbot: Right? I think a `Metadata` header like in the `workerd` KV HTTP interface would make sense.

packages/tre/src/plugins/kv/remote.ts
mrbbot: Ah you're right, it has the expected behaviour from `workerd`, but does not conform to the `Storage` contract. 👍

packages/tre/src/plugins/kv/remote.ts
mrbbot: I guess not, but then if it threw for whatever reason we'd end up with an unhandled rejection. Given the local cache is definitely not going to be the bottleneck in this function, I think we should keep it `await`ed.

packages/tre/src/storage/sqlite.ts
mrbbot: I need to too 😅

penalosa (approved):

PR #431, opened 2022-11-14
Allow access to private networks

mrbbot:


penalosa:
Was going to leave a comment and then saw https://github.com/cloudflare/workerd/issues/62—it's great that `0.0.0.0/0` works as a workaround, but is it worth fixing this in the runtime?

JacobMGEvans (approved):

dom96:
Made a PR to fix it in the runtime https://github.com/capnproto/capnproto/pull/1576 :)

PR #433, opened 2022-11-17
Assorted Cache API fixes

mrbbot:
Each of these fixes is split into a separate commit for easier reviewing. 🙂 
- **Enable persistence:** previously the `cachePersist` option was being ignored, so cached data couldn't be persisted
- **Disable Workers Sites asset caching:** we always want to serve the latest on-disk Workers Sites files so changes are reflected. However, `@cloudlfare/kv-asset-handler` automatically caches files, so we need to disable the Cache API for these URIs.
- **Allow cache to be disabled:** Miniflare 2 had a `cache: false` option for disabling the cache. The Cache API was still available, it just didn't do anything.
- **Facilitate cache usage warning on `workers.dev` subdomains:** Cache API operations are no-ops on `workers.dev` subdomains. This is a common cause of confusion, so this PR also brings over Miniflare 2's `cacheWarnUsage: true` option. We can set this in Wrangler if we detect the user will be publishing to a `workers.dev` zone.

### DIFF
--- a/packages/tre/src/plugins/cache/index.ts
+++ b/packages/tre/src/plugins/cache/index.ts
@@ -11,6 +11,7 @@ import {
   Plugin,
   encodePersist,
 } from "../shared";
+import { HEADER_CACHE_WARN_USAGE } from "./constants";
 import { CacheGateway } from "./gateway";
 import { CacheRouter } from "./router";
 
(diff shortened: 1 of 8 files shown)


packages/tre/src/plugins/cache/index.ts
```diff
@@ -13,7 +14,12 @@ import {
 import { CacheGateway } from "./gateway";
 import { CacheRouter } from "./router";
 
-export const CacheOptionsSchema = z.object({
+export const CacheOptionsSchema = z.object({});
+export const CacheSharedOptionsSchema = z.object({
+  cachePersist: PersistenceSchema,
+  // Ideally, these options would be configurable per-worker (i.e. part of
+  // `CacheOptionsSchema` instead). However, `workerd` can only have one global
+  // "cache" service, so we can't distinguish which worker called the Cache API.
```
penalosa, line 23: I may be misremembering, but I’m pretty sure the cacheApiOutbound property in the capnp file that controls which Cache service to look at is per-worker, not global. Regardless, not worth supporting for right now

penalosa (approved):
Lgtm 🎉

packages/tre/src/plugins/cache/index.ts
mrbbot: Turns out it was pretty easy to make this work per-worker: ad80536 🙂

PR #434, opened 2022-11-17
Add configurable logger

mrbbot:
This PR copies the `Log` implementation from Miniflare 2, and allows logging to be configured via the `log` shared option. Even though none of our gateways need a logger at the moment, I've added it in case we need it later.

### DIFF
--- a/packages/tre/src/shared/log.ts
+++ b/packages/tre/src/shared/log.ts
@@ -1,0 +1,123 @@
+import path from "path";
+import { Colorize, dim, green, grey, red, reset, yellow } from "kleur/colors";
+
+const cwd = process.cwd();
+const cwdNodeModules = path.join(cwd, "node_modules");
+
+export enum LogLevel {
+  NONE,
+  ERROR,
+  WARN,
+  INFO,
+  DEBUG,
+  VERBOSE,
+}
+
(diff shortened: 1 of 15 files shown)


packages/tre/src/shared/log.ts
```diff
@@ -0,0 +1,123 @@
...
+};
+
+const LEVEL_COLOUR: { [key in LogLevel]: Colorize } = {
+  [LogLevel.NONE]: reset,
+  [LogLevel.ERROR]: red,
+  [LogLevel.WARN]: yellow,
+  [LogLevel.INFO]: green,
+  [LogLevel.DEBUG]: grey,
+  [LogLevel.VERBOSE]: (input) => dim(grey(input as any)) as any,
+};
```
JacobMGEvans, line 32: **Non blocking** but Enums are rough on the TS server and compiler and turn into functions at runtime. 
Could use something like an Object literal or Map for the types?

JacobMGEvans (approved):
Looks like solid OOP code to me 😆

packages/tre/src/shared/log.ts
mrbbot: In the interest of minimising breaking API changes from Miniflare 2, I think I'll keep this as an `enum`, but agree, a `const` object literal would be nice here.

PR #435, opened 2022-11-17
Add support for built-in `workerd` services

mrbbot:
Specifically, this allows custom `network`, `external` and `disk` services to be specified as service bindings. Requests to `network` services are dispatched according to URL. Requests to `external` services are dispatched to the specified remote server. Requests to `disk` services are dispatched to an HTTP service backed by an on-disk directory.

This PR also fixes a bug where custom function service bindings with the same name in different Workers would dispatch all requests to the first Workers' function.

### DIFF
--- a/packages/tre/src/plugins/core/services.ts
+++ b/packages/tre/src/plugins/core/services.ts
@@ -1,0 +1,83 @@
+import { Request, Response } from "undici";
+import { z } from "zod";
+import {
+  ExternalServer,
+  HttpOptions_Style,
+  TlsOptions_Version,
+} from "../../runtime";
+import { zAwaitable } from "../../shared";
+
+// Zod validators for types in runtime/config/workerd.ts.
+// All options should be optional except where specifically stated.
+// TODO: autogenerate these with runtime/config/workerd.ts from capnp
+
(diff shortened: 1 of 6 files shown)


JacobMGEvans (approved):

packages/tre/src/plugins/core/services.ts
```diff
@@ -0,0 +1,83 @@
...
+        })
+      ),
+    }),
+  ])
+) as z.ZodType<ExternalServer>;
+// This type cast is required for `api-extractor` to produce a `.d.ts` rollup.
+// Rather than outputting a `z.ZodIntersection<...>` for this type, it will
+// just use `z.ZodType<ExternalServer>`. Without this, the extractor process
+// just ends up pinned at 100% CPU. Probably unbounded recursion? I guess this
+// type is too complex? Something to investigate... :thinking_face:
```
penalosa, line 65:
```suggestion
// type is too complex? Something to investigate... 🤔 
```

penalosa (approved):

packages/tre/src/plugins/core/services.ts
mrbbot: 🤔 🤔 🤔

PR #436, opened 2022-11-18
Add pretty-error page infrastructure

mrbbot:
This adds Miniflare 2's pretty-error page powered by [Youch](https://github.com/poppinss/youch) to Miniflare 3.

Unfortunately, due to a bug in `workerd`, errors thrown asynchronously by native APIs don't have `stack`s. This means we can't extract the `stack` trace from dispatching to the user worker by `try`/`catch`.

As a stop-gap solution, if the `MF-Experimental-Error-Stack` exists and is truthy on the response from the user worker, the body will be interpreted as a JSON-error of the form
`{ message?: string, name?: string, stack?: string }`. `stack` will be source-mapped if possible.

Another issue is that `workerd` gives all service-worker scripts the name "worker.js", so if multiple service-workers are defined, we can't identify which one threw. In this case, we don't display sources in the pretty-error page.

Hopefully, we can fix both of these issues in `workerd`. We should be able to reuse most of this infrastructure with `try`/`catch`s too.


penalosa (approved):

thetutlage:
Hello guys!

I am the creator of Youch and did a rewrite of the package with newer/better design and more extensible API.

I have tagged the beta release of Youch and the README covers it usage in-depth. https://github.com/poppinss/youch

Would love to collaborate with you guys and see if you need anything to make Youch a better fit for the Workers SDK

PR #439, opened 2022-11-22
Fix accidental destructure of string binding names

sgentle:
Hi!

Was just having a play around and I found a lil bug in the handling of simple `string[]`-style bindings in namespaceEntries. 

An extra pair of square brackets snuck in there, causing bindings like `r2Buckets: ['hello']` to bind as 'h' because the string is destructured like an array:
```js
> ['hello'].map(([x]) => [x, x])
[ [ 'h', 'h' ] ] 
```

Types didn't catch it 'cause technically that's a super chill and valid thing to do with your string arguments 🤷‍♂️


mrbbot (approved):
Great catch! Thanks for fixing this. 😃

PR #440, opened 2022-11-22
Export Fetch API

### LINKED ISSUE #438: Consider relax zod-checking of service bindings?
In `@miniflare/tre`, we're using Zod to check the types of any service bindings functions. Unless we resolve to the same undici version, `@miniflare/tre` gets its own, and our custom service binding will fail at these checks.

Either `@miniflare/tre` should export Fetch API things (e.g. `Request` and `Response`), or we should consider relaxing the Zod checking. Using Node's new built-ins seems to work just fine, so maybe these checks are actually not really necessary at all?

GregBrimble:
Fixes #438

### DIFF
--- a/packages/tre/src/index.ts
+++ b/packages/tre/src/index.ts
@@ -6,11 +6,25 @@ import exitHook from "exit-hook";
 import getPort from "get-port";
 import stoppable from "stoppable";
 import {
+  BodyInit,
+  File,
+  FormData,
+  Headers,
   HeadersInit,
+  ReferrerPolicy,
   Request,
+  RequestCache,
+  RequestCredentials,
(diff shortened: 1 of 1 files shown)


packages/tre/src/index.ts
```diff
@@ -614,3 +617,4 @@ export * from "./plugins";
 export * from "./runtime";
 export * from "./shared";
 export * from "./storage";
+export { FormData, File, Headers, Request, Response, fetch };
```
mrbbot, line 620: Might be useful to export the types here too, just so you can do `import { Headers, HeadersInit } from "..."` without having to `import ... from "undici"` too?

mrbbot (approved):

PR #443, opened 2022-11-22
Fix WSL support, closes cloudflare/wrangler2#2243

mrbbot:
Previously, Miniflare passed Windows paths to WSL, without first converting them to WSL-friendly `/mnt/...` paths. This PR fixes that, WSL detection, and also adds a new restart script similar to the Docker runtime to massively reduce script-reload time.

### DIFF
--- a/packages/tre/lib/docker-restart.sh
+++ b/packages/tre/lib/docker-restart.sh
@@ -1,6 +1,7 @@
 #!/usr/bin/env bash
 # Restarts a process on receiving SIGUSR1.
 # Usage: ./restart.sh <command> <...args>
+set -eo pipefail
 
 # Start process and record its PID
 "$@" &
--- a/packages/tre/src/index.ts
+++ b/packages/tre/src/index.ts
@@ -265,7 +265,13 @@ export class Miniflare {
     };
     this.#runtime = new this.#runtimeConstructor(opts);
     this.#removeRuntimeExitHook = exitHook(() => void this.#runtime?.dispose());
-    this.#runtimeEntryURL = new URL(`http://127.0.0.1:${opts.entryPort}`);
+
+    const accessibleHost =
+      this.#runtime.accessibleHostOverride ??
+      (host === "*" || host === "0.0.0.0" ? "127.0.0.1" : host);
+    this.#runtimeEntryURL = new URL(
+      `http://${accessibleHost}:${opts.entryPort}`
+    );
 
     const config = await this.#assembleConfig();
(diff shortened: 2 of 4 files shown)


packages/tre/lib/docker-restart.sh
```diff
@@ -1,6 +1,7 @@
 #!/usr/bin/env bash
 # Restarts a process on receiving SIGUSR1.
 # Usage: ./restart.sh <command> <...args>
+set -eo pipefail
 
 # Start process and record its PID
 "$@" &
```
penalosa, line 7: Fix permissions

packages/tre/src/index.ts
```diff
@@ -265,7 +265,7 @@ export class Miniflare {
     };
     this.#runtime = new this.#runtimeConstructor(opts);
     this.#removeRuntimeExitHook = exitHook(() => void this.#runtime?.dispose());
-    this.#runtimeEntryURL = new URL(`http://127.0.0.1:${opts.entryPort}`);
+    this.#runtimeEntryURL = new URL(`http://localhost:${opts.entryPort}`);
```
penalosa, line 268: Use `host` here, and further investigation of why `127.0.0.1 isn't working

packages/tre/src/runtime/index.ts
```diff
@@ -143,25 +138,107 @@ class NativeRuntime extends Runtime {
...
+
+    // 3. Otherwise, start new process
+    const runtimeProcess = childProcess.spawn(
+      WSL_EXE_PATH,
+      [
+        "--exec",
+        WSLRuntime.pathToWSL(WSL_RESTART_PATH),
+        WSLRuntime.pathToWSL(workerdPath),
+        ...this.getCommonArgs(),
+        `--socket-addr=${SOCKET_ENTRY}=*:${this.opts.entryPort}`,
```
penalosa, line 211: Use the `entryHost` option

penalosa (approved):

JacobMGEvans (approved):

packages/tre/src/index.ts
mrbbot: Seems like this is all that works 😕 https://learn.microsoft.com/en-us/windows/wsl/wsl-config#configuration-setting-for-wslconfig

packages/tre/src/runtime/index.ts
mrbbot: Again, seems like this is all that works. 😢

PR #445, opened 2022-11-23
Disable `shell: true` when using `NativeRuntime`

mrbbot:
When `dispose()`ing a `NativeRuntime`, I've observed the `workerd` process not being terminated properly with this option enabled. We don't actually need it, since we're specifying an absolute path to the `workerd` binary anyway.

From a user-perspective, this will help fix issues in Wrangler where reloading crashes with a port-in-use error, or where pressing 'x' doesn't actually exit.


penalosa (approved):
🎉

PR #447, opened 2022-11-25
Cache API tests

mrbbot:
This PR adds testing for Miniflare 3's Cache API. I've written these as integration tests. Along the way, I fixed some bugs with caching streamed responses, and using `cf.cacheKey`.

This PR is pretty big, because it also adds lots of the required integration testing infrastructure. Hopefully, it should be easier to add tests for other features after this.

### DIFF
--- a/packages/tre/src/plugins/cache/gateway.ts
+++ b/packages/tre/src/plugins/cache/gateway.ts
@@ -207,11 +207,12 @@ export class CacheGateway {
     private readonly clock: Clock
   ) {}
 
-  async match(request: Request): Promise<Response> {
+  async match(request: Request, cacheKey?: string): Promise<Response> {
     // Never cache Workers Sites requests, so we always return on-disk files
     if (isSitesRequest(request)) throw new CacheMiss();
 
-    const cached = await this.storage.get<CacheMetadata>(request.url);
+    cacheKey ??= request.url;
+    const cached = await this.storage.get<CacheMetadata>(cacheKey);
     if (cached?.metadata === undefined) throw new CacheMiss();
 
     const response = new CacheResponse(
@@ -228,7 +229,11 @@ export class CacheGateway {
     );
   }
 
-  async put(request: Request, value: ArrayBuffer): Promise<Response> {
+  async put(
+    request: Request,
+    value: ArrayBuffer,
+    cacheKey?: string
+  ): Promise<Response> {
     // Never cache Workers Sites requests, so we always return on-disk files
     if (isSitesRequest(request)) return new Response(null, { status: 204 });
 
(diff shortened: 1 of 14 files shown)


packages/tre/src/plugins/cache/gateway.ts
```diff
@@ -246,7 +253,8 @@ export class CacheGateway {
       throw new StorageFailure();
     }
 
-    await this.storage.put<CacheMetadata>(request.url, {
+    cacheKey ??= request.url;
```
JacobMGEvans, line 256: Clever

JacobMGEvans (approved):

packages/tre/src/plugins/cache/gateway.ts
```diff
@@ -207,11 +209,12 @@ export class CacheGateway {
     private readonly clock: Clock
   ) {}
 
-  async match(request: Request): Promise<Response> {
+  async match(request: Request, cacheKey?: string): Promise<Response> {
```
penalosa, line 212: I could be wrong, but I don't believe `workerd` ever sends a `cacheKey`—I think the request URI should always be treated as the key?

packages/tre/src/plugins/cache/gateway.ts
```diff
@@ -228,11 +231,15 @@ export class CacheGateway {
     );
   }
 
-  async put(request: Request, value: ArrayBuffer): Promise<Response> {
+  async put(
+    request: Request,
+    value: Uint8Array,
+    cacheKey?: string
```
penalosa, line 237: As above

packages/tre/src/plugins/cache/gateway.ts
mrbbot: Doesn't look like it. Adding...

```diff
--- a/packages/tre/src/plugins/cache/router.ts
+++ b/packages/tre/src/plugins/cache/router.ts
@@ -134,6 +134,7 @@ export class CacheRouter extends Router<CacheGateway> {
     const namespace = decodeNamespace(req.headers);
     const persist = decodePersist(req.headers);
     const cf = decodeCfBlob(req.headers);
+    console.log("GET", { uri, cf });
     const gateway = this.gatewayFactory.get(namespace, persist);
     const key = new Request(uri, req as RequestInit);
     return fallible(gateway.match(key, cf.cacheKey));
@@ -146,6 +147,7 @@ export class CacheRouter extends Router<CacheGateway> {
     const namespace = decodeNamespace(req.headers);
     const persist = decodePersist(req.headers);
     const cf = decodeCfBlob(req.headers);
+    console.log("PUT", { uri, cf });
     const gateway = this.gatewayFactory.get(namespace, persist);
     const bodyBuffer = Buffer.from(await req.arrayBuffer());
     const bodyArray = new Uint8Array(removeTransferEncodingChunked(bodyBuffer));
```

...and running...

```js
const key = new Request("http://localhost/path", { cf: { cacheKey: "key" } });
const res = new Response("body", {
  headers: { "Cache-Control": "max-age=3600" },
});
await caches.default.put(key, res);
const cachedRes = await caches.default.match(key);
return cachedRes ?? new Response("<miss>");
```

...in a worker, logs...

```
PUT { uri: 'http://localhost/path', cf: { cacheKey: 'key' } }
GET { uri: 'http://localhost/path', cf: { cacheKey: 'key' } }
```

penalosa (approved):

PR #449, opened 2022-11-29
fix(response): reset bodyStream when body is cloned

DSergiu:
## What
`Response.clone()` behaves differently than in the actual Workers or in Browser.

## Why
When response body gets cloned, the original `body.stream` is set to one of the cloned streams.
Miniflare's `Response` keeps an internal `bodyStream` reference which is not reset whenever the original `body.stream` is reset.

Thus, whenever you get a reference of the `body` before cloning it and attempting to loop over the body after cloning it, you get this error message: `TypeError: Invalid state: ReadableStream is locked` . 

[See code example that reproduces the issue here](https://github.com/cloudflare/miniflare/pull/449/files#diff-963ba9d1eeb501b4be74716fd399a23eaf366c36fd0af770154b6adf8c389e8cR167-R179).

```js
const res = new Response(new ArrayBuffer(10));
const bodyStream = res.body;
// Clone the response. undici will change the `body.stream` to a new clone.
res.clone();
// We should be able to loop over body. This is what http-server writeResponse() does.
for await (const chunk of res.body) { // Since `body` points to old/locked/used `bodyStream`, attempting to `stream.getReader()` would fail
  assert(chunk instanceof Uint8Array);
}
```

See undici code [here](https://github.com/nodejs/undici/blob/671e05fb90d4e047a3b219d802a8afeeb35ad883/lib/fetch/body.js#L291-L298).

```js
function cloneBody (body) {
  // To clone a body body, run these steps:

  // https://fetch.spec.whatwg.org/#concept-body-clone

  // 1. Let « out1, out2 » be the result of teeing body’s stream.
  const [out1, out2] = body.stream.tee()
  const out2Clone = structuredClone(out2, { transfer: [out2] })
  // This, for whatever reasons, unrefs out2Clone which allows
  // the process to exit by itself.
  const [, finalClone] = out2Clone.tee()

  // 2. Set body’s stream to out1.
  body.stream = out1

  // 3. Return a body whose stream is out2 and other members are copied from body.
  return {
    stream: finalClone,
    length: body.length,
    source: body.source
  }
}
```

## How

Reset internal `bodyStream` when `Response.clone()`.

### DIFF
--- a/packages/core/src/standards/http.ts
+++ b/packages/core/src/standards/http.ts
@@ -157,6 +157,7 @@ export function _headersFromIncomingRequest(
 /** @internal */
 export const _kInner = Symbol("kInner");
 
+const kBodyStream = Symbol("kBodyStream");
 const kInputGated = Symbol("kInputGated");
 const kFormDataFiles = Symbol("kFormDataFiles");
 const kCloned = Symbol("kCloned");
@@ -168,7 +169,7 @@ export class Body<Inner extends BaseRequest | BaseResponse> {
   [kInputGated] = false;
   [kFormDataFiles] = true; // Default to enabling form-data File parsing
   [kCloned] = false;
-  #bodyStream?: ReadableStream<Uint8Array>;
+  [kBodyStream]?: ReadableStream<Uint8Array>;
 
   constructor(inner: Inner) {
     // Allow forbidden header mutation after construction
@@ -191,16 +192,13 @@ export class Body<Inner extends BaseRequest | BaseResponse> {
     return this[_kInner].headers;
   }
 
-  set bodyStream(value: ReadableStream | undefined) {
-    this.#bodyStream = value;
-  }
-
   get body(): ReadableStream<Uint8Array> | null {
     const body = this[_kInner].body;
 
     if (body === null) return body;
     // Only transform body stream once
-    if (this.#bodyStream) return this.#bodyStream;
+    const bodyStream = this[kBodyStream];
+    if (bodyStream) return bodyStream;
     assert(body instanceof ReadableStream);
 
     // Cloudflare Workers allows you to byob-read all Request/Response bodies,
@@ -213,7 +211,7 @@ export class Body<Inner extends BaseRequest | BaseResponse> {
     // If we're not input gating, and body is already a byte stream, we're set,
     // just return it as is (this will be the case for incoming http requests)
     if (!this[kInputGated] && _isByteStream(body)) {
-      return (this.#bodyStream = body);
+      return (this[kBodyStream] = body);
     }
 
     // Otherwise, we need to create a "byte-TransformStream" that makes sure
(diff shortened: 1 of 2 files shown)


DSergiu:
Tagging @mrbbot @a-robinson for visibility.

packages/core/src/standards/http.ts
```diff
@@ -584,6 +588,7 @@ export class Response<
       throw new TypeError("Cannot clone a response to a WebSocket handshake.");
     }
     const innerClone = this[_kInner].clone();
+    this.bodyStream = undefined;
```
mrbbot, line 591: Could you replace this with a write to `this.#bodyStream` instead, then remove the `bodyStream` setter?
```suggestion
    this.#bodyStream = undefined;
```

packages/core/test/standards/http.spec.ts
```diff
@@ -150,6 +155,35 @@ test("Body: body isn't locked until read from", async (t) => {
...
+  t.true(res.body.locked);
+});
+test("Body: should reset bodyStream when body is cloned", async (t) => {
+  const res = new Response(new ArrayBuffer(10));
+  // noinspection SuspiciousTypeOfGuard
+  t.true(res instanceof Body);
+  const bodyStream = res.body;
+  assert(bodyStream instanceof ReadableStream);
+  // Clone the response. undici will change the `body.stream` to a new clone.
+  res.clone();
```
mrbbot, line 177: Could you also check we're able to read from the cloned `Response`'s `body`?

mrbbot (commented):
Hey! 👋 Thanks for the tag, and fixing this! I've added a couple comments. 👍

packages/core/src/standards/http.ts
DSergiu: I remember accessing the private var `#bodyStream` was not possible.

packages/core/test/standards/http.spec.ts
DSergiu: Testing cloned response's body is already tested by other tests. This test was meant to test that the original response is not affected after it is cloned.

But I guess there is no harm in stricter assertions. Here it is https://github.com/cloudflare/miniflare/pull/449/commits/4733197e8e142acb7b60c3cf3cbd5a428bbe7dd7

packages/core/src/standards/http.ts
mrbbot: Oh I see sorry! Could you replace `#bodyStream` with a `Symbol` like `kCloned` then? That way end users won't be able to access `bodyStream`. 🙂

DSergiu: Sounds like a "protected" modifier would be a good choice for `bodyStream`. 

Feel free to extend this PR if you'd like.

mrbbot: The problem with `protected` is that the restriction only applies when people are using the Miniflare API with TypeScript. Inside Worker code, developers are free to access `protected` members as the access check is only enforced during type checking and not at runtime. Happy to make that change though. 👍

DSergiu: I see what you mean. The issue seems to be that ECMAScript introduced hard privacy via hash `#` but there is no protected modifier (yet?).

If relying on TypeScript is not possible for the clients, it is convention to prefix props with `_` which are mean for internal use.

I did push a fix as suggested, using `Symbol`. See https://github.com/cloudflare/miniflare/pull/449/commits/0f581864540615e291def81ae2d547536243e96b

DSergiu:
> Is this an issue with `Request` bodies too?

Could be, I did not encounter an issue yet.

packages/core/src/standards/http.ts
DSergiu: @mrbbot I see the pipeline didn't run. Would you be able to start it?

PR #455, opened 2022-12-08
fix: rename queue binding to match runtime

zebp:
Renames the queue binding in the miniflare plugin to `WorkerQueue` to match the name in the Cloudflare runtime. This small discrepancy is particularly important for the [workers-rs](https://github.com/cloudflare/workers-rs) framework which [validates the types of bindings at runtime based on their constructor name](https://github.com/cloudflare/workers-rs/blob/7a78ac9cd6be4d27d6fc435fa8dc60d59d55405d/worker/src/env.rs#L66), making it impossible to use queues with miniflare.

### DIFF
--- a/packages/queues/src/broker.ts
+++ b/packages/queues/src/broker.ts
@@ -89,7 +89,7 @@ enum FlushType {
 
 export const kSetFlushCallback = Symbol("kSetFlushCallback");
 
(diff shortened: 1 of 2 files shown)


mrbbot (approved):
Great catch! Thank you!

PR #466, opened 2023-01-06
Implement `WebSocket` constructor

mrbbot:
The Workers runtime has supported the `new WebSocket` constructor, as an alternative to the `fetch` API for WebSocket clients, since February 2022. This PR adds support for it to Miniflare.


packages/web-sockets/src/websocket.ts
```diff
@@ -248,8 +332,71 @@ export const WebSocketPair: { new (): WebSocketPair } = function (
...
   }
-  this[0] = new WebSocket();
-  this[1] = new WebSocket();
+  this[0] = new WebSocket(kConstructOnly);
+  this[1] = new WebSocket(kConstructOnly);
   this[0][kPair] = this[1];
   this[1][kPair] = this[0];
 } as any;
+
+export async function coupleWebSocket(
```
mrbbot, line 341: This has been moved to avoid circular `import`s with `websocket.ts` and `couple.ts`.

packages/web-sockets/src/websocket.ts
```diff
@@ -48,22 +52,27 @@ export class ErrorEvent extends Event {
 // construct both sockets before setting the circular references
 const kPair = Symbol("kPair");
 
-export const kAccepted = Symbol("kAccepted");
-export const kCoupled = Symbol("kCoupled");
```
penalosa, line 52: Is this an important detail, or are these just unnecessary exports?

packages/web-sockets/src/websocket.ts
```diff
@@ -76,13 +85,73 @@ export class WebSocket extends InputGatedEventTarget<WebSocketEventMap> {
...
+
   #dispatchQueue?: ValueOf<WebSocketEventMap>[] = [];
   [kPair]: WebSocket;
   [kAccepted] = false;
   [kCoupled] = false;
   [kClosedOutgoing] = false;
   [kClosedIncoming] = false;
 
+  constructor(url: string | URL, protocols?: string | string[]);
+  constructor(flag: typeof kConstructOnly);
```
penalosa, line 100: I think this would be clearer (and more idiomatic) as a static method on WebSocket rather than a flag passed to the constructor, if that would be possible?

penalosa (approved):
LGTM

packages/web-sockets/src/websocket.ts
mrbbot: I don't think this is possible. If we had a static method, we'd still need to call `new WebSocket()` in it to actually get a `WebSocket` instance right? Then we have the exact same problem.

mrbbot: The problem is `new WebSocket()` needs to be publicly exposed to users to match the spec, so we can't make this private and have multiple static constructors.

penalosa: Ah, I see, that makes sense—I didn't quite realise this was directly exposed

PR #470, opened 2023-01-10
Assorted R2 fixes

mrbbot:
This PR implements a bunch of new R2 features to improve compatibility with the real Workers runtime. Best reviewed commit by commit.

@Frederik-Baetens if you've got cycles, it would be great if you could review this too, from an R2 perspective. 🙂

### DIFF
--- a/packages/r2/src/r2Object.ts
+++ b/packages/r2/src/r2Object.ts
@@ -1,9 +1,10 @@
+import assert from "assert";
 import { Blob } from "buffer";
 import crypto from "crypto";
 import { arrayBuffer } from "stream/consumers";
 import { ReadableStream } from "stream/web";
 import { TextDecoder } from "util";
-import { waitForOpenInputGate } from "@miniflare/shared";
+import { viewToBuffer, waitForOpenInputGate } from "@miniflare/shared";
 import { Headers } from "undici";
 import { R2Conditional, R2Range } from "./bucket";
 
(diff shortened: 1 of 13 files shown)


packages/r2/test/r2Object.spec.ts
```diff
@@ -496,3 +503,36 @@ test("R2Object: R2ObjectBody: push 'body' ReadableStream to Response", async (t)
...
+  assert(buffer !== undefined);
+  new MessageChannel().port1.postMessage(buffer, [buffer]);
+  t.is(buffer.byteLength, 0); // (detached)
+  t.is(checksums.md5?.byteLength, 16); // New buffer
+});
+test("Checksums: returns hex hashes with JSON.stringify()", async (t) => {
+  const checksums = new Checksums(stringChecksums);
+  t.deepEqual(checksums.toJSON(), stringChecksums);
+  t.is(JSON.stringify(checksums), JSON.stringify(stringChecksums));
+});
```
Frederik-Baetens, line 538: I'm so jealous of how easy it is to write miniflare tests compared to EW tests :cry:

packages/r2/src/r2Object.ts
```diff
@@ -253,6 +316,22 @@ export class R2Object {
     this.httpMetadata = metadata.httpMetadata;
     this.customMetadata = metadata.customMetadata;
     this.range = metadata.range;
+
+    // We always need to store an MD5 hash in `checksums`, but never explicitly
+    // stored one. Luckily, `R2Bucket#put()` always makes `etag` an MD5 hash.
```
Frederik-Baetens, line 321: Internally in R2, we do the same exact thing due to historic reasons 🤫

Frederik-Baetens (approved):
Very nice work, Bit insane that you have to duplicate this work from workerd though. Hope miniflare3 will improve that :)

PR #478, opened 2023-01-19
`nodejs_compat` flag

GregBrimble:
Adds the `nodejs_compat` flag and attaches the `node:async_hooks` partial implementation that is now in `workerd`.


packages/core/src/plugins/core.ts
```diff
@@ -1,5 +1,6 @@
 import { Blob } from "buffer";
 import fs from "fs/promises";
+import asyncHooks from "node:async_hooks";
```
mrbbot, line 3: Looks like this is never used.

mrbbot (approved):
Nice! 🎉 One tiny lint fix, then I'll merge... 😃

PR #474, opened 2023-01-15
Fix .raw() not executing statement in D1

repository:
Fixes https://github.com/cloudflare/wrangler2/issues/2238, hopefully

PR #480, opened 2023-01-23
Implement `D1Database` using D1JS and D1 API

### LINKED ISSUE #441: 🐛 BUG: .run() throws D1 Error in cloud
### What version of `Wrangler` are you using?

2.4.2

### What operating system are you using?

Mac

### Describe the Bug

```js
await env.DB.prepare('insert into users (name, email) values (?, ?) returning *').bind(name, email).run();
```

throws `Exception` in cloud

```js
"exceptions": [
    {
      "name": "Error",
      "message": "D1_ERROR",
      "timestamp": 1668795822085
    }
  ]
```

These three below work as expected in cloud

```js
await env.DB.prepare('insert into users (name, email) values (?, ?) returning *').bind(name, email).all();
```
```js
await env.DB.prepare('insert into users (name, email) values (?, ?) returning *').bind(name, email).first();
```
```js
await env.DB.prepare('insert into users (name, email) values (?, ?) returning *').bind(name, email).raw();
```
All four requests do work as expected locally

### LINKED ISSUE #442: BUG: Different responses when running a D1 query in the cloud and the local mode
### What version of `Wrangler` are you using?

2.4.2

### What operating system are you using?

Linux

### Describe the Bug

Hello! Consider the following code:

```javascript
export default {
  async fetch(
    request: Request,
    env: Env,
    ctx: ExecutionContext
  ): Promise<Response> {
    await env.db.exec(
      "create table if not exists foos (value text);"
    );

    const result = await env.db.prepare(
      "insert into foos (value) values (?)"
    ).bind("from param").run();

    return new Response(JSON.stringify(result));
  },
};
```

When ran in the cloud, the response looks like this:

```json
{
    "meta": {
        "changes": null,
        "duration": 9.485032001510262,
        "internal_stats": null,
        "last_row_id": null,
        "served_by": "primary-78d95a04-af3b-430b-8bd5-0196cf0c6331.db3"
    },
    "results": [],
    "success": true
}
```

However then running in the local mode (`npm start -- --local --persist`), the response is different:

```json
{
    "changes": 1,
    "duration": 15.273903001099825,
    "lastRowId": 17,
    "results": null,
    "served_by": "x-miniflare.db3",
    "success": true
}
```

mrbbot:
Previously, Miniflare had its own implementations of `BetaDatabase` and `Statement`. These were subtly different to the implementations in the D1 Wrangler shim D1JS, causing behaviour mismatches.

This change switches the implementation to use the shim, with Miniflare implementing the underlying D1 HTTP API instead. We'll need to do this anyway when adding D1 support to Miniflare 3 using `workerd`.

#### Specific changes:
- Throw when calling `D1PreparedStatement#run()` with statements that return data, closes #441
- Fix response envelope format, closes #442 and cloudflare/wrangler2#2504
- Fix binding/return of `BLOB`-typed values, closes cloudflare/wrangler2#2527
- Fix `D1Database#raw()` return, closes cloudflare/wrangler2#2238 (already fixed in #474)
- Add support for `D1Database#dump()`
- Run `D1Database#{batch,exec}()` statements in implicit transaction
- Only run first statement when calling `D1PreparedStatement#{first,run,all,raw}()`


ruslantalpa:
@mrbbot 
I think this PR broke queries like `select ?1 + ?2`

`statements.ts` used to have this code
```
      // For statements using ?1 ?2, etc, we want to pass them as varargs but
      // "better" sqlite3 wants them as an object of {1: params[0], 2: params[1], ...}
      if (this.#bindings.length > 0 && typeof this.#bindings[0] !== "object") {
        return prepared.bind(
          Object.fromEntries(this.#bindings.map((v, i) => [i + 1, v]))
        );
      } else {
        throw e;
      }
```

ruslantalpa:
i added these lines to one of the tests (to make sure it's not my code):
```
const sum = await db.prepare("SELECT ?1 + ?2 as sum").bind(1, 2).first("sum");
t.is(sum, 3);
```

and the test fails
```
d1 › test › d1js.ts › D1PreparedStatement: first

  packages/d1/test/d1js.spec.ts:212

   211:
   212:   const sum = await db.prepare("SELECT ?1 + ?2 as sum").bind(1, 2).first("sum");
   213:   t.is(sum, 3);

  Rejected promise returned by test. Reason:

  Error {
    message: 'D1_ERROR',
  }

  › D1Database._send (packages/d1/src/d1js.ts:148:13)
  › D1PreparedStatement.first (packages/d1/src/d1js.ts:215:7)
  › packages/d1/test/d1js.spec.ts:212:15

  ─
```

PR #486, opened 2023-01-30
Implement R2 multipart upload bindings

mrbbot:
This PR implements support for writing data to R2 buckets in multiple parts. See https://developers.cloudflare.com/r2/data-access/workers-api/workers-multipart-usage/. Miniflare's implementation is briefly described at the top of `packages/r2/src/multipart.ts`.

Unfortunately, Miniflare 2's storage abstraction is not very good at handling large data (doesn't support streaming reads/writes), or complex read/write operations (doesn't support transactions). This limits the reliability of this multipart implementation, but it should still be useful for testing. We should aim to improve this in Miniflare 3.

Whilst implementing this feature, I noticed a few quirks with R2:

- `httpMetadata` and `customMetadata` are always `{}` in the `R2Object` returned from `R2MultipartUpload#complete()`
- `etag`s for completed multipart uploads with the same part contents are not the same. I'd expect these to be deterministic.

### DIFF
--- a/packages/r2/src/multipart.ts
+++ b/packages/r2/src/multipart.ts
@@ -102,8 +102,15 @@ function validatePartNumber(partNumber: number) {
   );
 }
 
-function generateId() {
-  return crypto.randomBytes(128).toString("base64url");
+function generateId(likelyOnFilesystem = false) {
+  // Windows has a maximum path length of ~260 characters:
+  // https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation
+  // Miniflare R2 buckets will usually be backed by file-system storage,
+  // especially when using multipart uploads for large files. Therefore, reduce
+  // the size of the upload ID on Windows, preferring a longer ID otherwise
+  // to more closely match R2 behaviour.
+  const size = likelyOnFilesystem && process.platform === "win32" ? 32 : 128;
+  return crypto.randomBytes(size).toString("base64url");
 }
 function generateMultipartEtag(md5Hexes: string[]) {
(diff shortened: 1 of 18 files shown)


Frederik-Baetens:
> httpMetadata and customMetadata are always {} in the R2Object returned from R2MultipartUpload#complete()

Ah I know why that is :sweat_smile: That's not intentional and I'll get on fixing that

> etags for completed multipart uploads with the same part contents are not the same. I'd expect these to be deterministic.

We're aware of this, and we're considering changing how the etag for multipart uploads is calculated to be compatible with how S3 does it, which is deterministic.

packages/r2/src/multipart.ts
```diff
@@ -0,0 +1,568 @@
...
+  // to more closely match R2 behaviour.
+  const size = likelyOnFilesystem && process.platform === "win32" ? 32 : 128;
+  return crypto.randomBytes(size).toString("base64url");
+}
+function generateMultipartEtag(md5Hexes: string[]) {
+  // TODO: R2's multipart ETags don't seem to be deterministic, should ours be?
+  // https://stackoverflow.com/a/19896823
+  const hash = crypto.createHash("md5");
+  for (const md5Hex of md5Hexes) hash.update(Buffer.from(md5Hex, "hex"));
+  return `${hash.digest("hex")}-${md5Hexes.length}`;
```
Frederik-Baetens, line 120: Assuming that this is how S3 calculates their etags, we might move to this way of calculating them ourselves in the future.

Frederik-Baetens (approved):
impressive work :) I mostly took a look at the r2/src changes I hope that covers most of the user facing behavior.

PR #487, opened 2023-02-02
Throw an error for KV expirations above the maximum 32-bit signed integer

### LINKED ISSUE #485: KV expiration & TTLs can only be 32-bit signed integers in practice
Ran into this because I accidentally sent epoch milliseconds as `expiration` in KV. This also happens for `expirationTtl`.

### Reproduction
```ts
kvNamespace.put(“key”, “value”, { expiration: 1675058732043 })
```

### Expected behaviour
In the Workers environment, this throws a `TypeError`:

```
TypeError: Value out of range. Must be between -2147483648 and 2147483647 (inclusive).
```

### Actual behaviour
The put succeeds

huw:
Mostly self-explanatory, but a couple of thoughts:

1. I didn’t check whether `cacheTtl` on `get` is also subject to this limit.
2. I didn’t match the error to Cloudflare’s one, which is `TypeError: Value out of range. Must be between -2147483648 and 2147483647 (inclusive).`. My assumption is that Cloudflare are directly throwing a database error and this could be subject to change in the future. But if you want the errors to match let me know ^_^
3. I didn’t split out the message from the existing range ones, but can also do this if you’d like

You have more important things to do than edit this PR, so just let me know what changes you’d like and I will implement them ❤️

Resolves #485

### DIFF
--- a/packages/kv/src/namespace.ts
+++ b/packages/kv/src/namespace.ts
@@ -289,11 +289,11 @@ export class KVNamespace {
     let expiration = normaliseInt(options.expiration);
     const expirationTtl = normaliseInt(options.expirationTtl);
     if (expirationTtl !== undefined) {
-      if (isNaN(expirationTtl) || expirationTtl <= 0) {
+      if (isNaN(expirationTtl) || expirationTtl <= 0 || expirationTtl > 2147483647) {
         throwKVError(
           "PUT",
           400,
-          `Invalid expiration_ttl of ${options.expirationTtl}. Please specify integer greater than 0.`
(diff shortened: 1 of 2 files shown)


huw:
Sorry mate, I’ll throw on a changeset in a touch.

huw:
I triple-checked how Cloudflare was throwing and it was definitely just throwing raw `TypeErrors`, including for highly negative numbers. So I adjusted the order of operations a bit, too. Probably worth reporting internally because they don’t throw any sugar with it.

`kv

mrbbot (approved):
Awesome! Thank you! 😃

Issue #348, opened 2022-08-25
getMiniflareFetchMock global type not defined


lourd:
When trying to use the new fetch mocking capability, it doesn't appear that the `getMiniflareFetchMock` global function has a type definition that's set up for end users to consume. I would expect that adding 'jest-environment-miniflare' to the `compilerOptions.types` array of my tsconfig would add that function to the global types, but it does not. Which makes sense because looking at the `index.d.ts` in the published npm module, `getMiniflareFetchMock` is nowhere to be found.

mrbbot:
Hey! 👋 I agree these should be declared somewhere. They're actually here...
https://github.com/cloudflare/miniflare/blob/3e459b7899f3cacdd229408900a3937aeafd0cf9/packages/jest-environment-miniflare/src/index.ts#L33-L42
...but [API Extractor](https://api-extractor.com/), the tool we use for bundling type definitions, doesn't include them. 😢 

I would suggest you declare them yourself like this: https://github.com/mrbbot/miniflare-typescript-esbuild-jest/blob/master/test/globals.d.ts
...but I've just tried to upgrade that template and hit a bunch of errors due to conflicting `@cloudflare/workers-types` and `@types/node` (required by `undici`) types. 😅

Will have a think about this, but I suspect the quickest solution will be to manually define the `MockAgent` types yourself. 😢 

I'm also going to close this as a duplicate of #94.

aeneasr:
Could this be fixed? Looks like it's just some configuration you need to do in the API Extractor. Anyone using this library with TypeScript hits this sooner or later.

tizmagik:
I just hit this today. For now I just did this:

```typescript
import type { MockAgent } from "undici";
declare const getMiniflareFetchMock: () => MockAgent;
```

Which, yes, unfortunately means I had to add `undici` as a direct devDependency

lucaslago:
+1 a bit of a hassle to setup the types for jest environment ourselves

PR #494, opened 2023-02-07
Extract out test environment global types

### LINKED ISSUE #94: Helper types for Jest
`jest-environment-miniflare` provides two helper functions without type signatures. In order to avoid having to `//@ts-ignore` everywhere the helper functions are used could we get some ambient types such as the following added somehow?

```typescript
declare function getMiniflareBindings(): {
  [key: string]: DurableObjectNamespace
}
declare function getMiniflareDurableObjectStorage(
  id: DurableObjectId,
): DurableObjectStorage
```

mrbbot:
Miniflare's Jest and Vitest environments define a few global helper functions. Previously, the global type augmentations for these functions were included with the implementations. Unfortunately, API extractor (the tool we use to bundle `*.d.ts` files), doesn't support bundling global augmentations (microsoft/rushstack#1176).

This change extracts those augmentations into separate `globals.d.ts` files. This means users will be able to add
`jest-environment-miniflare/globals` or `vitest-environment-miniflare/globals` to their `types` configuration and get these types in their environment, e.g.

```json
{
  "compilerOptions": {
    "target": "esnext",
    "lib": ["esnext"],
    "module": "esnext",
    "moduleResolution": "node",
    "types": [
      "@cloudflare/workers-types",
      "vitest-environment-miniflare/globals"
    ]
  }
}
```

Closes #94
Ref #348

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1148,6 +1148,11 @@
         "mime": "^2.5.2"
       }
     },
+    "node_modules/@cloudflare/workers-types": {
+      "version": "4.20230115.0",
+      "resolved": "https://registry.npmjs.org/@cloudflare/workers-types/-/workers-types-4.20230115.0.tgz",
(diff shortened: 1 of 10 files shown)


penalosa (approved):
Lgtm, but needs a changeset

PR #495, opened 2023-02-07
Pass through WebSocket abnormal closure code

### LINKED ISSUE #465: [BUG] ws/lib/sender.js:162: 'First argument must be a valid error code number'
miniflare occasionally throws like this, see https://github.com/cloudflare/wrangler2/issues/2476 or https://github.com/cloudflare/wrangler2/issues/1749, 
or here #284 .

The problem comes here https://github.com/cloudflare/miniflare/blob/80046ce36c1637668d99d0679ca3f0a2a4beb0c7/packages/web-sockets/src/couple.ts#L57-L63 .

and for ws, it receives:
```js
    if (code === undefined) {
      buf = EMPTY_BUFFER;
    } else if (typeof code !== 'number' || !isValidStatusCode(code)) {
      throw new TypeError('First argument must be a valid error code number');
```
```js
function isValidStatusCode(code) {
  return (
    (code >= 1000 &&
      code <= 1014 &&
      code !== 1004 &&
      code !== 1005 &&
      code !== 1006) ||
    (code >= 3000 && code <= 4999)
  );
}
```

Maybe it just needs some tweaks to fit in.

mrbbot:
Previously, the abnormal closure close code `1006` was passed directly to `StandardWebSocket#close()`. This is forbidden, throwing an uncatchable validation error. With this change, the WebSocket is `terminate()`ed instead, propagating the `1006` code correctly to the other side of the connection.

Closes #465
Ref #493

### DIFF
--- a/packages/web-sockets/src/index.ts
+++ b/packages/web-sockets/src/index.ts
@@ -7,5 +7,6 @@ export {
   WebSocket,
   WebSocketPair,
   coupleWebSocket,
+  _kClose,
 } from "./websocket";
 export type { WebSocketEventMap } from "./websocket";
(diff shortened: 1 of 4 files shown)


penalosa (approved):
Why the rename to `_kClose`?

PR #497, opened 2023-02-09
Run API extractor on PRs and fix `@internal` WebSocket methods

mrbbot:
API extractor was failing as `[_kClose]()` wasn't marked as `@internal`, but `_kClose` was. We should be running `types:bundle` in PR CI to catch these types of issues.


penalosa (approved):
Is the jest check safe to remove?

PR #498, opened 2023-02-10
Add `node:{assert,buffer,events,util}` behind `nodejs_compat` flag

mrbbot:
This PR adds `node:assert`, `node:buffer`, `node:events` and `node:util` modules behind the `nodejs_compat` (and `experimental`) compatibility flags. We'll need to keep updating these as `workerd` implements more of the Node.js API. 🙂


package.json
```diff
@@ -57,6 +57,6 @@
     "node": ">=16.13"
   },
   "volta": {
-    "node": "19.3.0"
+    "node": "16.13.0"
```
mrbbot, line 60: We should be developing on our minimum supported version, bumping this previously was probably a bad idea...

packages/core/src/plugins/core.ts
```diff
@@ -426,7 +426,8 @@ export class CorePlugin extends Plugin<CoreOptions> implements CoreOptions {
 
     const nodejsCompat = ctx.compat.isEnabled("nodejs_compat");
     if (nodejsCompat) {
-      this.#additionalModules = additionalModules;
+      const experimental = ctx.compat.isEnabled("experimental");
```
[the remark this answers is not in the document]
mrbbot: It's both:https://github.com/cloudflare/workerd/blob/5cf69f7f05dd9f5a4e91ad4af30ba6d9503a1e7a/src/workerd/io/compatibility-date.capnp#L259. `"experimental"` is a compatibility flag annotated with `$experimental` requiring the `--experimental` flag to use. 😅 

The work-in-progress `nodejs_compat` stuff is gated behind the `"experimental"` flag: https://github.com/cloudflare/workerd/blob/552c616d6f7d84bbd4a4d652ac8d9aee17f4310d/src/workerd/api/node/node.h#L39.

packages/core/src/plugins/node/assert.ts
```diff
@@ -0,0 +1,23 @@
...
+export const match = assert.match;
+export const notDeepEqual = assert.notDeepEqual;
+export const notDeepStrictEqual = assert.notDeepStrictEqual;
+export const notEqual = assert.notEqual;
+export const notStrictEqual = assert.notStrictEqual;
+export const ok = assert.ok;
+export const rejects = assert.rejects;
+export const strict = assert.strict;
+export const strictEqual = assert.strictEqual;
+export const throws = assert.throws;
```
penalosa, line 23: Why the explicit exports instead of `export *`? Are there exports of `assert` that aren't implemented?

packages/core/src/plugins/node/util.ts
```diff
@@ -0,0 +1,9 @@
+import { types } from "node:util";
+
+// TODO(soon): update once https://github.com/cloudflare/workerd/pull/347 merged
```
penalosa, line 3: This has been merged now, will this need updated?

packages/core/src/plugins/node/assert.ts
mrbbot: Yep, `assert.CallTracker`. Also, if more exports are added in future Node versions, I'd ideally like this to stay the same across all of Miniflare's supported Node versions, and in sync with `workerd`.

packages/core/src/plugins/node/util.ts
mrbbot: Yep! Will do that now. 👍

penalosa (approved):

jasnell:
On this, currently, only the `async_hooks` (for `AsyncLocalStorage`) and `events` modules are exposed *only* using the `nodejs_compat` flag. The `buffer`, `assert`, and `util` are behind the additional `experimental` flag. Those will be moved out from behind the experimental flag once we have more tests in place for them in CI.

PR #500, opened 2023-02-10
Add support for D1 `meta.{last_row_id,changes}`

mrbbot:
This PR adds support for `meta.last_row_id` and `meta.changes`. These appear to default to 0 in case of read-only statements. See https://discord.com/channels/595317990191398933/992060581832032316/1073479458952065094.

/cc @geelen (can't add you as a reviewer 😢)

### DIFF
--- a/packages/d1/src/api.ts
+++ b/packages/d1/src/api.ts
@@ -30,32 +30,38 @@ interface ResponseMeta {
 interface SuccessResponse {
   results: any;
   duration: number;
-  lastRowId: number | null;
-  changes: number | null;
   success: true;
   served_by: string;
   meta: ResponseMeta | null;
+  // These are deprecated in place of `meta`
+  lastRowId: number | null;
+  changes: number | null;
 }
 
 const served_by = "miniflare.db";
 
-function ok(results: any, start: number): SuccessResponse {
-  const duration = performance.now() - start;
+interface OkMeta {
+  start: number;
+  last_row_id?: number;
+  changes?: number;
+}
+function ok(results: any, meta: OkMeta): SuccessResponse {
+  const duration = performance.now() - meta.start;
   return {
     results,
     duration,
-    // These are all `null`ed out in D1
-    lastRowId: null,
-    changes: null,
     success: true,
     served_by,
     meta: {
       duration,
-      last_row_id: null,
-      changes: null,
+      last_row_id: meta.last_row_id ?? 0,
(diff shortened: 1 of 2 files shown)


packages/d1/src/api.ts
```diff
@@ -30,32 +30,38 @@ interface ResponseMeta {
...
   results: any;
   duration: number;
-  lastRowId: number | null;
-  changes: number | null;
   success: true;
   served_by: string;
   meta: ResponseMeta | null;
+  // These are deprecated in place of `meta`
+  lastRowId: number | null;
+  changes: number | null;
```
geelen, line 38: I think we can remove these now. They've never had data in them, and now they never will (only inside `.meta`)

packages/d1/src/api.ts
```diff
@@ -30,32 +30,38 @@ interface ResponseMeta {
...
-      last_row_id: null,
-      changes: null,
+      last_row_id: meta.last_row_id ?? 0,
+      changes: meta.changes ?? 0,
       served_by,
       internal_stats: null,
     },
+    // These are deprecated in place of `meta`
+    lastRowId: null,
+    changes: null,
```
geelen, line 64: Here too

packages/d1/src/api.ts
```diff
@@ -118,22 +124,26 @@ export class D1DatabaseAPI {
...
     const sql = splitSqlQuery(query.sql)[0];
     const stmt = this.db.prepare(sql);
     // `/execute` only supports queries that don't return data
     if (returnsData(stmt)) throw new Error(EXECUTE_RETURNS_DATA_MESSAGE);
     const params = normaliseParams(query.params);
-    stmt.run(params);
-    return ok(null, start);
+    const result = stmt.run(params);
+    meta.last_row_id = Number(result.lastInsertRowid);
+    meta.changes = result.changes;
```
geelen, line 145: Eyeballing this, it looks right to me.

geelen (approved):

packages/d1/src/api.ts
rozenmd: Yeah if we were keeping these it'd be more correct to hardcode the type to `null` tho

mrbbot: Cool, I'll replace the type with `null` since they are still included in internal API responses as far as I can tell.

penalosa (approved):
Assuming @rozenmd and/or @geelen are on board, LGTM!

PR #508, opened 2023-02-19
chore: bump undici and other packages to resolve audit warnings

Cherry:
This PR bumps undici to 5.20.0 which resolves https://github.com/advisories/GHSA-5r9g-qh6m-jxff and https://github.com/advisories/GHSA-r6ch-mqf9-qc9w. This removes the scary `npm audit` warnings when installing wrangler.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -113,9 +113,10 @@
       }
     },
     "node_modules/@babel/core/node_modules/json5": {
-      "version": "2.2.1",
+      "version": "2.2.3",
+      "resolved": "https://registry.npmjs.org/json5/-/json5-2.2.3.tgz",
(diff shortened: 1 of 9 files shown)


Cherry:
Looks like a few tests/typechecks are failing: https://github.com/cloudflare/miniflare/actions/runs/4218389719/jobs/7322866480

But testing this on Windows is proving difficult:
```
No files matching the pattern "'packages/*/{src,test}/**/*.ts'" were found.
```
And then when I patch that up by using `"` instead of `'`, I get:

```
✖ 42390 problems (42390 errors, 0 warnings)
  42390 errors and 0 warnings potentially fixable with the `--fix` option.
```

So I think I'll just leave this one to someone else. 😅

PR #509, opened 2023-02-19
chore: fix linting on windows

Cherry:
eslint and Windows don't play nicely with `'` for the file globs.

This also fixes line endings being auto-determined and instead forced to LF on Windows. Previously without this, and after fixing the linting script, you'd be met with the following unless you were forcing LF line endings like in a git config:

```
✖ 42390 problems (42390 errors, 0 warnings)
  42390 errors and 0 warnings potentially fixable with the `--fix` option.
```
as prettier saw the CRLF line endings as invalid for every file. With `.gitattributes`, these are now normalised so Windows users can run the command and test more efficiently.


mrbbot (approved):
Thanks for the PR! 😃

PR #510, opened 2023-02-19
Bump undici

WalshyDev:
This PR bumps undici to 5.20.0 which resolves https://github.com/advisories/GHSA-5r9g-qh6m-jxff and https://github.com/advisories/GHSA-r6ch-mqf9-qc9w. This removes the scary npm audit warnings when installing wrangler.

Continuing #508 

Few things to note:
- `Request#duplex` now required if `body` is provided in `RequestInit`
  - https://github.com/nodejs/node/issues/46221
  - https://github.com/whatwg/fetch/pull/1457
  - Workers doesn't support this property yet so just passing it down to undici and avoiding Workers
- Was getting segfaults on Node 16.16.0
  - https://github.com/nodejs/node/issues/43617
  - Bumped up Node to 16.17 resolves the issue

Still working through the few tests still failing... not sure why they're failing yet though.

>   Caused by: Error: To use the new ReadableStream() constructor, enable the streams_enable_constructors feature flag.

For some reason it's detecting ReadableStream in use even for a very basic test:
```
test("fetch mock", async () => {
  const fetchMock = getMiniflareFetchMock();

  fetchMock.disableNetConnect();
  fetchMock
    .get("https://example.com")
    .intercept({ path: "/" })
    .reply(200, "body");

  const response = await fetch("https://example.com");
  expect(await response.text()).toBe("body");
});
```


types/global.d.ts
```diff
@@ -2,6 +2,8 @@ declare global {
   // TODO(soon): remove once included in `@types/node`
   // eslint-disable-next-line no-var
   var CryptoKey: typeof import("crypto").webcrypto.CryptoKey;
+  // eslint-disable-next-line no-var
+  var MessagePort: typeof import("worker_threads").MessagePort;
```
mrbbot, line 6: This was required for `undici` type checking to pass.

packages/vitest-environment-miniflare/src/index.ts
```diff
@@ -56,9 +59,28 @@ function setupIsolatedStorage(storageFactory: StackedMemoryStorageFactory) {
...
   async setup(global, options) {
+    // Since `undici@5.14.0`, stream classes are loaded from the global scope
+    // if available (https://github.com/nodejs/undici/pull/1793). Make sure
+    // `undici` sets module variables for stream classes before we assign
+    // Miniflare's versions. Accessing these would throw if the
+    // `streams_enable_constructors` compatibility flag wasn't enabled.
+    globalThis.ReadableStream = stream.ReadableStream;
+    globalThis.WritableStream = stream.WritableStream;
+    globalThis.TransformStream = stream.TransformStream;
+    require("undici/lib/fetch");
```
mrbbot, line 82: Just importing `undici` doesn't work here, as the `fetch` implementation is lazily imported

packages/core/src/standards/http.ts
```diff
@@ -370,25 +374,54 @@ export function _isBodyStream(stream: ReadableStream): boolean {
...
+  return result;
+}
+
 export class Request extends Body<BaseRequest> {
   // noinspection TypeScriptFieldCanBeMadeReadonly
   #cf?: IncomingRequestCfProperties | RequestInitCfProperties;
 
   constructor(input: RequestInfo, init?: RequestInit) {
+    // If body is a FixedLengthStream, set Content-Length to its expected length
+    const contentLength: number | undefined = (init?.body as any)?.[
```
mrbbot, line 421: We may replace `init.body` later on with a different `ReadableStream`, so extract `contentLength` first

penalosa, line 421: Can this be typed as something other than `any`? (I suspect not, given it seems like `kContentLength` is internal, but worth checking)

packages/core/src/standards/http.ts
```diff
@@ -397,19 +430,30 @@ export class Request extends Body<BaseRequest> {
       // Don't pass our strange hybrid Request to undici
       if (input instanceof Request) input = input[_kInner];
       if (init instanceof Request) init = init[_kInner] as RequestInit;
+      // We may mutate some `init` properties below, make sure these aren't
+      // visible to the caller
+      init = init && cloneRequestInit(init);
```
penalosa, line 435: I feel like a ternary here might be clearer?

packages/core/src/standards/streams.ts
```diff
@@ -80,6 +80,27 @@ function convertToByteStream(
...
+      let result = await reader.read();
+      while (!result.done && result.value.byteLength === 0) {
+        result = await reader.read();
+      }
+      if (result.done) {
+        queueMicrotask(() => controller.close());
+      } else if (result.value.byteLength > 0) {
+        controller.enqueue(result.value);
+      }
+    },
```
penalosa, line 99: Should this be reading the whole stream in one invocation of `pull`?

penalosa (approved):

packages/core/src/standards/streams.ts
mrbbot: It will only read the whole stream if all chunks are empty, which seems unlikely. The `while` loop is just there to ensure we get a non-empty result to enqueue.

PR #519, opened 2023-02-22
Miniflare 3: Add support for D1

### LINKED ISSUE #479: 🐛 BUG: --experimental-local is not supported with d1
### What version of `Wrangler` are you using?

2.4.4

### What operating system are you using?

mac

### Describe the Bug

Execution of ```wrangler dev --experimental-local --persist``` leads to the next error:
service core:user:rub: Uncaught TypeError: Cannot read properties of undefined (reading 'middleware')
  at d1-beta-facade.entry.js:65:31
✘ [ERROR] local worker: MiniflareCoreError [ERR_RUNTIME_FAILURE]: The Workers runtime failed to start. There is likely additional logging output above.

Execution of ```wrangler dev --local --persist``` works fine

mrbbot:
This PR ports Miniflare 2's D1 implementation to Miniflare 3. We'd already laid lots of the groundwork for this by reworking Miniflare 2 to implement D1's underlying API instead of the Worker binding API.

Miniflare 3 will add a regular `Fetcher` service binding for each database specified in `d1Databases`. Note this is different to Miniflare 2 which provided actual `D1Database` bindings. If we wanted to replicate this behaviour, we'd need to transform code similar to what Wrangler does at the moment.

This needs a small change to the D1 facade application code, as `--experimental-local` now needs to be treated like remote mode. The facade itself also needs to be updated to pass full URLs to the underlying service bindings, as `workerd` does not provide host-less `Fetcher`s.

This PR also adds `Error#cause` support when passing `JsonError`s with the `MF-Experimental-Error-Stack` header, and propagates these errors through `dispatchFetch()` for easier testing.

Closes #479
Closes cloudflare/workers-sdk#2667

/cc @geelen @rozenmd


rozenmd:
Is there a logical order to follow, to make it easier to review?

packages/tre/src/plugins/core/errors/index.ts
```diff
@@ -196,36 +196,73 @@ function getSourceMappedStack(workerSrcOpts: SourceOptions[], error: Error) {
...
+    name: z.ostring(),
+    stack: z.ostring(),
+    cause: JsonErrorSchema.optional(),
+  })
+);
+
+export function reviveError(
+  workerSrcOpts: SourceOptions[],
+  jsonError: JsonError
+): Error {
```
penalosa, line 218: I'm not sure I really understand this function—what's going on here?

packages/tre/src/plugins/d1/gateway.ts
```diff
@@ -0,0 +1,226 @@
...
+function err(error: unknown): D1ErrorResponse {
+  return {
+    error: String(error),
+    success: false,
+    served_by,
+  };
+}
+
+export class D1Error extends HttpError {
+  constructor(cause: unknown) {
```
penalosa, line 85:
```suggestion
  constructor(cause: Error) {
```

packages/tre/src/plugins/d1/gateway.ts
```diff
@@ -0,0 +1,226 @@
...
+  return rows.map((row) =>
+    Object.fromEntries(
+      Object.entries(row).map(([key, value]) => [
+        key,
+        // If `value` is an array, convert it to a regular numeric array
+        value instanceof Buffer ? Array.from(value) : value,
+      ])
+    )
+  );
+}
```
penalosa, line 112: Is there any way to type this better than `any`?

penalosa (commented):
I'm slightly concerned by the storage interfaces—it's getting a bit confusing to have sqlite databases in each storage provider, as well as a file system temporary path for 'in-memory' storage. I know miniflare 2 supported multiple storage providers, but maybe there's an argument for miniflare 3 to just support store things as best it can, and choose the best method based on the data (i.e. sqlite for durable objects and D1, file system for R2, sqlite for KV etc...)

packages/tre/src/shared/error.ts
```diff
@@ -28,4 +30,12 @@ export class HttpError extends MiniflareError<number> {
...
     super(code, message, cause);
   }
+
+  toResponse(): Response {
+    return new Response(this.message, {
+      status: this.code,
+      // Custom statusMessage is required for runtime error messages
+      statusText: this.message.substring(0, 512),
+    });
+  }
```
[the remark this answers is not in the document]
mrbbot: Yeah, this makes wayyyy more sense 😅

packages/tre/src/plugins/core/errors/index.ts
mrbbot: At a high level, this function takes a JSON-serialisable representation of an `Error`, and converts it to an `Error`. 

`Error`s may have `cause`s, so it needs to do this recursively.

If we threw a `TypeError` in the Worker, we'd like to construct a `TypeError` here, so it looks like the error has been thrown through a regular function call, not a HTTP request (i.e. we want `instanceof TypeError` to pass in Node for `TypeError`s thrown in Workers).

Because constructing an `Error` captures the stack trace at point of construction, we override the stack trace to the one from the Worker in the JSON-serialised error. Finally, we try to apply source-mapping given an array of Worker sources.

packages/tre/src/plugins/d1/gateway.ts
mrbbot: Yeah, we could probably enumerate the valid D1 param/return types. 👍

mrbbot (commented):
> I'm slightly concerned by the storage interfaces

Me too! 😅 

I think this storage interface worked well in the early days of Miniflare, but as more complex data storage solutions have been added to Workers, it's started to show some rough edges. In particular, it doesn't handle transactions natively, nor does it provide streaming APIs.

After I've ported the R2 tests, I'm going to try implement a new storage system. My plan is for it to have two parts: an SQLite-backed metadata store, and a range-addressable, streaming blob store. We'll provide in-memory and file-system backed implementations of each.

Then, we'll need to update our gateways to use the new system, something like:
- **KV:** metadata in SQLite (for easy listing), values in blob-store _(should be possible to use blob-store for Workers Sites too)_
- **Cache:** metadata in SQLite, values in blob-store
- **R2:** metadata in SQLite (for easy listing, transactional updates especially for multipart uploads), values in blob-store
- **D1:** everything in SQLite
- **Durable Objects:** neither, using `workerd`'s built-in SQLite backed storage

packages/tre/src/plugins/core/errors/index.ts
```diff
@@ -196,36 +196,73 @@ function getSourceMappedStack(workerSrcOpts: SourceOptions[], error: Error) {
...
+  let cause: Error | undefined;
+  if (jsonError.cause !== undefined) {
+    cause = reviveError(workerSrcOpts, jsonError.cause);
+  }
+
+  // If this is one of hte built-in error types, construct an instance of that
+  let ctor = Error;
+  if (jsonError.name !== undefined && jsonError.name in globalThis) {
+    const maybeCtor = (globalThis as Record<string, unknown>)[jsonError.name];
```
penalosa, line 227: This seems like a bit of a security issue (although it's all running locally with presumably trusted code, so...). What happens if someone sets the `jsonError.name` to `eval`, and `jsonMessage` to some executable code? I think `maybeCtor` should probably be allow-listed to just `Error` and `TypeError`.

packages/tre/src/plugins/core/errors/index.ts
```diff
@@ -196,36 +196,73 @@ function getSourceMappedStack(workerSrcOpts: SourceOptions[], error: Error) {
...
+    if (Object.getPrototypeOf(maybeCtor) === Error) {
+      ctor = maybeCtor as typeof Error;
+    }
+  }
+
+  // Construct the error, copying over the correct name and stack trace
+  const error = new ctor(jsonError.message, { cause });
+  if (jsonError.name !== undefined) error.name = jsonError.name;
```
penalosa, line 235: Why is this line needed? If `jsonError.name !== undefined` then presumably the error was constructed with the right class to get the correct name? (Also see above re allow listing)

packages/tre/src/plugins/core/errors/index.ts
```diff
@@ -196,36 +196,73 @@ function getSourceMappedStack(workerSrcOpts: SourceOptions[], error: Error) {
...
-
-  // Try to apply source-mapping
-  error.stack = getSourceMappedStack(workerSrcOpts, error);
+  const error = reviveError(workerSrcOpts, caught);
 
   // Log source-mapped error to console if logging enabled
   log.error(error);
 
   // Lazily import `youch` when required
   const Youch: typeof import("youch").default = require("youch");
```
penalosa, line 262: I assume there's some reason `Youch` doesn't like `await import("youch")`?

packages/tre/src/plugins/core/errors/index.ts
```diff
@@ -196,36 +196,73 @@ function getSourceMappedStack(workerSrcOpts: SourceOptions[], error: Error) {
...
-  const youch = new Youch(error, {
+  // `cause` is usually more useful than the error itself, display that instead
+  // TODO(someday): would be nice if we could display both
+  const youch = new Youch(error.cause ?? error, {
```
penalosa, line 265: Why will cause be more useful? Regardless, this should probably be consistent with the `log.error()` above

packages/tre/src/plugins/d1/gateway.ts
```diff
@@ -0,0 +1,226 @@
...
+import { z } from "zod";
+import { Response } from "../../http";
+import { HttpError, Log } from "../../shared";
+import { Storage } from "../../storage";
+import splitSqlQuery from "./splitter";
+
+// query
+export const D1SingleQuerySchema = z.object({
+  sql: z.string(),
+  params: z.array(z.any()).nullable().optional(),
```
penalosa, line 16: `z.any()`: Does D1 have a limited set of param types that could be used here?

packages/tre/src/plugins/d1/gateway.ts
```diff
@@ -0,0 +1,226 @@
...
+}
+export interface D1ResponseMeta {
+  duration: number;
+  last_row_id: number | null;
+  changes: number | null;
+  served_by: string;
+  internal_stats: null;
+}
+export interface D1SuccessResponse {
+  results: any;
```
penalosa, line 40: Can this be better typed?

packages/tre/src/plugins/d1/gateway.ts
```diff
@@ -0,0 +1,226 @@
...
+  served_by: string;
+  meta: D1ResponseMeta | null;
+  // These are deprecated in place of `meta`
+  lastRowId: null;
+  changes: null;
+}
+
+const served_by = "miniflare.db";
+
+interface OkMeta {
```
penalosa, line 52: How does this differ from `D1ResponseMeta`?

packages/tre/src/plugins/d1/gateway.ts
```diff
@@ -0,0 +1,226 @@
...
+    this.db = storage.getSqliteDatabase();
+  }
+
+  #query: QueryRunner = (query) => {
+    const meta: OkMeta = { start: performance.now() };
+    // D1 only respects the first statement
+    const sql = splitSqlQuery(query.sql)[0];
+    const stmt = this.db.prepare(sql);
+    const params = normaliseParams(query.params);
+    let results: any[];
```
penalosa, line 144:
```suggestion
    let results: unknown[];
```
maybe?

packages/tre/src/plugins/d1/gateway.ts
```diff
@@ -0,0 +1,226 @@
...
+    if (returnsData(stmt)) {
+      results = stmt.all(params);
+    } else {
+      // `/query` does support queries that don't return data,
+      // returning `[]` instead of `null`
+      const result = stmt.run(params);
+      results = [];
+      meta.last_row_id = Number(result.lastInsertRowid);
+      meta.changes = result.changes;
```
penalosa, line 153: Should these not be set on the previous `if` branch as well? Statements that return data can also change data, right?

packages/tre/src/plugins/d1/gateway.ts
```diff
@@ -0,0 +1,226 @@
...
+    // D1 only respects the first statement
+    const sql = splitSqlQuery(query.sql)[0];
+    const stmt = this.db.prepare(sql);
+    // `/execute` only supports queries that don't return data
+    if (returnsData(stmt)) throw new Error(EXECUTE_RETURNS_DATA_MESSAGE);
+    const params = normaliseParams(query.params);
+    const result = stmt.run(params);
+    meta.last_row_id = Number(result.lastInsertRowid);
+    meta.changes = result.changes;
+    return ok(null, meta);
```
penalosa, line 169: Wait, so `execute` returns `null`, while `query` returns an empty array for statements that don't return data? That seems like a confusing API design to me (cc @rozenmd—what's the reasoning for this?)

packages/tre/src/storage/file.ts
```diff
@@ -317,4 +319,10 @@ export class FileStorage extends LocalStorage {
     }
     return keys;
   }
+
+  getSqliteDatabase(): DatabaseType {
+    if (this.#sqliteDatabase !== undefined) return this.#sqliteDatabase;
+    mkdirSync(path.dirname(this.root), { recursive: true });
+    return (this.#sqliteDatabase = new Database(this.root + ".sqlite3"));
```
penalosa, line 326: I could be completely, but this seems like it might put the file somewhere weird?

If I set `this.root` to `~/.mf`, will this not put the database at `~/.mf.sqlite3`? I think I'd expect it to be put inside the root folder

packages/tre/src/storage/sqlite.ts
```diff
@@ -205,4 +205,8 @@ export class SqliteStorage extends LocalStorage {
       sql`SELECT key, attributes FROM ${unsafe_raw(this.namespace)}`
     );
   }
+
+  getSqliteDatabase(): DatabaseType {
+    return this.database;
```
penalosa, line 210: I almost wonder if this should be a different database. You probably shouldn't be able to inspect your KV data through D1, for instance. Although, if storage is changing soon, that's not a big deal or a blocker for this PR

packages/tre/src/plugins/core/errors/index.ts
```diff
@@ -196,36 +196,73 @@ function getSourceMappedStack(workerSrcOpts: SourceOptions[], error: Error) {
...
+  workerSrcOpts: SourceOptions[],
+  jsonError: JsonError
+): Error {
+  let cause: Error | undefined;
+  if (jsonError.cause !== undefined) {
+    cause = reviveError(workerSrcOpts, jsonError.cause);
+  }
+
+  // If this is one of hte built-in error types, construct an instance of that
+  let ctor = Error;
```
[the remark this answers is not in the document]
mrbbot: I think `ctor` is generally the convention for this, e.g. in Node: https://github.com/nodejs/node/blob/b3663e0262f68740078b90c55b1ffc0f8ac1ac76/lib/internal/webstreams/readablestream.js#L2510.

packages/tre/src/plugins/core/errors/index.ts
mrbbot: I think the `Object.getPrototypeOf(maybeCtor)` check should prevent this. There are also `EvalError`, `InternalError`, `RangeError`, `ReferenceError`, `SyntaxError`, `TypeError` and `URIError`. Could still do an allow-list though?

packages/tre/src/plugins/core/errors/index.ts
mrbbot: Not necessarily, what if the user defined their own subclass of `Error`? We won't be able to construct that in Node (we'll fallback to `Error`), but we'd still like the name to reflect the custom class.

packages/tre/src/plugins/core/errors/index.ts
mrbbot: Hmmm, apparently the `es/no-dynamic-import` lint rule is enabled. I think this way breaking things in the past with Jest and Miniflare 2. FWIW, Miniflare 3 is still compiled to CommonJS code, so this `require` is probably better.

packages/tre/src/plugins/core/errors/index.ts
mrbbot: For D1, errors are wrapped in a new `Error` that just contains `D1_ERROR`. The `cause` has the actually useful SQLite errors. `log.error()` will log both. 🙂

packages/tre/src/plugins/d1/gateway.ts
mrbbot: Yeah, we should be able to enumerate them as above.

packages/tre/src/plugins/d1/gateway.ts
mrbbot: Yeah, we should be able to enumerate them as above.

packages/tre/src/plugins/d1/gateway.ts
mrbbot: It contains a `start` time not a `duration`, `last_row_id` and `changes` may be `undefined` instead of `null`.

packages/tre/src/plugins/d1/gateway.ts
mrbbot: I _think_ I checked this, and they didn't return in this case, but will double check. 👍

packages/tre/src/plugins/core/errors/index.ts
rozenmd: The temptation to get rid of D1_ERROR is real 😭

packages/tre/src/storage/file.ts
mrbbot: Hmmm, this is what we did in Miniflare 2, but I guess it's kinda weird. Something we'll probably fix with the new storage system, since that's going to be a massive breaking change anyways.

packages/tre/src/storage/sqlite.ts
mrbbot: Yeah, I guess KV should never be calling this method. D1 won't use the other methods either.

packages/tre/src/plugins/d1/gateway.ts
rozenmd: @geelen / @xtuc  would know the exact details, but we seem to only use `query`?

packages/tre/src/plugins/d1/gateway.ts
mrbbot: Ah, actually SQLite exposes SQL functions for these! 🎉 
- https://www.sqlite.org/lang_corefunc.html#changes
- https://www.sqlite.org/lang_corefunc.html#last_insert_rowid

packages/tre/src/plugins/core/errors/index.ts
mrbbot: bac9101

packages/tre/src/plugins/d1/gateway.ts
mrbbot: 46bc3d6

packages/tre/src/plugins/d1/gateway.ts
mrbbot: 46bc3d6

packages/tre/src/plugins/core/errors/index.ts
```diff
@@ -196,36 +196,73 @@ function getSourceMappedStack(workerSrcOpts: SourceOptions[], error: Error) {
...
+  message?: string;
+  name?: string;
+  stack?: string;
+  cause?: JsonError;
+}
+export const JsonErrorSchema: z.ZodType<JsonError> = z.lazy(() =>
+  z.object({
+    message: z.ostring(),
+    name: z.ostring(),
+    stack: z.ostring(),
```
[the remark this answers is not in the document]
mrbbot: bac9101

packages/tre/src/plugins/core/errors/index.ts
mrbbot: bac9101

packages/tre/src/plugins/d1/gateway.ts
mrbbot: 46bc3d6

packages/tre/src/plugins/d1/gateway.ts
mrbbot: 46bc3d6

packages/tre/src/plugins/d1/gateway.ts
mrbbot: 46bc3d6, changed this to what `better-sqlite3` returns

packages/tre/src/plugins/d1/gateway.ts
mrbbot: 46bc3d6

penalosa (approved):
Looks good! Thanks for addressing my comments! Re the new storage system, is there an issue open for that somewhere?

PR #520, opened 2023-02-23
[Miniflare 3] Add support for routing to multiple Workers

mrbbot:
This change ports Miniflare 2's `mounts` router to Miniflare 3. This attempts to replicate the logic of the `route(s)` field in the Wrangler configuration file: https://developers.cloudflare.com/workers/platform/triggers/routes/#matching-behavior

Internally, this is implemented by binding all routable workers as service bindings in the entry service. The first worker is always bound as a "fallback", in case no routes match. Validation has been added to ensure we a) have a fallback, b) don't have workers with duplicate names that would cause bindings with the same name, and c) all routable/fallback workers have code so they actually get added as `workerd` services.

I've also split some of the core plugin services into "global services". These need to be constructed from a different set of options to the regular plugin services, and relying on the de-duping behaviour here was a bit silly.

### DIFF
--- a/packages/tre/src/index.ts
+++ b/packages/tre/src/index.ts
@@ -168,7 +168,9 @@ function getWorkerRoutes(
   const allRoutes = new Map<string, string[]>();
   for (const workerOpts of allWorkerOpts) {
     if (workerOpts.core.routes !== undefined) {
-      allRoutes.set(workerOpts.core.name ?? "", workerOpts.core.routes);
+      const name = workerOpts.core.name ?? "";
+      assert(!allRoutes.has(name));
+      allRoutes.set(name, workerOpts.core.routes);
     }
   }
   return allRoutes;
--- a/packages/tre/src/index.ts
+++ b/packages/tre/src/index.ts
@@ -124,7 +124,9 @@ function validateOptions(
     if (names.has(name)) {
       throw new MiniflareCoreError(
         "ERR_DUPLICATE_NAME",
-        `Multiple workers defined with the same name: "${name}"`
+        name === ""
+          ? "Multiple workers defined without a `name`"
+          : `Multiple workers defined with the same \`name\`: "${name}"`
       );
     }
     names.add(name);
--- a/packages/tre/src/index.ts
+++ b/packages/tre/src/index.ts
@@ -35,6 +35,7 @@ import {
   Plugins,
   SERVICE_ENTRY,
   SOCKET_ENTRY,
+  getGlobalServices,
   maybeGetSitesManifestModule,
   normaliseDurableObject,
 } from "./plugins";
@@ -100,6 +101,9 @@ function validateOptions(
   const sharedOpts = opts;
   const multipleWorkers = "workers" in opts;
   const workerOpts = multipleWorkers ? opts.workers : [opts];
+  if (workerOpts.length === 0) {
+    throw new MiniflareCoreError("ERR_NO_WORKERS", "No workers defined");
+  }
 
   // Initialise return values
   const pluginSharedOpts = {} as PluginSharedOptions;
@@ -117,6 +121,19 @@ function validateOptions(
       // @ts-expect-error pluginWorkerOpts[i][key] could be any plugin's
       pluginWorkerOpts[i][key] = plugin.options.parse(workerOpts[i], { path });
     }
+  }
+
+  // Validate names unique
+  const names = new Set<string>();
+  for (const opts of pluginWorkerOpts) {
+    const name = opts.core.name ?? "";
+    if (names.has(name)) {
+      throw new MiniflareCoreError(
+        "ERR_DUPLICATE_NAME",
+        `Multiple workers defined with the same name: "${name}"`
+      );
+    }
+    names.add(name);
   }
 
   return [pluginSharedOpts, pluginWorkerOpts];
@@ -148,6 +165,19 @@ function getDurableObjectClassNames(
     }
   }
   return serviceClassNames;
+}
+
+// Collects all routes from all worker services
+function getWorkerRoutes(
+  allWorkerOpts: PluginWorkerOptions[]
+): Map<string, string[]> {
+  const allRoutes = new Map<string, string[]>();
+  for (const workerOpts of allWorkerOpts) {
+    if (workerOpts.core.routes !== undefined) {
+      allRoutes.set(workerOpts.core.name ?? "", workerOpts.core.routes);
+    }
+  }
+  return allRoutes;
 }
 
 // ===== `Miniflare` Internal Storage & Routing =====
@@ -622,7 +652,16 @@ export class Miniflare {
 
     sharedOpts.core.cf = await setupCf(this.#log, sharedOpts.core.cf);
 
-    const services: Service[] = [];
+    const durableObjectClassNames = getDurableObjectClassNames(allWorkerOpts);
+    const allWorkerRoutes = getWorkerRoutes(allWorkerOpts);
+
+    const services: Service[] = getGlobalServices({
+      optionsVersion,
+      sharedOptions: sharedOpts.core,
+      allWorkerRoutes,
+      fallbackWorkerName: this.#workerOpts[0].core.name,
+      loopbackPort,
+    });
     const sockets: Socket[] = [
       {
         name: SOCKET_ENTRY,
@@ -633,10 +672,13 @@ export class Miniflare {
       },
     ];
 
-    const durableObjectClassNames = getDurableObjectClassNames(allWorkerOpts);
-
     // Dedupe services by name
     const serviceNames = new Set<string>();
+    for (const service of services) {
+      // Global services should all have unique names
+      assert(service.name !== undefined && !serviceNames.has(service.name));
+      serviceNames.add(service.name);
+    }
 
     for (let i = 0; i < allWorkerOpts.length; i++) {
       const workerOpts = allWorkerOpts[i];
@@ -662,13 +704,11 @@ export class Miniflare {
         const pluginServices = await plugin.getServices({
           log: this.#log,
           options: workerOpts[key],
-          optionsVersion,
           sharedOptions: sharedOpts[key],
           workerBindings,
           workerIndex: i,
           durableObjectClassNames,
           additionalModules,
-          loopbackPort,
           tmpPath: this.#tmpPath,
         });
         if (pluginServices !== undefined) {
--- a/packages/tre/src/index.ts
+++ b/packages/tre/src/index.ts
@@ -35,6 +35,8 @@ import {
   Plugins,
   SERVICE_ENTRY,
   SOCKET_ENTRY,
+  SharedOptions,
+  WorkerOptions,
   getGlobalServices,
   maybeGetSitesManifestModule,
   normaliseDurableObject,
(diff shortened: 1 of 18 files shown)


packages/tre/src/plugins/cache/index.ts
```diff
@@ -87,7 +82,7 @@ export const CACHE_PLUGIN: Plugin<
               name: BINDING_JSON_CACHE_WARN_USAGE,
               json: JSON.stringify(options.cacheWarnUsage ?? false),
             },
-            loopbackBinding,
+            WORKER_BINDING_SERVICE_LOOPBACK,
           ],
           compatibilityDate: "2022-09-01",
```
penalosa, line 87: In the spirit of extracting shared options, the compatibility date could probably also be extracted

packages/tre/src/plugins/core/index.ts
```diff
@@ -14,9 +14,14 @@ import { getCacheServiceName } from "../cache";
 import { DURABLE_OBJECTS_STORAGE_SERVICE_NAME } from "../do";
 import {
   BINDING_SERVICE_LOOPBACK,
+  CORE_PLUGIN_NAME,
```
penalosa, line 17: Why the move of `CORE_PLUGIN_NAME` out of `core`?

packages/tre/src/plugins/core/index.ts
```diff
@@ -129,7 +134,10 @@ const LIVE_RELOAD_SCRIPT_TEMPLATE = (
 
 // Using `>=` for version check to handle multiple `setOptions` calls before
 // reload complete.
-export const SCRIPT_ENTRY = `async function handleEvent(event) {
+export const SCRIPT_ENTRY = `
+const matchRoutes = ${matchRoutes.toString()};
```
penalosa, line 138: Is this including a stringified function?

packages/tre/src/plugins/core/index.ts
```diff
@@ -129,7 +134,10 @@ const LIVE_RELOAD_SCRIPT_TEMPLATE = (
 
 // Using `>=` for version check to handle multiple `setOptions` calls before
 // reload complete.
-export const SCRIPT_ENTRY = `async function handleEvent(event) {
+export const SCRIPT_ENTRY = `
```
penalosa, line 137: These scripts are getting a lot more complex. Have you considered different ways of writing them so they're not restricted to string templates? (`fs.readFile` etc...)

packages/tre/src/plugins/core/index.ts
```diff
@@ -412,10 +377,13 @@ export const CORE_PLUGIN: Plugin<
...
-        name: BINDING_SERVICE_USER,
-        service: { name },
-      });
+    } else if (workerIndex === 0 || options.routes?.length) {
+      throw new MiniflareCoreError(
+        "ERR_ROUTABLE_NO_SCRIPT",
+        `Worker [${workerIndex}] ${
+          options.name === undefined ? "" : `("${options.name}") `
+        }must have code defined as it's routable or the fallback`
+      );
```
penalosa, line 386: Why can any workers have no code? What does it mean for a worker to not have code defined?

packages/tre/src/plugins/core/index.ts
```diff
@@ -451,6 +419,74 @@ export const CORE_PLUGIN: Plugin<
...
+  fallbackWorkerName,
+  loopbackPort,
+}: GlobalServicesOptions): Service[] {
+  // Collect list of workers we could route to, then parse and sort all routes
+  const routableWorkers = [...allWorkerRoutes.keys()];
+  const routes = parseRoutes(allWorkerRoutes);
+
+  // Define core/shared services.
+  const serviceEntryBindings: Worker_Binding[] = [
+    WORKER_BINDING_SERVICE_LOOPBACK, // For converting stack-traces to pretty-error pages
```
penalosa, line 442: It looks like every plugin also defines this service?

packages/tre/src/plugins/shared/routing.ts
```diff
@@ -0,0 +1,142 @@
...
+  allowHostnamePrefix: boolean;
+  hostname: string;
+  path: string;
+  allowPathSuffix: boolean;
+}
+
+const A_MORE_SPECIFIC = -1;
+const B_MORE_SPECIFIC = 1;
+
+export function parseRoutes(allRoutes: Map<string, string[]>): WorkerRoute[] {
```
penalosa, line 22: I imagine there's code to do this in EWC somewhere—have you checked the logic against that? (I'm assuming this is copied from Miniflare 2)

packages/tre/src/index.ts
```diff
@@ -100,6 +101,9 @@ function validateOptions(
   const sharedOpts = opts;
   const multipleWorkers = "workers" in opts;
   const workerOpts = multipleWorkers ? opts.workers : [opts];
```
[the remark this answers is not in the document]
mrbbot: Ooo interesting, I like that improved `MiniflareOptions` type. 

> have you considered making the `Miniflare` constructor take two parameters?

I haven't but I think I'd still rather everything was in one object. Most users will just have a single Worker, and I don't want them to think about which object this belongs too. For advanced multi-worker use cases, I think splitting into the `workers` array is ok, and closer to Miniflare 2's `mounts` API.

packages/tre/src/index.ts
```diff
@@ -119,6 +123,19 @@ function validateOptions(
     }
```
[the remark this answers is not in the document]
mrbbot: Hmmm, I think removing the `as const` should be ok. It looks like that still keeps the specific Zod types so options inference/completions still work. Will make that change. 👍

packages/tre/src/index.ts
```diff
@@ -150,6 +167,19 @@ function getDurableObjectClassNames(
...
 }
 
+// Collects all routes from all worker services
+function getWorkerRoutes(
+  allWorkerOpts: PluginWorkerOptions[]
+): Map<string, string[]> {
+  const allRoutes = new Map<string, string[]>();
+  for (const workerOpts of allWorkerOpts) {
+    if (workerOpts.core.routes !== undefined) {
+      allRoutes.set(workerOpts.core.name ?? "", workerOpts.core.routes);
```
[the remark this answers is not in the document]
mrbbot: It would, but we've validate names are unique beforehand. Perhaps an assertion that `allRoutes` doesn't contain the key would be a good idea?

> I wonder if worker names should be required

In the single worker case, I don't think they should be. I do agree with you later on though that some of the error messages should be better in this case.

packages/tre/src/index.ts
```diff
@@ -633,10 +672,13 @@ export class Miniflare {
       },
     ];
 
-    const durableObjectClassNames = getDurableObjectClassNames(allWorkerOpts);
-
     // Dedupe services by name
     const serviceNames = new Set<string>();
```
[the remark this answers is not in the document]
mrbbot: That's a much better idea 😃 , will make that switch 👍

packages/tre/src/plugins/core/index.ts
mrbbot: This was so we could use it in `SERVICE_LOOPBACK` without a cyclic dependency on `core`:

https://github.com/cloudflare/miniflare/blob/9229b3085966b28ddeaf6aa804354899bdf9dcdc/packages/tre/src/plugins/shared/constants.ts#L10

Though I think moving global services to shared makes a lot more sense. Then we could just rename this `loopback` or something.

packages/tre/src/plugins/core/index.ts
mrbbot: Yes... 🙈 We should probably be properly bundling these as you elude to in the next comment.

packages/tre/src/plugins/core/index.ts
mrbbot: Yeah, we should probably be properly bundling these and writing them in TypeScript. We could still have some shared constants imported by both the Workers and Node code. Would need to make sure these were separate though so we didn't pollute type environments. Something for a future PR I think. 👍

packages/tre/src/plugins/core/index.ts
mrbbot: Hmmm, I suppose that doesn't really make sense, since we don't expose `workered`'s other service types (disk, network, etc) outside of `serviceBindings`. Will make code required for all Workers. 👍

packages/tre/src/plugins/core/index.ts
mrbbot: I don't think so, they probably reference this service, but they shouldn't be defining it themselves? 😕

packages/tre/src/plugins/shared/routing.ts
mrbbot: I haven't, you're correct this was copied from Miniflare 2, when I didn't have access to this. Will see if I can find it. 👍

packages/tre/src/index.ts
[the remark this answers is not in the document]
mrbbot: As discussed, this doesn't seem to be working in TypeScript 4, we can revisit this later. 👍

packages/tre/src/index.ts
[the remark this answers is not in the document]
mrbbot: f095aadbcc2f383e89b52f4faf3b040e8552b570

packages/tre/src/plugins/core/index.ts
```diff
@@ -451,6 +419,74 @@ export const CORE_PLUGIN: Plugin<
...
 };
 
+export interface GlobalServicesOptions {
+  optionsVersion: number;
+  sharedOptions: z.infer<typeof CoreSharedOptionsSchema>;
+  allWorkerRoutes: Map<string, string[]>;
+  fallbackWorkerName: string | undefined;
+  loopbackPort: number;
+}
+export function getGlobalServices({
```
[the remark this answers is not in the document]
mrbbot: 4e99a58fa4a901a3dd33ea167df50e2b4f310b63

packages/tre/src/index.ts
```diff
@@ -119,6 +123,19 @@ function validateOptions(
...
 
+  // Validate names unique
+  const names = new Set<string>();
+  for (const opts of pluginWorkerOpts) {
+    const name = opts.core.name ?? "";
+    if (names.has(name)) {
+      throw new MiniflareCoreError(
+        "ERR_DUPLICATE_NAME",
+        `Multiple workers defined with the same name: "${name}"`
+      );
```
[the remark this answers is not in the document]
mrbbot: 1abbdbc716432139efcd05a0468a758f94e2c99f

packages/tre/src/plugins/cache/index.ts
mrbbot: 0e2c68d822be0cf3cd4a8a5a903d22534929c149

packages/tre/src/plugins/core/index.ts
mrbbot: 4e99a58f

packages/tre/src/plugins/core/index.ts
mrbbot: 839b02c8

packages/tre/src/plugins/shared/routing.ts
mrbbot: 0ed30499689e52085cc76a2439abd29ba3a90632

mrbbot: (and a629360c to fixup test)

penalosa (approved):

PR #524, opened 2023-02-27
[Miniflare 3] Implement more of R2 and add tests

mrbbot:
This PR brings over some of the improvements from https://github.com/cloudflare/miniflare/pull/470, and adds tests for the R2 implementation.

Specifically we now: _(see commit descriptions for more details)_

- Validate requests from `workerd` using `zod` to ensure type safety in the rest of the implementation *(chunky)*
- Support deleting multiple keys in `R2Bucket#delete()`
- Support ranged-gets with a `Range` header in `R2Bucket#get()`
- Return `range` information from `R2Bucket#head()`
- Support the `R2Bucket#list()` `startAfter` option
- Support `sha*` checksums in `R2Bucket#put()` and return these from `R2Bucket#{get,head}`
- Support conditional `R2Bucket#put()`
- Support `secondsGranularity` option in `R2Conditional`s
- Remove unpaired surrogate pair key validation
- Make `R2Object#version` a hex string to match the real implementation
- Validate `customMetadata` size
- Test the R2 implantation  *(chunky)*

Best reviewed commit-by-commit... 😅


packages/tre/src/plugins/r2/gateway.ts
```diff
@@ -59,12 +60,26 @@ export class R2Gateway {
...
+    validate.key(key).condition(meta, options.onlyIf);
+
+    let range = options.range ?? {};
+    if (options.rangeHeader !== undefined) {
+      const ranges = _parseRanges(options.rangeHeader, meta.size);
+      if (ranges?.length === 1) {
+        // If the header contained a single range, convert it to an R2Range.
+        // Note `start` and `end` are inclusive.
+        const [start, end] = ranges[0];
+        range = { offset: start, length: end - start + 1 };
```
penalosa, line 75: `+ 1`? I'm sure I'm being silly, but why is this needed?

packages/tre/src/plugins/r2/gateway.ts
```diff
@@ -59,12 +60,26 @@ export class R2Gateway {
...
+      } else {
+        // If the header was invalid, or contained multiple ranges, just return
```
penalosa, line 77: Are multiple ranges not supported?

.github/workflows/test.yml
```diff
@@ -18,14 +18,14 @@ jobs:
 
     strategy:
       matrix:
-        os: [ubuntu-latest, windows-latest]
-        node: [16.13.0, 18.x]
+        os: [ubuntu-latest] # `windows-latest` not yet supported
```
penalosa, line 21: Is this removing testing for Miniflare 2 on Windows as well?

package.json
```diff
@@ -16,6 +16,7 @@
     "packages/*"
   ],
   "scripts": {
+    "postinstall": "patch-package",
```
penalosa, line 19: Why is this needed?

packages/tre/src/plugins/r2/validator.ts
```diff
@@ -67,6 +70,14 @@ export type R2Hashes = Record<
   Buffer | undefined
 >;
 
+function serialisedLength(x: string) {
+  //  Adapted from internal R2 gateway implementation
+  for (let i = 0; i < x.length; i++) {
+    if (x.charCodeAt(i) >= 256) return x.length * 2;
```
penalosa, line 76: What is this doing? This feels ...fishy. I may be wrong, but this _seems_ to be saying "if _any_ character has a character code greater than 256, return the string length doubled". If this is for checking the serialised length, what about serialising the string with a `TextEncoder` and checking the byte length?

packages/tre/package.json
```diff
@@ -42,7 +42,7 @@
     "workerd": "^1.20230221.0",
     "ws": "^8.11.0",
     "youch": "^3.2.2",
-    "zod": "^3.18.0"
+    "zod": "^3.20.6"
```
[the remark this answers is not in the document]
mrbbot: Nothing crazy, most significant changes seem to be a new `pipe()` method and `z.coerce` which is what we updated for. 👍

packages/tre/src/plugins/r2/schemas.ts
```diff
@@ -0,0 +1,243 @@
...
+    3: HexDataSchema.optional(),
+    4: HexDataSchema.optional(),
+  })
+  .transform((checksums) => ({
+    md5: checksums["0"],
+    sha1: checksums["1"],
+    sha256: checksums["2"],
+    sha384: checksums["3"],
+    sha512: checksums["4"],
+  }));
```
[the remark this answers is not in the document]
mrbbot: The runtime will send an object like `{ 0: "aaa...", 2: "bbb..." }`. We want to validate against that, but then transform that object into `{md5: Buffer, sha256: Buffer}`. `md5: HexDataSchema.optional()` would fail because the runtime isn't sending objects with `md5` keys.

packages/tre/src/plugins/r2/gateway.ts
mrbbot: They are not 😞 Just checked and R2 returns the full range here.

.github/workflows/test.yml
mrbbot: It would be, if we ever merged `tre` into `master`, but because we're not planning to do that, so no. 🙂

package.json
mrbbot: We [patch `capnp-ts`](https://github.com/cloudflare/miniflare/blob/tre/patches/capnp-ts%2B0.7.0.patch) to fix type checking under `--isolatedModules`. I submitted a PR to fix this but it hasn't been merged yet: https://github.com/jdiaz5513/capnp-ts/pull/170. It wasn't until I added GitHub actions CI that I realised I'd missed this `postinstall` script off when installing `patch-package`.

packages/tre/src/plugins/r2/validator.ts
mrbbot: Yeah... it's definitely weird. I had the same thought when I saw the actual implementation. I'll ping you internally. 👍

packages/tre/src/plugins/r2/errors.ts
```diff
@@ -22,22 +23,20 @@ enum CfCode {
   InvalidRange = 10039,
 }
 
-export class R2Error extends Error {
-  status: number;
+export class R2Error extends HttpError {
   v4Code: number;
   object?: R2Object;
+
   constructor(status: number, message: string, v4Code: number) {
```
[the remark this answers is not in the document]
mrbbot: ace5152a0e48f1e071b40e9f11bcc7d7329bc491

penalosa (approved):

Issue #529, opened 2023-03-02
miniflare d1 in jest fails on queries without data results


jschlesser:
I get errors when using d1 inside of jest (to create temp d1 databases in tests) when trying to execute queries that don't return data in their results.  create table, insert, etc... ,    select queries run just fine once tables are created and data is present.

I don't know exactly why executing better-sqlite3 in a jest environment does this but somehow TypeError is turning into a plain object.  However, the execution path for non data result queries is dependent on it being a TypeError, see [here](https://github.com/cloudflare/miniflare/blob/3dfd60bd2047a4a95e9110964a577b2dcafb0827/packages/d1/src/api.ts#L105)

Code execution down that path in the failure mode results in the following error when executing something like a create table or insert query.

`TypeError: The columns() method is only for statements that return data`

I have created a [project](https://github.com/jschlesser/miniflare-d1-jest-debug) that demonstrates this.  Note that the jest config and tsconfig are based on the configs for writing an adapter for auth.js.  I encountered this error in the process of developing a D1 adapter for that project.

The project is set up so you only have to clone it, `npm install`, `npm run build`,  `npm run dev` (non jest, success), `npm run test` (inside of jest, fail).

My environment is an M1 Mac running the latest OS and node version v19.6.0 which is the system node version.

I think/hope that the fix is fairly simple, changing the guard in that section of code to look for an instance of `TypeError` or `object`.   I tested that change by hacking the index.js in the node_modules/@miniflare/d1/dist/src/index.js and it allowed everything to work as I expected.  I'm happy to make a PR if that is the correct way to fix.

jschlesser:
Another quick note.
It looks like d1 is using the fact that a call to `stmt.columns()` on a SqliteStatement throws an error to determine whether or not columns are returned.  I thought maybe there is a better way to determine that. After looking at the better-sqlite3 code it wasn't obvious to me that there is a way.  It looks like it's baked into the native part of the code.   I suppose that's why trapping the error and looking at the error type and message is the method being used in the first place.  It's not obvious to me that there is a 'cleaner' way to resolve the issue in d1 combined with jest outside of abandoning swc and typescript support in jest.  I'm not sure that would solve it either, I was just thinking about a cleaner solution to figuring out if a query has a data result.

jschlesser:
[...]

PR #533, opened 2023-03-05
[miniflare D1] replaces returnsData in statement with a natively supported function

jschlesser:
I would like to point out that while this PR simplifies the codebase and makes it less fragile and implements the actual intent and is probably better for the longer term, it may not be the exact same behavior.  However, its not dependent on an error message changing and it will avoid issues arising from TypeError not working working with instanceof comparisons in certain uses of the vm api to execute code.  Also D1 is technically beta 🤷

### DIFF
--- a/packages/d1/src/api.ts
+++ b/packages/d1/src/api.ts
@@ -92,22 +92,8 @@ function normaliseResults(rows: any[]): any[] {
   );
 }
 
-const DOESNT_RETURN_DATA_MESSAGE =
-  "The columns() method is only for statements that return data";
 const EXECUTE_RETURNS_DATA_MESSAGE =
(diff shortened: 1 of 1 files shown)


jschlesser:
References this [issue](https://github.com/cloudflare/miniflare/issues/529)

jschlesser:
Also note that this property is a direct pass through to the native sqlite returns_data method
[source code reference](https://github.com/WiseLibs/better-sqlite3/blob/17ed0689217aafee4c78ce42dec559c716cf4ead/src/objects/statement.lzz#L141)

The original D1 miniflare returnsData method could technically throw an exception, It's not clear to me that this patch ever would.

mrbbot (approved):

PR #540, opened 2023-03-20
Implement Durable Object jurisdiction API

DaniFoldi:
This Pull Request implements a `.jurisdiction` method on Durable Object namespaces, which just returns the original namespace.

The jurisdiction option is ignored, just like the option in `newUniqueId`. In the real runtime, these would create a different ID, but I suppose it's not often that these are used to differentiate between IDs, and changing this behaviour would be a breaking change.

### DIFF
--- a/packages/durable-objects/src/namespace.ts
+++ b/packages/durable-objects/src/namespace.ts
@@ -230,6 +230,11 @@ export class DurableObjectNamespace {
     this.#ctx = ctx;
   }
 
+  jurisdiction(_name: string): DurableObjectNamespace {
+    // Ignore jurisdiction restrictions in local mode
+    return this;
+  }
+
   newUniqueId(_options?: NewUniqueIdOptions): DurableObjectId {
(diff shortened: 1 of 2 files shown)


packages/durable-objects/src/namespace.ts
```diff
@@ -230,6 +230,11 @@ export class DurableObjectNamespace {
     this.#ctx = ctx;
   }
 
+  jurisdiction(_name: string): DurableObjectNamespace {
+    // Ignore jurisdiction restrictions in local mode
```
mrbbot, line 234: Does the Workers runtime perform any validation on `_name` here? Might be nice to implement that if it does, so users know their jurisdictions are valid.

DaniFoldi: The runtime throws a TypeError if an unsupported value is supplied, which should now be the case here as well, and I've added a test case for it as well.

mrbbot (approved):

PR #542, opened 2023-03-22
Implement non-standard SubtleCrypto timingSafeEqual

DaniFoldi:
`workerd` implements a non-standard extension method `timingSafeEqual` to the Web Crypto API under Subtle Crypto. Fortunately, node also provides a similar implementation, under the regular `crypto`.

This PR is not concerned about the safety of the implementation, as miniflare is intended for local development purposes only.

Marking as draft as the types for crypto.subtle come from node's types in miniflare and typecheck fails. @mrbbot how should we approach overriding the type for the crypto module? The method exists in [workerd](https://github.com/cloudflare/workerd/blob/9e5d8f18f172c207a95334b22910678498c6c078/src/workerd/api/crypto.h#L541) and workers-types.

### DIFF
--- a/packages/core/test/standards/crypto.spec.ts
+++ b/packages/core/test/standards/crypto.spec.ts
@@ -184,6 +184,22 @@ test("crypto: sign/verify: supports other algorithm", async (t) => {
   t.true(await crypto.subtle.verify("HMAC", key, signature, data));
 });
 
+test("crypto: timingSafeEqual equals", (t) => {
+  const array1 = new Uint8Array(12);
+  array1.fill(0, 0);
+  const array2 = new Uint8Array(12);
+  array2.fill(0, 0);
+  t.is(crypto.subtle.timingSafeEqual(array1, array2), true);
+});
+test("crypto: timingSafeEqual not equals", (t) => {
+  const array1 = new Uint8Array(12);
+  array1.fill(0, 0);
+  const array2 = new Uint8Array(12);
+  array2.fill(0, 0);
+  array2[7] = 1;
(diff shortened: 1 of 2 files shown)


packages/core/test/standards/crypto.spec.ts
```diff
@@ -184,6 +184,22 @@ test("crypto: sign/verify: supports other algorithm", async (t) => {
   t.true(await crypto.subtle.verify("HMAC", key, signature, data));
 });
 
+test("crypto: timingSafeEqual equals", (t) => {
+  const array1 = new Uint8Array(12);
+  array1.fill(0, 0);
+  const array2 = new Uint8Array(12);
+  array2.fill(0, 0);
+  t.is(crypto.subtle.timingSafeEqual(array1, array2), true);
```
mrbbot, line 192: Nit:
```suggestion
  t.true(crypto.subtle.timingSafeEqual(array1, array2));
```

packages/core/test/standards/crypto.spec.ts
```diff
@@ -184,6 +184,22 @@ test("crypto: sign/verify: supports other algorithm", async (t) => {
...
+});
+test("crypto: timingSafeEqual not equals", (t) => {
+  const array1 = new Uint8Array(12);
+  array1.fill(0, 0);
+  const array2 = new Uint8Array(12);
+  array2.fill(0, 0);
+  array2[7] = 1;
+  t.is(crypto.subtle.timingSafeEqual(array1, array2), false);
```
mrbbot, line 200:
```suggestion
  t.false(crypto.subtle.timingSafeEqual(array1, array2));
```

mrbbot (approved):

PR #550, opened 2023-03-30
Add logging for requests to miniflare

jspspike:
Addresses [this issue](https://github.com/cloudflare/workers-sdk/issues/2722)

### DIFF
--- a/packages/tre/src/plugins/core/index.ts
+++ b/packages/tre/src/plugins/core/index.ts
@@ -130,6 +130,7 @@ export const SCRIPT_ENTRY = `
 const matchRoutes = ${matchRoutes.toString()};
 
 async function handleEvent(event) {
+  const startTime = Date.now();
   const probe = event.request.headers.get("${HEADER_PROBE}");
   if (probe !== null) {
     const probeMin = parseInt(probe);
@@ -177,6 +178,18 @@ async function handleEvent(event) {
         });
       }
     }
+
+    event.waitUntil(${BINDING_SERVICE_LOOPBACK}.fetch("http://localhost/core/log", {
+        method: "POST",
+        headers: { "Content-Type": "application/json" },
+        body: JSON.stringify({
+          "status": response.status,
+          "statusText": response.statusText,
+          "method": event.request.method,
+          "url": response.url,
+          "time": Date.now() - startTime,
+        }),
+    }));
 
     const liveReloadScript = globalThis.${BINDING_DATA_LIVE_RELOAD_SCRIPT};
     if (
--- a/packages/tre/src/shared/log.ts
+++ b/packages/tre/src/shared/log.ts
@@ -1,5 +1,17 @@
 import path from "path";
-import { Colorize, dim, green, grey, red, reset, yellow } from "kleur/colors";
+import {
+  Colorize,
+  blue,
+  bold,
+  dim,
+  green,
+  grey,
+  red,
+  reset,
+  yellow,
+} from "kleur/colors";
+import { z } from "zod";
+import { Request } from "../http";
 
 const cwd = process.cwd();
 const cwdNodeModules = path.join(cwd, "node_modules");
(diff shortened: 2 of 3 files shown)


packages/tre/src/plugins/core/log.ts
```diff
@@ -0,0 +1,34 @@
...
+export interface ResponseInfo {
+  status: number;
+  statusText: string;
+  method: string;
+  url: string;
+  time: number;
+}
+
+export async function formatResponse(request: Request) {
+  const info: ResponseInfo = JSON.parse(await request.text());
```
mrbbot, line 13:
```suggestion
  const info: ResponseInfo = await request.json();
```

packages/tre/src/plugins/core/index.ts
```diff
@@ -178,6 +178,20 @@ async function handleEvent(event) {
       }
     }
 
+    const headers = new Headers();
+    headers.append("Content-Type", "application/json")
+    await ${BINDING_SERVICE_LOOPBACK}.fetch("http://localhost/core/log", {
```
mrbbot, line 183: Could we avoid blocking returning the response here, by doing this `fetch` in the background?
```suggestion
    event.waitUntil(${BINDING_SERVICE_LOOPBACK}.fetch("http://localhost/core/log", {
```
(...and the corresponding `)`)

mrbbot (commented):
Nice! 🎉 Thanks for doing this! 😃 I've added some comments...

packages/tre/src/plugins/core/index.ts
```diff
@@ -178,6 +178,20 @@ async function handleEvent(event) {
...
+        method: "POST",
+        headers: headers,
+        body: JSON.stringify({
+            "status": response.status,
+            "statusText": response.statusText,
+            "method": event.request.method,
+            "url": response.url,
+            "time": Date.now(),
```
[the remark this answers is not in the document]
jspspike: So `Date.now()` in a worker only returns the time from the beginning of the request according to [this documentation](https://developers.cloudflare.com/workers/learning/security-model#step-1-disallow-timers-and-multi-threading). So we'd have to use separate timers if we want to support this feature. I could call `Date.now()` outside of `formatResponse` but thought it would be essentially the same.

packages/tre/src/plugins/core/log.ts
jspspike: So I did this originally but got a type error since `json()` returns `unknown`, do you know how I would be able to fix that?

packages/tre/src/plugins/core/index.ts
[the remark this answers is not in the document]
mrbbot: Ah yeah, `Date.now()` in `workerd` always returns the actual time. Those security considerations aren't too important for single tenant runtimes like `workerd`.

packages/tre/src/shared/log.ts
```diff
@@ -121,3 +133,41 @@ export class NoOpLog extends Log {
...
+
+export interface ResponseInfo {
+  status: number;
+  statusText: string;
+  method: string;
+  url: string;
+  time: number;
+}
+
+const ResponseInfoSchema: z.ZodType<ResponseInfo> = z.object({
```
mrbbot, line 145: TypeScript should be able to infer the correct types from `zod` here.
```suggestion
const ResponseInfoSchema = z.object({
```

packages/tre/src/shared/log.ts
```diff
@@ -121,3 +133,41 @@ export class NoOpLog extends Log {
...
+  status: z.number(),
+  statusText: z.string(),
+  method: z.string(),
+  url: z.string(),
+  time: z.number(),
+});
+
+export async function formatResponse(request: Request) {
+  const info: ResponseInfo = ResponseInfoSchema.parse(await request.json());
```
mrbbot, line 154:
```suggestion
  const info = ResponseInfoSchema.parse(await request.json());
```

packages/tre/src/plugins/core/index.ts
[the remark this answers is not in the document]
jspspike: Ah I didn't realize that it was different workerd

packages/tre/src/plugins/core/index.ts
```diff
@@ -178,6 +179,18 @@ async function handleEvent(event) {
...
+        method: "POST",
+        headers: { "Content-Type": "application/json" },
+        body: JSON.stringify({
+            "status": response.status,
+            "statusText": response.statusText,
+            "method": event.request.method,
+            "url": response.url,
+            "time": Date.now() - startTime,
+        }),
+    }));
```
mrbbot, line 192: It looks like this is indented with 4 spaces rather than 2. 🙂 Guess the linter isn't working since this is a string literal.

mrbbot (approved):
Nice! Approve with one tiny nit 🙂

Issue #235, opened 2022-04-05
List isn't in lexicographical order


vlovich:
I'm using fast-check to fuzz test our behavior and ran into what looks like a Miniflare bug.

```
  await storage.put({
    '!': {},
    ', ': {},
  })
  const keys = Array.from((await storage.list()).keys())
  console.log(keys)
```

This prints `[ ', ', '!' ]` when in actuality it should print `[ '!', ', ' ]`. I verified this to be the case both by examining the output of `[ ', ', '!' ].sort()` in Node and testing this behavior in Cloudflare.

vlovich:
Might be useful to write a fuzz check that list always returns things in lexicographical order because there's other failing test cases. Here's an example test case I have:

```
test('mf list is lexicographical', async () => {
  mf = new Miniflare({
    modules: true,
    buildCommand: '',
    wranglerConfigEnv: 'dev',
    scriptPath: 'worker/index.mjs',
    watch: false,
    envPath: false,
    cache: false,
    durableObjectsPersist: false,
    durableObjects: {
      DurableObject: 'DurableObject',
    },
  })

  const storage = await mf.getDurableObjectStorage((await mf.getBindings()).DurableObject.newUniqueId())

  await fc.assert(fc.asyncProperty(
    fc.array(fc.string({ minLength: 0, maxLength: 1024 }), {minLength: 1, maxLength: 10}),
    async (keys) => {
      await storage.deleteAll()
      const records: Record<string, object> = {}
      for (const key of keys) {
        records[key] = {}
      }
      await storage.put(records)
      expect(Array.from((await storage.list()).keys())).toEqual(keys.sort())
    }
  ))
})
```

vlovich:
Not sure if this will catch more tests (as it could be the same underlying bug), but this is also broken:
```
test('mf list is lexicographical', async () => {
  const storage = await mf.getDurableObjectStorage(
    (await mf.getBindings()).BucketStore.newUniqueId(),
  )

  await fc.assert(
    fc.asyncProperty(
      fc.array(fc.string({ minLength: 0, maxLength: 1024 }), {
        minLength: 1,
        maxLength: 10,
      }),
      fc.string({ minLength: 0, maxLength: 1024 }),
      async (keys, start) => {
        await storage.deleteAll()
        const records: Record<string, object> = {}
        for (const key of keys) {
          records[key] = {}
        }
        await storage.put(records)
        expect(
          Array.from((await storage.list({ start: start })).keys()).every(
[...]

PR #555, opened 2023-04-13
[Miniflare 3] Implement new storage system

mrbbot:
This PR implements a new underlying storage system for Miniflare, based off #525 and internal discussions. The storage system is split into 2 parts: an SQLite-backed metadata store, and a streaming file-system backed/in-memory blob store. Blobs use unguessable identifiers to support atomic transactions with the SQLite metadata store. See `BlobStore` for more details. This is a breaking change to the persistence format.

This PR also re-implements the KV gateway using the new storage system, as an example of what a data store consuming this system might look like. Notably, this implementation uses an intermediate expiring key-value-metadata store, that will be shared by the Cache gateway in the future.

Best reviewed commit-by-commit.

Closes DEVX-583, DEVX-584, DEVX-585, DEVX-586, DEVX-587, DEVX-588 and DEVX-589.

### DIFF
--- a/packages/tre/src/plugins/kv/gateway.ts
+++ b/packages/tre/src/plugins/kv/gateway.ts
@@ -87,7 +87,7 @@ export interface KVGatewayGetOptions {
 export interface KVGatewayGetResult<Metadata = unknown> {
   value: ReadableStream<Uint8Array>;
   expiration?: number; // seconds since unix epoch
-  metadata: Metadata;
+  metadata?: Metadata;
 }
 
 export interface KVGatewayPutOptions<Metadata = unknown> {
@@ -107,10 +107,50 @@ export interface KVGatewayListKey {
   expiration?: number; // seconds since unix epoch
   metadata?: string; // JSON-stringified metadata
 }
-export interface KVGatewayListResult {
+export type KVGatewayListResult = {
   keys: KVGatewayListKey[];
-  cursor?: string;
-  list_complete: boolean;
+} & (
+  | { list_complete: false; cursor: string }
+  | { list_complete: true; cursor: undefined }
+);
+
+export function validateGetOptions(
+  key: string,
+  options?: KVGatewayGetOptions
+): void {
+  validateKey(key);
+  // Validate cacheTtl, but ignore it as there's only one "edge location":
+  // the user's computer
+  const cacheTtl = options?.cacheTtl;
+  if (cacheTtl !== undefined && (isNaN(cacheTtl) || cacheTtl < MIN_CACHE_TTL)) {
+    throw new KVError(
+      400,
+      `Invalid ${PARAM_CACHE_TTL} of ${cacheTtl}. Cache TTL must be at least ${MIN_CACHE_TTL}.`
+    );
+  }
+}
+
+export function validateListOptions(options: KVGatewayListOptions): void {
+  // Validate key limit
+  const limit = options.limit;
+  if (limit !== undefined) {
+    if (isNaN(limit) || limit < 1) {
+      throw new KVError(
+        400,
+        `Invalid ${PARAM_LIST_LIMIT} of ${limit}. Please specify an integer greater than 0.`
+      );
+    }
+    if (limit > MAX_LIST_KEYS) {
+      throw new KVError(
+        400,
+        `Invalid ${PARAM_LIST_LIMIT} of ${limit}. Please specify an integer less than ${MAX_LIST_KEYS}.`
+      );
+    }
+  }
+
+  // Validate key prefix
+  const prefix = options.prefix;
+  if (prefix !== undefined) validateKeyLength(prefix);
 }
 
 export class KVGateway {
@@ -129,20 +169,7 @@ export class KVGateway {
     key: string,
     options?: KVGatewayGetOptions
   ): Promise<KVGatewayGetResult<Metadata> | undefined> {
-    validateKey(key);
-    // Validate cacheTtl, but ignore it as there's only one "edge location":
-    // the user's computer
-    const cacheTtl = options?.cacheTtl;
-    if (
-      cacheTtl !== undefined &&
-      (isNaN(cacheTtl) || cacheTtl < MIN_CACHE_TTL)
-    ) {
-      throw new KVError(
-        400,
-        `Invalid ${PARAM_CACHE_TTL} of ${cacheTtl}. Cache TTL must be at least ${MIN_CACHE_TTL}.`
-      );
-    }
-
+    validateGetOptions(key, options);
     const entry = await this.storage.get(key);
     if (entry === null) return;
     return {
@@ -233,36 +260,19 @@ export class KVGateway {
   }
 
   async list(options: KVGatewayListOptions = {}): Promise<KVGatewayListResult> {
-    // Validate key limit
-    const limit = options.limit ?? MAX_LIST_KEYS;
-    if (isNaN(limit) || limit < 1) {
-      throw new KVError(
-        400,
-        `Invalid ${PARAM_LIST_LIMIT} of ${limit}. Please specify an integer greater than 0.`
-      );
-    }
-    if (limit > MAX_LIST_KEYS) {
-      throw new KVError(
-        400,
-        `Invalid ${PARAM_LIST_LIMIT} of ${limit}. Please specify an integer less than ${MAX_LIST_KEYS}.`
-      );
-    }
-
-    // Validate key prefix
-    const prefix = options.prefix;
-    if (prefix !== undefined) validateKeyLength(prefix);
-
-    const cursor = options.cursor;
+    validateListOptions(options);
+    const { limit = MAX_LIST_KEYS, prefix, cursor } = options;
     const res = await this.storage.list({ limit, prefix, cursor });
-    return {
-      keys: res.keys.map((key) => ({
-        name: key.key,
-        expiration: maybeApply(millisToSeconds, key.expiration),
-        // workerd expects metadata to be a JSON-serialised string
-        metadata: maybeApply(JSON.stringify, key.metadata),
-      })),
-      cursor: res.cursor,
-      list_complete: res.cursor === undefined,
-    };
+    const keys = res.keys.map<KVGatewayListKey>((key) => ({
+      name: key.key,
+      expiration: maybeApply(millisToSeconds, key.expiration),
+      // workerd expects metadata to be a JSON-serialised string
+      metadata: maybeApply(JSON.stringify, key.metadata),
+    }));
+    if (res.cursor === undefined) {
+      return { keys, list_complete: true, cursor: undefined };
+    } else {
+      return { keys, list_complete: false, cursor: res.cursor };
+    }
   }
 }
--- a/packages/tre/src/plugins/kv/gateway.ts
+++ b/packages/tre/src/plugins/kv/gateway.ts
@@ -1,5 +1,14 @@
-import { Clock, HttpError, Log, millisToSeconds } from "../../shared";
-import { Storage, StoredKeyMeta, StoredValueMeta } from "../../storage";
+import { ReadableStream, TransformStream } from "stream/web";
+import {
+  Clock,
+  HttpError,
+  Log,
+  maybeApply,
+  millisToSeconds,
+  secondsToMillis,
+} from "../../shared";
+import { Storage } from "../../storage";
+import { KeyValueStorage } from "../../storage2";
 import {
   MAX_KEY_SIZE,
   MAX_LIST_KEYS,
@@ -47,14 +56,45 @@ function normaliseInt(value: string | number | undefined): number | undefined {
   }
 }
 
+function createMaxValueSizeError(length: number) {
+  return new KVError(
+    413,
+    `Value length of ${length} exceeds limit of ${MAX_VALUE_SIZE}.`
+  );
+}
+class MaxLengthStream extends TransformStream<Uint8Array, Uint8Array> {
+  constructor(maxLength: number, errorFactory: (length: number) => Error) {
+    let length = 0;
+    super({
+      transform(chunk, controller) {
+        length += chunk.byteLength;
+        // If we exceeded the maximum length, don't enqueue the chunk as we'll
+        // be aborting the stream, but don't error just yet, so we get the
+        // correct final length in the error
+        if (length <= maxLength) controller.enqueue(chunk);
+      },
+      flush(controller) {
+        // If we exceeded the maximum length, abort the stream
+        if (length > maxLength) controller.error(errorFactory(length));
+      },
+    });
+  }
+}
+
 export interface KVGatewayGetOptions {
   cacheTtl?: number;
 }
+export interface KVGatewayGetResult<Metadata = unknown> {
+  value: ReadableStream<Uint8Array>;
+  expiration?: number; // seconds since unix epoch
+  metadata: Metadata;
+}
 
-export interface KVGatewayPutOptions<Meta = unknown> {
-  expiration?: string | number;
-  expirationTtl?: string | number;
-  metadata?: Meta;
+export interface KVGatewayPutOptions<Metadata = unknown> {
+  expiration?: string | number; // seconds since unix epoch
+  expirationTtl?: string | number; // seconds relative to now
+  metadata?: Metadata;
+  valueLengthHint?: number;
 }
 
 export interface KVGatewayListOptions {
@@ -62,23 +102,33 @@ export interface KVGatewayListOptions {
   prefix?: string;
   cursor?: string;
 }
-export interface KVGatewayListResult<Meta = unknown> {
-  keys: StoredKeyMeta<Meta>[];
+export interface KVGatewayListKey {
+  name: string;
+  expiration?: number; // seconds since unix epoch
+  metadata?: string; // JSON-stringified metadata
+}
+export interface KVGatewayListResult {
+  keys: KVGatewayListKey[];
   cursor?: string;
   list_complete: boolean;
 }
 
 export class KVGateway {
+  private readonly storage: KeyValueStorage;
+
   constructor(
     private readonly log: Log,
-    private readonly storage: Storage,
+    legacyStorage: Storage,
     private readonly clock: Clock
-  ) {}
+  ) {
+    const storage = legacyStorage.getNewStorage();
+    this.storage = new KeyValueStorage(storage, clock);
+  }
 
-  async get(
+  async get<Metadata = unknown>(
     key: string,
     options?: KVGatewayGetOptions
-  ): Promise<StoredValueMeta | undefined> {
+  ): Promise<KVGatewayGetResult<Metadata> | undefined> {
     validateKey(key);
     // Validate cacheTtl, but ignore it as there's only one "edge location":
     // the user's computer
@@ -92,12 +142,19 @@ export class KVGateway {
         `Invalid ${PARAM_CACHE_TTL} of ${cacheTtl}. Cache TTL must be at least ${MIN_CACHE_TTL}.`
       );
     }
-    return this.storage.get(key, false, cacheTtl);
+
+    const entry = await this.storage.get(key);
+    if (entry === null) return;
+    return {
+      value: entry.value,
+      expiration: maybeApply(millisToSeconds, entry.expiration),
+      metadata: entry.metadata as Metadata,
+    };
   }
 
   async put(
     key: string,
-    value: Uint8Array,
+    value: ReadableStream<Uint8Array>,
     options: KVGatewayPutOptions = {}
   ): Promise<void> {
     validateKey(key);
@@ -135,13 +192,7 @@ export class KVGateway {
       }
     }
 
-    // Validate value and metadata size
-    if (value.byteLength > MAX_VALUE_SIZE) {
-      throw new KVError(
-        413,
-        `Value length of ${value.byteLength} exceeds limit of ${MAX_VALUE_SIZE}.`
-      );
-    }
+    // Validate metadata size
     if (options.metadata !== undefined) {
       const metadataJSON = JSON.stringify(options.metadata);
       const metadataLength = Buffer.byteLength(metadataJSON);
@@ -153,9 +204,25 @@ export class KVGateway {
       }
     }
 
-    return this.storage.put(key, {
+    // Validate value size
+    const valueLengthHint = options.valueLengthHint;
+    if (valueLengthHint !== undefined && valueLengthHint > MAX_VALUE_SIZE) {
+      // If we know the size of the value (i.e. from `Content-Length`) use that
+      throw createMaxValueSizeError(valueLengthHint);
+    } else {
+      // Otherwise, pipe through a transform stream that counts the number of
+      // bytes and errors if it exceeds the max. This error will be thrown
+      // within the `storage.put()` call below and will be propagated up to the
+      // caller.
+      value = value.pipeThrough(
+        new MaxLengthStream(MAX_VALUE_SIZE, createMaxValueSizeError)
+      );
+    }
+
+    return this.storage.put({
+      key,
       value,
-      expiration,
+      expiration: maybeApply(secondsToMillis, expiration),
       metadata: options.metadata,
     });
   }
@@ -189,13 +256,13 @@ export class KVGateway {
     const res = await this.storage.list({ limit, prefix, cursor });
     return {
       keys: res.keys.map((key) => ({
-        ...key,
+        name: key.key,
+        expiration: maybeApply(millisToSeconds, key.expiration),
         // workerd expects metadata to be a JSON-serialised string
-        metadata:
-          key.metadata === undefined ? undefined : JSON.stringify(key.metadata),
+        metadata: maybeApply(JSON.stringify, key.metadata),
       })),
-      cursor: res.cursor === "" ? undefined : res.cursor,
-      list_complete: res.cursor === "",
+      cursor: res.cursor,
+      list_complete: res.cursor === undefined,
     };
   }
 }
--- a/packages/tre/src/plugins/kv/router.ts
+++ b/packages/tre/src/plugins/kv/router.ts
@@ -20,7 +20,15 @@ import {
   PARAM_LIST_PREFIX,
   PARAM_URL_ENCODED,
 } from "./constants";
-import { KVError, KVGateway } from "./gateway";
+import {
+  KVError,
+  KVGateway,
+  KVGatewayGetOptions,
+  KVGatewayGetResult,
+  KVGatewayListOptions,
+  KVGatewayListResult,
+} from "./gateway";
+import { sitesGatewayGet, sitesGatewayList } from "./sites";
 
 export interface KVParams {
   namespace: string;
@@ -45,20 +53,24 @@ export class KVRouter extends Router<KVGateway> {
   get: RouteHandler<KVParams> = async (req, params, url) => {
     // Get gateway with (persistent) storage
     const persist = decodePersist(req.headers);
-    // Workers Sites: if this is a sites request, persist should be used as the
-    // root without any additional namespace
-    const namespace =
-      req.headers.get(HEADER_SITES) === null ? params.namespace : "";
-    const gateway = this.gatewayFactory.get(namespace, persist);
 
     // Decode URL parameters
     const key = decodeKey(params, url.searchParams);
     const cacheTtlParam = url.searchParams.get(PARAM_CACHE_TTL);
-    const cacheTtl =
-      cacheTtlParam === null ? undefined : parseInt(cacheTtlParam);
+    const options: KVGatewayGetOptions = {
+      cacheTtl: cacheTtlParam === null ? undefined : parseInt(cacheTtlParam),
+    };
 
     // Get value from storage
-    const value = await gateway.get(key, { cacheTtl });
+    let value: KVGatewayGetResult | undefined;
+    if (req.headers.get(HEADER_SITES) === null) {
+      const gateway = this.gatewayFactory.get(params.namespace, persist);
+      value = await gateway.get(key, options);
+    } else {
+      // Workers Sites: if this is a sites request, persist should be used as
+      // the root without any additional namespace
+      value = await sitesGatewayGet(persist, key, options);
+    }
     if (value === undefined) throw new KVError(404, "Not Found");
 
     // Return value in runtime-friendly format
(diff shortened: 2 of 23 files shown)


packages/tre/src/shared/types.ts
```diff
@@ -26,3 +26,10 @@ export type Json = Literal | { [key: string]: Json } | Json[];
 export const JsonSchema: z.ZodType<Json> = z.lazy(() =>
   z.union([LiteralSchema, z.array(JsonSchema), z.record(JsonSchema)])
 );
+
+export function maybeMap<From, To>(
+  maybeValue: From | undefined,
+  f: (value: From) => To
+): To | undefined {
+  return maybeValue === undefined ? undefined : f(maybeValue);
+}
```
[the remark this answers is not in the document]
mrbbot: `From` could be any type, not just an array, so may not have a `map` function. This is more like an option/maybe monad bind (e.g. https://doc.rust-lang.org/std/option/enum.Option.html#method.map).

packages/tre/src/storage/file.ts
```diff
@@ -8,6 +8,7 @@ import {
   sanitisePath,
   viewToArray,
 } from "../shared";
+import { NewStorage, createFileStorage } from "../storage2";
```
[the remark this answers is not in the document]
mrbbot: Both storage systems co-exists so we can incrementally migrate the gateways. Once all gateways have been re-implemented using the new storage system, we can remove the old one and rename `NewStorage` to `Storage`.

packages/tre/src/storage2/blob/store.ts
```diff
@@ -0,0 +1,174 @@
...
+  MultipartOptions,
+  MultipartReadableStream,
+  createArrayReadableStream,
+  createFileReadableStream,
+  createFileWritableStream,
+  createMultipartArrayReadableStream,
+  createMultipartFileReadableStream,
+} from "./streams";
+
+const MODE_READ_ONLY = 0o444;
```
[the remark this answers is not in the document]
mrbbot: We could do `fs.constants.S_IRUSR | fs.constants.S_IRGRP | fs.constants.S_IROTH`, but it looks like `S_IRGRP` and `S_IROTH` might not be available on Windows: https://nodejs.org/api/fs.html#file-mode-constants

packages/tre/src/shared/types.ts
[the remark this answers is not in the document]
penalosa: Oh _that_ type of map

packages/tre/src/storage2/storage.ts
```diff
@@ -0,0 +1,26 @@
...
+export function createMemoryStorage(): NewStorage {
+  const db = new Database(":memory:") as TypedDatabase;
+  const blob = new MemoryBlobStore();
+  return { db, blob };
+}
+
+export function createFileStorage(root: string): NewStorage {
+  root = path.resolve(root);
+  fs.mkdirSync(root, { recursive: true });
+  const db = new Database(path.join(root, "db.sqlite")) as TypedDatabase;
```
penalosa, line 22: I feel like it might be nice to give this a more descriptive name? Something like `metadata.db`?

packages/tre/src/storage2/sql.ts
```diff
@@ -0,0 +1,21 @@
+import { Database, Statement } from "better-sqlite3";
+
+// `better-sqlite3`'s `Statement` methods have `any`-typed return values.
```
penalosa, line 3: Can this fix be upstreamed?

packages/tre/src/storage2/keyvalue.ts
```diff
@@ -0,0 +1,255 @@
...
+  InclusiveRange,
+  MultipartOptions,
+  MultipartReadableStream,
+} from "./blob";
+import { TypedDatabase } from "./sql";
+import { NewStorage } from "./storage";
+
+export interface KeyEntry<Metadata = unknown> {
+  key: string;
+  expiration?: number; // milliseconds since unix epoch
```
penalosa, line 14: Why is this not part of the metadata?

packages/tre/src/storage2/keyvalue.ts
```diff
@@ -0,0 +1,255 @@
...
+}
+
+export interface KeyEntriesQuery {
+  prefix?: string;
+  cursor?: string;
+  limit: number;
+}
+export interface KeyEntries<Metadata = unknown> {
+  keys: KeyEntry<Metadata>[];
+  cursor?: string;
```
penalosa, line 32: Does this need a `list_complete`?

packages/tre/src/storage2/keyvalue.ts
```diff
@@ -0,0 +1,255 @@
...
+}
+
+interface Row {
+  key: string;
+  blob_id: string;
+  expiration: number | null; // milliseconds since unix epoch
+  metadata: string | null;
+}
+const SQL_SCHEMA = `
+CREATE TABLE IF NOT EXISTS _mf_entries (
```
penalosa, line 42: Why `_mf_entries`? Would a more descriptive name work here?

packages/tre/src/storage2/keyvalue.ts
```diff
@@ -0,0 +1,255 @@
...
+  key TEXT PRIMARY KEY,
+  blob_id TEXT NOT NULL,
+  expiration INTEGER,
+  metadata TEXT
+);
+CREATE INDEX IF NOT EXISTS _mf_entries_expiration_idx ON _mf_entries(expiration);
+`;
+function sqlStmts(db: TypedDatabase) {
```
penalosa, line 50: I feel like this should be a class, with an instance property for `db`?

packages/tre/src/storage2/keyvalue.ts
```diff
@@ -0,0 +1,255 @@
...
+      {
+        now: number;
+        escaped_prefix: string;
+        start_after: string;
+        limit: number;
+      },
+      Omit<Row, "blob_id">
+    >(
+      `SELECT key, expiration, metadata FROM _mf_entries
+        WHERE key LIKE :escaped_prefix || '%' ESCAPE '\\'
```
penalosa, line 89: Is `LIKE` a prefix match?

packages/tre/src/storage2/keyvalue.ts
```diff
@@ -0,0 +1,255 @@
...
+        AND key > :start_after
+        AND (expiration IS NULL OR expiration >= :now)
+        ORDER BY key LIMIT :limit`
+    ),
+  };
+}
+
+function escapePrefix(prefix: string) {
+  // Prefix all instances of `\`, `_` and `%` with `\`
+  return prefix.replaceAll(/[\\_%]/g, "\\$&");
```
penalosa, line 99: What does `replaceAll` with the `g` flag do? Are both necessary?

packages/tre/src/storage2/keyvalue.ts
```diff
@@ -0,0 +1,255 @@
...
+}
+
+export class KeyValueStorage<Metadata = unknown> {
+  readonly #stmts: ReturnType<typeof sqlStmts>;
+
+  constructor(
+    private readonly storage: NewStorage,
+    private readonly clock = defaultClock
+  ) {
+    storage.db.pragma("case_sensitive_like = TRUE");
```
penalosa, line 117: Should this be defined with the construction of the database itself? Where the WAL is set?

packages/tre/src/storage2/keyvalue.ts
```diff
@@ -0,0 +1,255 @@
...
+    storage.db.exec(SQL_SCHEMA);
+    this.#stmts = sqlStmts(storage.db);
+  }
+
+  #expired(entry: Pick<Row, "expiration">) {
```
penalosa, line 122: `hasExpired`?

packages/tre/src/storage2/keyvalue.ts
```diff
@@ -0,0 +1,255 @@
...
+  ): Promise<KeyValueEntry<Metadata> | null>;
+  get(
+    key: string,
+    ranges: InclusiveRange[],
+    optsFactory: (metadata: Metadata) => MultipartOptions
+  ): Promise<KeyMultipartValueEntry<Metadata> | null>;
+  async get(
+    key: string,
+    ranges?: InclusiveRange | InclusiveRange[],
+    optsFactory?: (metadata: Metadata) => MultipartOptions
```
penalosa, line 147: Why does this have to be a factory function? Could it be static options?

packages/tre/src/storage2/keyvalue.ts
```diff
@@ -0,0 +1,255 @@
+import assert from "assert";
```
penalosa, line 1: I really appreciate the comments in this file!

packages/tre/src/storage2/keyvalue.ts
```diff
@@ -0,0 +1,255 @@
...
+    return entry.expiration !== null && entry.expiration < this.clock();
+  }
+
+  #backgroundDelete(blobId: string) {
+    // Once rows are deleted, or if they failed to insert, we delete the
+    // corresponding blobs in the background, ignoring errors. Blob IDs are
+    // unguessable, so if there aren't any references to them, they can't be
+    // accessed. This means if we fail to delete a blob for any reason, it
+    // doesn't matter, we'll just have a dangling file taking up disk space.
+    void this.storage.blob.delete(blobId).catch(() => {});
```
penalosa, line 132: Is this a _background_ delete? For a background delete, maybe something like `process.nextTick` or `setTimeout`?

packages/tre/src/plugins/kv/gateway.ts
```diff
@@ -47,57 +56,130 @@ function normaliseInt(value: string | number | undefined): number | undefined {
...
+  constructor(maxLength: number, errorFactory: (length: number) => Error) {
+    let length = 0;
+    super({
+      transform(chunk, controller) {
+        length += chunk.byteLength;
+        // If we exceeded the maximum length, don't enqueue the chunk as we'll
+        // be aborting the stream, but don't error just yet, so we get the
+        // correct final length in the error
+        if (length <= maxLength) controller.enqueue(chunk);
+      },
```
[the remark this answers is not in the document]
mrbbot: `transform` is called until the `writable` side of the transform stream is closed (chunks exhausted), at which point `flush` is called once.

packages/tre/src/plugins/kv/gateway.ts
```diff
@@ -47,57 +56,130 @@ function normaliseInt(value: string | number | undefined): number | undefined {
...
   list_complete: boolean;
 }
 
+export function validateGetOptions(
+  key: string,
+  options?: KVGatewayGetOptions
+): void {
+  validateKey(key);
+  // Validate cacheTtl, but ignore it as there's only one "edge location":
+  // the user's computer
```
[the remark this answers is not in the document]
mrbbot: I've so far avoided adding random failures to Miniflare, because I wanted tests using it to be as deterministic as possible, but it would definitely be interesting to explore. Could try simulate things like transient network failures too.

packages/tre/src/storage2/keyvalue.ts
mrbbot: No, `db` will be used for D1 too, where the user has full-access to the database. We don't want to set pragmas that affect query results globally.

packages/tre/src/storage2/keyvalue.ts
mrbbot: For Cache, `contentType` for `MultipartOptions` needs to be extracted from headers stored in the `metadata`.

packages/tre/src/storage2/keyvalue.ts
mrbbot: Fair point, will wrap this in `queueMicrotask`. 👍

packages/tre/src/plugins/kv/sites.ts
```diff
@@ -1,23 +1,61 @@
...
+): AsyncGenerator<string> {
+  const fileEntries = await fs.readdir(currentPath, { withFileTypes: true });
+  for (const fileEntry of fileEntries) {
+    const filePath = path.join(currentPath, fileEntry.name);
+    if (fileEntry.isDirectory()) {
+      yield* walkKeysInner(rootPath, filePath);
+    } else {
+      // Get key name by removing root directory & path separator
+      // (assumes `rootPath` is fully-resolved)
+      yield filePath.substring(rootPath.length + 1);
```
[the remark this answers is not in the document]
mrbbot: Hmmm, I don't think this should be a problem. On both Windows and Unix, the path separator is a single character. FWIW, code like this has been used in Miniflare since version 1.

packages/tre/src/plugins/kv/sites.ts
```diff
@@ -221,3 +255,69 @@ export function getSitesService(options: SitesOptions): Service {
...
+): Promise<KVGatewayGetResult | undefined> {
+  // `persist` is a resolved path set in `getSitesService()`
+  assert(typeof persist === "string");
+
+  validateGetOptions(key, opts);
+  const filePath = path.join(persist, key);
+  if (!filePath.startsWith(persist)) return;
+  try {
+    return { value: await createFileReadableStream(filePath) };
+  } catch (e: any) {
```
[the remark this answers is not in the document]
mrbbot: 😅 Fair, although we're being pretty defensive

packages/tre/src/plugins/kv/sites.ts
```diff
@@ -221,3 +255,69 @@ export function getSitesService(options: SitesOptions): Service {
...
+  assert(typeof persist === "string");
+
+  validateListOptions(opts);
+  const { limit = MAX_LIST_KEYS, prefix, cursor } = opts;
+
+  // Get sorted array of all keys matching prefix
+  const keys: KVGatewayListResult["keys"] = [];
+  for await (const name of walkKeys(persist)) {
+    if (prefix === undefined || name.startsWith(prefix)) keys.push({ name });
+  }
```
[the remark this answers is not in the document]
mrbbot: This is actually handled in `SCRIPT_SITE` above. Potentially we should move it out of there and into these functions though, now that we've got specific site handlers. 🤔

packages/tre/src/plugins/kv/sites.ts
```diff
@@ -221,3 +255,69 @@ export function getSitesService(options: SitesOptions): Service {
...
+  keys.sort((a, b) => lexicographicCompare(a.name, b.name));
```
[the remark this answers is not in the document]
mrbbot: We want lexicographic comparison here instead, they're slightly different (see #235)

packages/tre/src/storage2/storage.ts
mrbbot: Maybe, but for D1 this is going to be the database.

packages/tre/src/storage2/keyvalue.ts
mrbbot: `metadata` is used for KV user `metadata` which we'd like to keep separate from the expiration. We could store something like `{ expiration, metadata }` in `metadata`, but then we'd have to use JSON-aware SQL query operators when purging expired values.

packages/tre/src/storage2/keyvalue.ts
mrbbot: No, we can just rely on the existence of `cursor`.

packages/tre/src/storage2/keyvalue.ts
mrbbot: What were you thinking? This will be shared by both KV and Cache, so I think a general name is ok here.

packages/tre/src/storage2/keyvalue.ts
mrbbot: The nice thing about this being a function is that we get type inference on the return type. This means we can declare the parameter/return types next to the SQL queries. With a class, we'd have to declare them separately, as `db` would have to be passed in the constructor:

```ts
class Statements {
  getByKey: Statement<Pick<Row, "key">, Row>;
  ...

  constructor(db: TypedDatabase) {
    this.getByKey = db.prepare("SELECT ...");
    ...
  }

}
```

packages/tre/src/storage2/keyvalue.ts
mrbbot: It's a wildcard match. `%` matches any string a characters, `_` matches a single character. `||` concatenates strings.

packages/tre/src/storage2/keyvalue.ts
mrbbot: Hmmm, good point. Looks like `prefix.replace(/.../g, ...)` has the same effect, but `prefix.replaceAll(/.../)` throws with `TypeError: String.prototype.replaceAll called with a non-global RegExp argument`.

packages/tre/src/shared/types.ts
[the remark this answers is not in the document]
mrbbot: Renamed to `maybeApply` to make this clearer, 4d302d97d231e45d9124cc3b68d7ead7fd20cb3e

packages/tre/src/plugins/kv/gateway.ts
```diff
@@ -47,57 +56,130 @@ function normaliseInt(value: string | number | undefined): number | undefined {
   }
 }
 
+function createMaxValueSizeError(length: number) {
+  return new KVError(
+    413,
+    `Value length of ${length} exceeds limit of ${MAX_VALUE_SIZE}.`
+  );
+}
```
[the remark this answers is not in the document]
mrbbot: Ah actually, `MaxLengthStream` needs to be passed a error factory function, which `createMaxValueSizeError` currently satisfies. If we changed this to a subclass, we'd need to define a similar function/lambda to construct it.

packages/tre/src/plugins/kv/gateway.ts
```diff
@@ -47,57 +56,130 @@ function normaliseInt(value: string | number | undefined): number | undefined {
...
   prefix?: string;
   cursor?: string;
 }
-export interface KVGatewayListResult<Meta = unknown> {
-  keys: StoredKeyMeta<Meta>[];
+export interface KVGatewayListResult {
+  keys: {
+    name: string;
+    expiration?: number; // seconds since unix epoch
+    metadata?: string; // JSON-stringified metadata
```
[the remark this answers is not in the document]
mrbbot: 92a96d8c0eb96787284eb3de53a65afe04184160

packages/tre/src/plugins/kv/gateway.ts
```diff
@@ -47,57 +56,130 @@ function normaliseInt(value: string | number | undefined): number | undefined {
...
+  }[];
   cursor?: string;
   list_complete: boolean;
```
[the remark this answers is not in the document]
mrbbot: 1f18a2736d49c01b335bb9508757d8a5ef96bdf4

packages/tre/src/plugins/kv/sites.ts
```diff
@@ -1,23 +1,61 @@
...
+    if (fileEntry.isDirectory()) {
+      yield* walkKeysInner(rootPath, filePath);
+    } else {
+      // Get key name by removing root directory & path separator
+      // (assumes `rootPath` is fully-resolved)
+      yield filePath.substring(rootPath.length + 1);
+    }
+  }
+}
+function walkKeys(rootPath: string): AsyncGenerator<string> {
```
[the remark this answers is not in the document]
mrbbot: a5c2c3a25a24e6e3f5bbba254b3246f6ded9b021

packages/tre/src/plugins/kv/router.ts
```diff
@@ -88,9 +101,22 @@ export class KVRouter extends Router<KVGateway> {
...
-    await gateway.put(key, value, { expiration, expirationTtl, metadata });
+    // Put value into storage
+    const value = req.body;
+    assert(value !== null);
+
+    // If we know the value length, avoid passing the body through a transform
+    // stream to count it (trusting `workerd` to send correct value here).
+    // Safety of `!`: `parseInt(null)` is `NaN`
+    const contentLength = parseInt(req.headers.get("Content-Length")!);
+    const valueLengthHint = isNaN(contentLength) ? undefined : contentLength;
```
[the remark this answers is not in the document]
mrbbot: 845adb3702826e34fce341ec9db8e1f94d67a6ed

packages/tre/src/storage2/keyvalue.ts
mrbbot: edb7765c668ccbc52f1cfed893005cde91329cb5

packages/tre/src/storage2/keyvalue.ts
mrbbot: b3ee610b6c94d341344c0430f13d44beb9ec6100

packages/tre/src/storage2/keyvalue.ts
mrbbot: b757e470b866d2185087ec1365355c5bb6d1d521

packages/tre/src/storage2/blob/store.ts
```diff
@@ -0,0 +1,191 @@
...
+    ranges: InclusiveRange[],
+    opts: MultipartOptions
+  ): Promise<MultipartReadableStream | null>;
+
+  put(stream: ReadableStream<Uint8Array>): Promise<BlobId>;
+
+  delete(id: BlobId): Promise<void>;
+}
+
+function generateBlobId(): BlobId {
```
penalosa, line 53: What about just using a UUIDv4? Or a ULID if the time component is important?

packages/tre/src/storage2/blob/store.ts
```diff
@@ -0,0 +1,191 @@
...
+  ): Promise<MultipartReadableStream | null>;
+  async get(
+    id: BlobId,
+    ranges?: InclusiveRange | InclusiveRange[],
+    opts?: MultipartOptions
+  ): Promise<ReadableStream<Uint8Array> | MultipartReadableStream | null> {
+    const blob = this.#blobs.get(id);
+    if (blob === undefined) return null;
+    if (Array.isArray(ranges)) {
+      assert(opts !== undefined);
```
penalosa, line 80: Can this be encoded in the types somehow? Currently they seem to allow `ranges: InclusiveRange, opts: MultipartOptions` and `ranges: InclusiveRange[], opts: undefined`, both of which should be disallowed, I think

packages/tre/src/storage2/blob/store.ts
```diff
@@ -0,0 +1,191 @@
...
+      return createMultipartArrayReadableStream(blob, ranges, opts);
+    } else {
+      return createArrayReadableStream(blob, ranges);
+    }
+  }
+
+  async put(stream: ReadableStream<Uint8Array>): Promise<BlobId> {
+    const id = generateBlobId();
+    const buffer = await arrayBuffer(stream);
+    const blob = new Uint8Array(buffer);
```
penalosa, line 90: I'm confused. Why convert to an arraybuffer and then back to a `Uint8Array`?

packages/tre/src/storage2/blob/store.ts
```diff
@@ -0,0 +1,191 @@
...
+
+    // Store blob, making sure we're storing a new key
+    assert(!this.#blobs.has(id));
```
penalosa, line 93: This could probably do with a message.

packages/tre/src/storage2/blob/store.ts
```diff
@@ -0,0 +1,191 @@
...
+export class FileBlobStore implements BlobStore {
+  readonly #root: string;
+
+  constructor(root: string) {
+    this.#root = path.resolve(root);
+  }
+
+  #idFilePath(id: BlobId) {
+    const filePath = path.join(this.#root, id);
+    return filePath.startsWith(this.#root) ? filePath : null;
```
penalosa, line 113: Is this a verification that `id` doesn't contain `..` or similar? If it is, `filePath` will need [normalised](https://nodejs.org/api/path.html#pathnormalizepath)

packages/tre/src/storage2/blob/store.ts
```diff
@@ -0,0 +1,191 @@
...
+  get(
+    id: BlobId,
+    ranges: InclusiveRange[],
+    opts: MultipartOptions
+  ): Promise<MultipartReadableStream | null>;
+  async get(
+    id: BlobId,
+    ranges?: InclusiveRange | InclusiveRange[],
+    opts?: MultipartOptions
+  ): Promise<ReadableStream<Uint8Array> | MultipartReadableStream | null> {
```
penalosa, line 129: I'm probably misunderstanding these types, but should this be `Promise<ReadableStream<Uint8Array> | null> | MultipartReadableStream>`?

packages/tre/src/storage2/blob/store.ts
```diff
@@ -0,0 +1,191 @@
...
+      }
+    } catch (e: unknown) {
+      if (
+        typeof e === "object" &&
+        e !== null &&
+        "code" in e &&
+        // @ts-expect-error `e.code` should be `unknown`, fixed in TypeScript 4.9
+        e.code === "ENOENT"
+      ) {
+        return null;
```
penalosa, line 150: `unknown`

packages/tre/src/storage2/blob/store.ts
mrbbot: Could do, just wanted something practically guaranteed unique, without adding another dependency.

packages/tre/src/storage2/blob/store.ts
mrbbot: This is encoded in the overloads, but TypeScript doesn't bring those forward to the implementation. Is that ok?

packages/tre/src/storage2/blob/store.ts
mrbbot: We're consuming a stream here. The `stream/consumers` module doesn't provide a `Uint8Array` consumer. `Uint8Array`s are just views over an `ArrayBuffer` though, they don't copy any data. 👍

packages/tre/src/storage2/blob/store.ts
mrbbot: This `id` isn't passed from outside the function, and should be guaranteed unique if `generateBlobId()` is doing its job right. This assertion should really never fail (admittedly you could say that about most assertions 😅). It's essentially here to document the uniqueness constraint.

packages/tre/src/storage2/blob/store.ts
mrbbot: Good catch, presumably `path.resolve` instead of `path.join` would do this?

packages/tre/src/storage2/blob/store.ts
mrbbot: No, this function is `async`, so must return a `Promise<T>`. That `T` will either be:
- A `ReadableStream<Uint8Array>` for a single-part stream
- A `MultipartReadableStream` for a multi-part stream
- `null` if the blob couldn't be found

packages/tre/src/storage2/blob/store.ts
mrbbot: ?

packages/tre/src/storage2/blob/store.ts
```diff
@@ -0,0 +1,191 @@
...
+    // Get path for this ID, this should never be null as blob IDs are encoded
+    const filePath = this.#idFilePath(id);
+    assert(filePath !== null);
+
+    // Write stream to file with exclusive flag to assert new file creation
+    await fs.mkdir(path.dirname(filePath), { recursive: true });
+    const writable = await createFileWritableStream(filePath, true /* excl. */);
+    await stream.pipeTo(writable);
+    // Mark the created file as read-only, this still allows it to be deleted
+    await fs.chmod(filePath, MODE_READ_ONLY);
```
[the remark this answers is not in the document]
mrbbot: Not sure that would work, then we'd be writing to a read-only file? Will double check this on Windows, but pretty sure it made it read-only. 👍

packages/tre/src/storage2/blob/store.ts
mrbbot: Ah, actually `path.join` normalises the result already: https://nodejs.org/api/path.html#pathjoinpaths 👍

packages/tre/src/storage2/blob/store.ts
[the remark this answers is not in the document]
mrbbot: Ah swapping them does work! 😃 Kinda makes sense actually, since I assume file permissions are checked at handle open time. Checked and `MODE_READ_ONLY` creates a read-only file on Windows too. 👍

packages/tre/src/storage2/blob/store.ts
mrbbot: Added add test to cover this case though in 8f750a8fd170173c83f1d8898ba467740939c6d7

packages/tre/src/storage2/blob/store.ts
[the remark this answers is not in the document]
mrbbot: e39ebba3f18df922e818b52a58402965dc45f1d8

packages/tre/src/storage2/blob/store.ts
penalosa: Not remotely blocking for this PR, but I _think_ `crypto.randomUUID()` should be a dependency-free way to get a UUID

packages/tre/src/storage2/blob/store.ts
penalosa: Ah, I entirely misread this!

packages/tre/src/storage2/blob/store.ts
penalosa: Oh interesting, I didn't realise `path.join` did that—TIL!

packages/tre/src/storage2/blob/store.ts
penalosa: Not a clue what this was meaning

packages/tre/src/storage2/index.ts
```diff
@@ -0,0 +1,4 @@
+export * from "./blob";
+export * from "./keyvalue";
+export * from "./sql";
+export * from "./storage";
```
penalosa, line 4: This file exports a lot. Do consumers need access to all the exports, or is just `storage` enough?

penalosa (commented):
This is looking really really good—I'm very excited to see this in Miniflare! Two remaining questions, and then this is good to go from my perspective:
- It would be great to see some tests for this system (as far as I can tell only the blob storage system is tested right now)

packages/tre/src/storage2/index.ts
mrbbot: I think so. `./blob/streams.ts` is used by the Workers Sites gateway for direct file access. `./blob` also needs to be exported so we can import the blob stores in tests from `@miniflare/tre`. `./keyvalue` is used by the KV/Cache gateways. `./sql` exports the `TypedDatabase` type which will be used by the R2 gateway (similar to `sqlStmts` in the abstract key-value store).

penalosa: Sounds good!

penalosa:
> We have tests for the blob stores and the streams. We don't have tests for the SQL stuff since it's more or less better-sqlite3 passed straight through. Were you thinking about tests for the createMemoryStorage/createFileStorage functions and KeyValueStorage classes? I'd argue the existing plugin integration tests are sufficient for those. We have tests for file-system persistence too. What do you think?

I was specifically thinking about `KeyValueStorage `, but I take your point that the gateway tests should be sufficient. I still think there's some value in testing `KeyValueStorage ` directly, but testing the observed behaviour through the gateway should be sufficient.

> In Wrangler, we can change the persistence directories to `./.wrangler/v3/kv` or something, and warn if we detect the old paths.

This seems like a reasonable tradeoff—lets make sure we add this messaging when upgrading to this version of Miniflare in Wrangler

penalosa (approved):

PR #565, opened 2023-05-02
[Miniflare 3] Re-implement R2 gateway and add support for multipart uploads using new storage system

mrbbot:
This PR builds on #555, and re-implements the R2 gateway using the new storage system. This was the last gateway we needed to migrate. 🙂 It also adds support for multipart uploads, similar to #486, but now with transactions provided by SQLite and the streaming support of the new blob store.

This is split into two commits, one for re-implementing what we had already with the new storage system, and one for extending that with multipart support. You may want to review the SQLite schema before everything else: https://github.com/cloudflare/miniflare/blob/d6e29896a240c10a9337fd1b55807f3aaa8a3f3c/packages/tre/src/plugins/r2/schemas.ts#L4-L71

Closes DEVX-592 and DEVX-593


JacobMGEvans (approved):

packages/tre/src/plugins/r2/validator.ts
```diff
@@ -96,17 +101,53 @@ export class Validator {
...
-    if (value.byteLength > MAX_VALUE_SIZE) {
+  range(
+    options: Pick<R2GetOptions, "rangeHeader" | "range">,
+    size: number
+  ): InclusiveRange | undefined {
+    if (options.rangeHeader !== undefined) {
+      const ranges = _parseRanges(options.rangeHeader, size);
+      // If the header contained a single range, use it. Otherwise, if the
+      // header was invalid, or contained multiple ranges, just return the full
+      // response.
```
petebacondarwin, line 122: "just return the full response" is indicated by returning `undefined`?

packages/tre/src/plugins/r2/validator.ts
```diff
@@ -96,17 +101,53 @@ export class Validator {
...
+      if (ranges?.length === 1) {
+        const [start, end] = ranges[0];
+        return { start, end };
+      }
+    } else if (options.range !== undefined) {
+      let { offset, length, suffix } = options.range;
+      // Eliminate suffix is specified
```
petebacondarwin, line 129:
```suggestion
    // Eliminate suffix if specified
```

petebacondarwin: What does eliminate mean here?
Doesn't the logic here mean that if a suffix length is specified you return only the suffix?

packages/tre/src/plugins/r2/router.ts
```diff
@@ -13,23 +14,46 @@ import { R2Gateway, R2Objects } from "./gateway";
...
+  if (chunksLength > metadataSize) {
+    const identity = new TransformStream();
+    const writer = identity.writable.getWriter();
+    // The promise returned by `writer.write()` will only resolve once the chunk
+    // is read, which won't be until after this function returns, so we can't
+    // use `await` here
+    void writer.write(atLeastMetadata.subarray(metadataSize)).then(() => {
+      // Release the writer without closing the stream
+      writer.releaseLock();
+      return body.pipeTo(identity.writable);
```
petebacondarwin, line 57: Do we care about returning here?

packages/tre/src/plugins/r2/gateway.ts
```diff
@@ -42,160 +73,334 @@ export interface R2Objects {
...
+      .condition(meta, options.onlyIf)
+      .range(options, row.size);
+
+    assert(row.blob_id !== null); // TODO(soon): add multipart support
+    const value = await this.#storage.blob.get(row.blob_id, range);
+    if (value === null) throw new NoSuchKey();
+
+    let valueRange: R2Range | undefined;
+    if (value.range !== undefined) {
+      assert(!Array.isArray(value.range));
```
petebacondarwin: Why do we need to assert this? Is it possible? Is that what old ranges looked like?

petebacondarwin: Or is it just for type narrowing?

packages/tre/src/plugins/r2/gateway.ts
```diff
@@ -42,160 +73,334 @@ export interface R2Objects {
...
+
+  constructor(private readonly log: Log, legacyStorage: Storage) {
+    this.#storage = legacyStorage.getNewStorage();
+    this.#storage.db.pragma("case_sensitive_like = TRUE");
+    this.#storage.db.exec(SQL_SCHEMA);
+    this.#stmts = sqlStmts(this.#storage.db);
+  }
+
+  #backgroundDelete(blobId: string) {
+    void this.#storage.blob.delete(blobId).catch(() => {});
```
petebacondarwin: Do nothing at all if a delete errors? No logging - maybe debug?

packages/tre/src/plugins/r2/router.ts
```diff
@@ -77,6 +83,16 @@ function encodeResult(result: R2Object | R2ObjectBody | R2Objects) {
...
+function encodeJSONResult(result: unknown) {
+  const encoded = JSON.stringify(result);
+  return new Response(encoded, {
+    headers: {
+      [CfHeader.MetadataSize]: `${Buffer.byteLength(encoded)}`,
+      "Content-Type": "application/json",
+    },
+  });
+}
+
```
petebacondarwin, line 95: NIT: could this be:

```ts
Response.json(result, { headers: { [CfHeader.MetadataSize]: `${Buffer.byteLength(encoded)}`} })
```

packages/tre/test/shared/sync.spec.ts
```diff
@@ -0,0 +1,87 @@
+import { setTimeout } from "timers/promises";
+import { Mutex, WaitGroup } from "@miniflare/tre";
+import test from "ava";
+
+test("Mutex: runs closures exclusively", async (t) => {
```
petebacondarwin, line 5: No `DeferredPromise` tests?

packages/tre/src/index.ts
```diff
@@ -852,6 +854,17 @@ export class Miniflare {
     return response;
   }
 
+  /** @internal */
+  _getPluginStorage(
```
petebacondarwin, line 858: I couldn't see where this is used, if at all...

packages/tre/src/plugins/r2/gateway.ts
```diff
@@ -28,6 +51,40 @@ import {
...
+// all its parts will be deleted.
+//
+// Note that when completing or aborting an upload, the upload record is NOT
+// deleted. This is because uploads can be aborted more than once, and even
+// aborted after completion (although in this case, aborting is a no-op). We
+// need to be able to distinguish between a completed upload, an aborted upload
+// and an upload that never existed to handle this, and match R2's error
+// messages.
+//
+// If regular `R2Bucket#{put,delete}()` methods are called on completed
```
petebacondarwin, line 83: What about these methods on incomplete uploads?

packages/tre/src/plugins/r2/gateway.ts
```diff
@@ -73,11 +130,33 @@ export interface R2Objects {
 
 const validate = new Validator();
 
+function generateVersion() {
+  return crypto.randomBytes(16).toString("hex");
```
petebacondarwin, line 134: For future, if you want to move this code to a Worker then this API is not available and it is better to use the `crypto.webcrypto.getRandomValues()` api.

packages/tre/src/plugins/r2/gateway.ts
```diff
@@ -112,28 +191,131 @@ function sqlStmts(db: TypedDatabase) {
...
   return {
-    getByKey: db.prepare<Pick<ObjectRow, "key">, ObjectRow>(`
-      SELECT key, blob_id, version, size, etag, uploaded, checksums, http_metadata, custom_metadata
-      FROM _mf_objects WHERE key = :key
-    `),
+    getByKey: stmtGetByKey,
+    getPartsByKey: db.transaction((key: string) => {
+      const row = stmtGetByKey.get({ key });
+      if (row === undefined) return;
+      if (row.blob_id === null) {
```
petebacondarwin, line 278: A comment to say that (I think) having a blob_id indicates it is not multi-part?

petebacondarwin: Or more accurately having `null` as blob_id makes it multipart.

packages/tre/src/plugins/r2/gateway.ts
```diff
@@ -177,22 +359,298 @@ function sqlStmts(db: TypedDatabase) {
...
+        stmtPutPart.run(newRow);
+        return partRow?.blob_id;
+      }
+    ),
+    completeMultipartUpload: db.transaction(
+      (key: string, upload_id: string, selectedParts: R2PublishedPart[]) => {
+        // 1. Check the upload exists and is in-progress
+        const uploadRow = stmtGetUploadMetadata.get({ key, upload_id });
+        if (uploadRow?.state === MultipartUploadState.COMPLETED) {
+          throw new NoSuchUpload();
```
petebacondarwin, line 392: I thought that completing an already completed upload was a noop?

packages/tre/src/plugins/r2/gateway.ts
```diff
@@ -177,22 +359,298 @@ function sqlStmts(db: TypedDatabase) {
...
+        }
+        //    ...then check again in ascending part number order, throwing an
+        //    internal error. We won't know where the current last element ends
+        //    up in the sort, so we just check all parts again.
+        //
+        //    Also check that all but the last parts are the same size...
+        parts.sort((a, b) => a.part_number - b.part_number);
+        let partSize: number | undefined;
+        for (const part of parts.slice(0, -1)) {
+          if (partSize === undefined) partSize = part.size;
```
petebacondarwin, line 443:
```ts
partSize ??= part.size;
```

Maybe?

packages/tre/src/plugins/r2/gateway.ts
```diff
@@ -177,22 +359,298 @@ function sqlStmts(db: TypedDatabase) {
...
+      }
+    ),
+    abortMultipartUpload: db.transaction((key: string, upload_id: string) => {
+      // 1. Make sure this multipart upload exists, ignoring finalised states
+      const uploadRow = stmtGetUploadState.get({ key, upload_id });
+      if (uploadRow === undefined) {
+        throw new InternalError();
+      } else if (uploadRow.state > MultipartUploadState.IN_PROGRESS) {
+        // If this upload has already been finalised, return here. `abort()` can
+        // be called multiple times, and on already `complete()`ed uploads. In
```
petebacondarwin, line 515: I would have imagined that aborting a completed upload would be an error?
But if this is what R2 does then, heh?

packages/tre/src/plugins/r2/gateway.ts
```diff
@@ -177,22 +359,298 @@ function sqlStmts(db: TypedDatabase) {
...
+
+  #releaseBlob(blobId: BlobId) {
+    this.#inUseBlobs.get(blobId)?.done();
+  }
+
+  #backgroundDelete(blobId: BlobId) {
+    this.timers.queueMicrotask(async () => {
+      // Wait for all multipart gets using this blob to complete
+      await this.#inUseBlobs.get(blobId)?.wait();
+      return this.#storage.blob.delete(blobId).catch(() => {});
```
petebacondarwin, line 598: I guess there is no chance of a race-condition here where a new client starts using a blob between the wait group being released and the blob being deleted?

petebacondarwin (approved):
LGTM with a few comments. A lot of the second commit appeared to be just replacing clock with timers, which would have been nice in a separate commit.
Nice work!

packages/tre/src/plugins/r2/validator.ts
mrbbot: Yep, I'll add a comment to make this clearer. 👍

packages/tre/src/plugins/r2/validator.ts
mrbbot: Eliminate in an algebraic sense. We have three variables, `offset`, `length` and `suffix` that are all kinda related, and we'd like to _eliminate_ `suffix` from the equation, so we can just operate on `offset` and `length`.

packages/tre/src/plugins/r2/router.ts
mrbbot: Probably doesn't matter, but I guess it would if we ever `.catch()`ed on the `.then()`.

packages/tre/src/plugins/r2/gateway.ts
mrbbot: Just for type narrowing, R2 doesn't support get requests with multiple ranges, so `workerd` will never send requests for them.

packages/tre/src/plugins/r2/gateway.ts
mrbbot: Added a regular error log. 👍

packages/tre/src/plugins/r2/router.ts
mrbbot: Ah, not in this case, since we need to call `Buffer.byteLength()` on the encoded JSON.

packages/tre/test/shared/sync.spec.ts
mrbbot: Added 👍

packages/tre/src/index.ts
mrbbot: It's used in the R2 tests, to make sure we delete dangling blobs/uploaded parts. Admittedly, this is probably testing the underlying implementation too much, as opposed to user facing behaviour, but I think it's ok.

packages/tre/src/plugins/r2/gateway.ts
mrbbot: They won't delete parts for them, added a comment to clarify 👍

packages/tre/src/plugins/r2/gateway.ts
mrbbot: Added a comment 👍

packages/tre/src/plugins/r2/gateway.ts
mrbbot: Double checked this, this is what R2 does:

1. Completing a completed upload gives `Error: completeMultipartUpload: The specified multipart upload does not exist. (10024)`
2. Aborting a completed upload is a no-op
3. Completing an aborted upload gives `Error: completeMultipartUpload: The specified multipart upload does not exist. (10024)`
4. Aborting an aborted upload is a no-op
5. Completing a non-existent upload gives `Error: completeMultipartUpload: We encountered an internal error. Please try again. (10001)`
6. Aborting a non-existent upload gives `Error: abortMultipartUpload: We encountered an internal error. Please try again. (10001)`

Of these, case 3 was implemented incorrectly, and returned an internal error instead. Just fixed that and updated the test. 👍

packages/tre/src/plugins/r2/gateway.ts
mrbbot: Yeah, me too. Double checked this above. 👍

packages/tre/src/plugins/r2/gateway.ts
mrbbot: I don't think so. We'll only call `#backgroundDelete` once the corresponding SQLite rows referencing the blob have been deleted, and that transaction has been committed. Once this is done, future gets shouldn't be able to reference the blobs anymore, so `#inUseBlobs` wait groups will only be recording tasks from before that delete transaction committed. Therefore, once the `Promise` returned from `wait()` resolves, we know there won't be any more get operations referencing the blob, and we're safe to delete it.

petebacondarwin (approved):

packages/tre/src/plugins/r2/router.ts
petebacondarwin: Ah!

PR #566, opened 2023-05-04
[Miniflare 3] Implement Queues

mrbbot:
This PR adds support for [Queues](https://developers.cloudflare.com/queues/) to Miniflare 3. It builds on https://github.com/cloudflare/workerd/pull/541, https://github.com/cloudflare/workerd/pull/543, and https://github.com/cloudflare/workerd/pull/545 (thanks @a-robinson!).

https://github.com/cloudflare/workerd/pull/545 adds a `queue()` method on service bindings to dispatch `queue` events to Workers. It expects regular, de-serialised JavaScript objects, but the Queue producer bindings from https://github.com/cloudflare/workerd/pull/541 send V8-serialised objects instead. Unfortunately, `workerd` doesn't expose the V8 (de)serialiser in Workers yet, so we need to re-serialise all messages into a format we can deserialise in a Worker. We use [`devalue`](https://github.com/Rich-Harris/devalue) for this. This is the first time we're using an `npm` dependency in an internal Miniflare Worker. To support this, and because it's something we've been wanting to do for a while, the core entrypoint Worker has been moved to a separate TypeScript file, which is built as an additional entrypoint with `esbuild`, and type-checked under `@cloudflare/workers-types`. This is a separate commit to aid reviewing. We'll migrate other internal Workers to this format soon.

Another minor issue is that `workerd` uses V8 serialisation version 15 [when sending messages](https://github.com/cloudflare/workerd/blob/575eba6747054fb810f8a8138c2bf04b22339f77/src/workerd/api/queue.c%2B%2B#L17). This is only supported by V8 [versions 10.0.29 and above](https://github.com/v8/v8/commit/fc23bc1de29f415f5e3bc080055b67fb3ea19c53). For reference, the V8 versions associated with notable Node versions are:
- Miniflare's minimum supported version: Node 16.13.0 --> V8 9.4
- Last Node 17/unsupported version: Node 17.9.1  --> V8 9.6
- First supported version: Node 18.0.0  --> V8 10.1

This means that Queues will only be supported in Miniflare 3 if you're using Node 18 or above. I think this is ok, as I'm pretty sure Queues is still in beta, and it was previously [marked experimental in Miniflare 2](https://github.com/cloudflare/miniflare/blob/1a2cc368bfbe4d117e781ed4e74309c8da0c10d1/packages/queues/src/plugin.ts#L89-L91).

---

I've added some other questions that would be good to get opinions on too. 🙂


packages/tre/src/plugins/queues/index.ts
```diff
@@ -0,0 +1,92 @@
...
+import { QueuesGateway } from "./gateway";
+import { QueuesRouter } from "./router";
+
+export const QueuesOptionsSchema = z.object({
+  queueProducers: z
+    .union([z.record(z.string()), z.string().array()])
+    .optional(),
+  queueConsumers: z
+    .union([z.record(QueueConsumerOptionsSchema), z.string().array()])
+    .optional(),
```
mrbbot, line 21: This is a slightly different API to Miniflare 2...

https://github.com/cloudflare/miniflare/blob/e3673a5df09c98e674932d01ec5c14b3ce5becec/packages/queues/src/plugin.ts#L16-L32

...but I think it more closely matches the rest of Miniflare's options. Queues was previously marked as experimental, so I think this change is ok, but open to suggestions.

packages/tre/src/plugins/queues/index.ts
```diff
@@ -0,0 +1,92 @@
...
+// - Miniflare's minimum supported version: Node 16.13.0 --> V8 9.4
+// - Last Node 17/unsupported version:      Node 17.9.1  --> V8 9.6
+// - First supported version:               Node 18.0.0  --> V8 10.1
+//
+// See also https://github.com/nodejs/node/issues/42192.
+/** @internal */
+export const _QUEUES_COMPATIBLE_V8_VERSION =
+  semiver(process.versions.v8, "10.0.29") >= 0;
+
+function assertCompatibleV8Version() {
```
mrbbot, line 39: Ideally, we'd just bump Miniflare 3's minimum supported Node version to 18, but I'm not sure if we want to do that for Wrangler 3 too?

packages/tre/src/workers/core/devalue.ts
```diff
@@ -0,0 +1,99 @@
...
+  },
+  Error(value) {
+    for (const ctor of ALLOWED_ERROR_CONSTRUCTORS) {
+      if (value instanceof ctor && value.name === ctor.name) {
+        return [value.name, value.message, value.stack, value.cause];
+      }
+    }
+    if (value instanceof Error) {
+      return ["Error", value.message, value.stack, value.cause];
+    }
```
mrbbot, line 58: This is kinda similar to the `reviveError` function we use for the pretty error page. Was thinking whether we could reuse this function there, but don't think including `devalue` in the JSON error middleware is worth it.

packages/tre/src/plugins/queues/gateway.ts
```diff
@@ -0,0 +1,250 @@
...
+  #pendingFlush?: PendingFlush;
+
+  constructor(
+    private readonly log: Log,
+    _storage: Storage,
+    private readonly timers: Timers,
+    private readonly queueName: string,
+    private readonly dispatchFetch: DispatchFetch
+  ) {
+    this.#queueUrl = new URL(`http://localhost/${queueName}`);
```
penalosa, line 104: Is this _actually_ `localhost:80`? If not, what about something like `http://dummy` for clarity?

packages/tre/src/plugins/queues/gateway.ts
```diff
@@ -0,0 +1,250 @@
...
+
+  async #dispatchBatch(workerName: string, batch: Message[]) {
+    // The `queue()` method on a service binding expects regular, de-serialised
+    // JavaScript objects. Unfortunately, `workerd` doesn't expose the V8
+    // serialiser in Workers yet, so we need to re-serialise all messages into a
+    // format we can deserialise in a Worker.
+    // TODO: stop re-serialising messages once `v8` module added to `workerd` as
+    //  part of Node.js compat work, can also remove V8 version restriction too
+    // eslint-disable-next-line es/no-dynamic-import
+    const { stringify } = await import("devalue");
```
penalosa, line 115: Why does this have to be a dynamic import?

packages/tre/src/plugins/queues/gateway.ts
```diff
@@ -0,0 +1,250 @@
...
+    const jsonMessages = batch.map((message) => {
+      const body = v8.deserialize(message.serialisedBody);
+      const serialisedBody = stringify(body, structuredSerializableReducers);
```
penalosa, line 118: Presumably not, but why can't you do 
```ts
const serialisedBody = stringify({ id: message.id, timestamp: message.timestamp, body }, structuredSerializableReducers);
```

packages/tre/src/plugins/queues/index.ts
penalosa: If it's required for Queues...

packages/tre/src/plugins/shared/index.ts
```diff
@@ -4,8 +4,29 @@ import { Awaitable, Log, OptionalZodTypeOf } from "../../shared";
...
+  // https://developers.cloudflare.com/queues/platform/limits/
+  maxBatchSize: z.number().min(0).max(100).optional(),
+  maxBatchTimeout: z.number().min(0).max(30).optional(), // seconds
+  maxRetires: z.number().min(0).max(100).optional(),
+  deadLetterQueue: z.ostring(),
+});
+export type QueueConsumerOptions = z.infer<typeof QueueConsumerOptionsSchema>;
+export interface QueueConsumer extends QueueConsumerOptions {
+  workerName: string;
+  deadLetterConsumer?: QueueConsumer;
```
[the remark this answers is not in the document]
penalosa: This seems like it allows infinite recursion through `deadLetterConsumer`'s? Does Queues have a limit?

packages/tre/src/runtime/config/workerd.capnp
```diff
@@ -325,6 +329,14 @@ struct Worker {
       # R2 bucket and admin API bindings. Similar to KV namespaces, these turn operations into
       # HTTP requests aimed at the named service.
 
+      wrapped @14 :WrappedBinding;
```
penalosa, line 332: Can this file be retrieved from `workerd` somehow? in order to keep it in sync?

packages/tre/src/runtime/config/workerd.capnp.d.ts
```diff
@@ -12,6 +12,7 @@ export declare class Config extends __S {
     };
     static _Services: capnp.ListCtor<Service>;
     static _Sockets: capnp.ListCtor<Socket>;
+    static _Extensions: capnp.ListCtor<Extension>;
```
penalosa, line 15: I kinda feel like this file shouldn't be committed. Could it be generated as part of the `dev`/`build` process so that it's always in sync with `workerd.capnp`?

packages/tre/src/runtime/config/workerd.capnp.js
```diff
@@ -27,10 +27,16 @@ class Config extends capnp_ts_1.Struct {
     hasV8Flags() { return !capnp_ts_1.Struct.isNull(capnp_ts_1.Struct.getPointer(2, this)); }
     initV8Flags(length) { return capnp_ts_1.Struct.initList(2, capnp.TextList, length, this); }
     setV8Flags(value) { capnp_ts_1.Struct.copyFrom(value, capnp_ts_1.Struct.getPointer(2, this)); }
+    adoptExtensions(value) { capnp_ts_1.Struct.adopt(value, capnp_ts_1.Struct.getPointer(3, this)); }
```
penalosa, line 30: Same comment as in sibling files

packages/tre/src/workers/core/devalue.ts
penalosa: How big is it? What's the bundle size impact?

packages/tre/src/workers/core/entry.worker.ts
```diff
@@ -0,0 +1,205 @@
+// @ts-expect-error "devalue" is ESM-only, so TypeScript requires us to use the
+//  `*.mts` extension, or set `"type": "module"` in our `package.json`. We can't
+//  do the first as ambient types from `@cloudflare/workers-types` don't seem to
+//  work, and the second would affect all consumers of `miniflare`.
+import { unflatten } from "devalue";
```
penalosa, line 5: https://devblogs.microsoft.com/typescript/announcing-typescript-5-0-beta/#moduleresolution-bundler should fix this if the Typescript version can be upgraded to 5?

packages/tre/test/plugins/queues/index.spec.ts
```diff
@@ -0,0 +1,628 @@
...
+  t.is(batches.length, 2);
+  // (third batch isn't full, so need to wait for max batch timeout)
+  timers.timestamp += 1000;
+  await timers.waitForTasks();
+  t.is(batches.length, 3);
+  t.is(batches[0]?.length, 5);
+  t.is(batches[1]?.length, 5);
+  t.is(batches[2]?.length, 1);
+  batches = [];
+});
```
penalosa, line 157: Is there a reason non of these tests check message values?

penalosa (commented):
Looking good! How does this work with Wrangler? i.e. if we don't have both workers available as source

packages/tre/src/workers/core/devalue.ts
mrbbot: Pretty small, `stringify` is ~4KB. Tbh it's not that much duplicate code, so maybe we just leave it?

packages/tre/src/plugins/core/errors/index.ts
```diff
@@ -192,7 +192,6 @@ function getSourceMappedStack(workerSrcOpts: SourceOptions[], error: Error) {
 // As a stop-gap solution, user workers should send an HTTP 500 JSON response
 // matching the schema below with the `MF-Experimental-Error-Stack` header set
 // to a truthy value, in order to display the pretty-error page.
```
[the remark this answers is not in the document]
mrbbot: The constant has moved to `CoreHeaders` in `src/workers/core/constants.ts`. I think this comment is still ok though, since it refers to the header by value, not variable name.

packages/tre/src/plugins/core/index.ts
```diff
@@ -132,116 +116,12 @@ const LIVE_RELOAD_SCRIPT_TEMPLATE = (
...
-  } catch (e) {
-    // TODO: return pretty-error page here
-    return new Response(e.stack);
-  }
-}
-addEventListener("fetch", (event) => {
-  event.respondWith(handleEvent(event));
-});
-`;
+const SCRIPT_ENTRY = readWorker("core", "entry");
```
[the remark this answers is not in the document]
mrbbot: Ooooh this is a great idea! 3364a03d07e925e9eb4499dd39213d3ef8189bbc

packages/tre/src/plugins/core/index.ts
```diff
@@ -475,8 +355,9 @@ export function getGlobalServices({
     {
       name: SERVICE_ENTRY,
       worker: {
-        serviceWorkerScript: SCRIPT_ENTRY,
-        compatibilityDate: "2022-09-01",
+        modules: [SCRIPT_ENTRY],
+        compatibilityDate: "2023-04-04",
+        compatibilityFlags: ["nodejs_compat", "service_binding_extra_handlers"],
```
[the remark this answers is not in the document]
mrbbot: `src/workers/core/devalue.ts` imports `Buffer`, for decoding `base64`, and `assert` for validating types at runtime. We could probably write our own base 64 functions, but if `nodejs_compat` is available, we might as well use it. This is only for the entrypoint worker too, it doesn't affect regular user workers. 👍

packages/tre/src/workers/core/constants.ts
```diff
@@ -0,0 +1,29 @@
...
+  SERVICE_LOOPBACK: "MINIFLARE_LOOPBACK",
+  JSON_VERSION: "MINIFLARE_VERSION",
+  SERVICE_USER_ROUTE_PREFIX: "MINIFLARE_USER_ROUTE_",
+  SERVICE_USER_FALLBACK: "MINIFLARE_USER_FALLBACK",
+  TEXT_CUSTOM_SERVICE: "MINIFLARE_CUSTOM_SERVICE",
+  JSON_CF_BLOB: "CF_BLOB",
+  JSON_ROUTES: "MINIFLARE_ROUTES",
+  JSON_LOG_LEVEL: "MINIFLARE_LOG_LEVEL",
+  DATA_LIVE_RELOAD_SCRIPT: "MINIFLARE_LIVE_RELOAD_SCRIPT",
+} as const;
```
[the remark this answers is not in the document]
mrbbot: If they were in the core plugin, they'd get typed under `@types/node`, and any file that imported them (e.g. a worker script) would inherit the `@types/node` environment. This is incompatible with `@cloudflare/workers-types`. Therefore, we put them somewhere where they're typed under a neutral, non-Node, non-Workers environment. Note, only `*.worker.ts` files are typed under `@cloudflare/workers-types`.

packages/tre/src/plugins/shared/constants.ts
```diff
@@ -13,17 +14,17 @@ export const HEADER_PERSIST = "MF-Persist";
...
 
 export const WORKER_BINDING_SERVICE_LOOPBACK: Worker_Binding = {
-  name: BINDING_SERVICE_LOOPBACK,
+  name: CoreBindings.SERVICE_LOOPBACK,
   service: { name: SERVICE_LOOPBACK },
 };
 
 // TODO: make this an inherited worker in core plugin
+// TODO: do we need to URI encode namespace?
 const SCRIPT_PLUGIN_NAMESPACE_PERSIST_COMPAT_DATE = "2022-09-01";
```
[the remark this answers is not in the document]
mrbbot: Could do, I'd rather the compatibility dates were colocated with the script they control though.

packages/tre/src/plugins/shared/constants.ts
```diff
@@ -13,17 +14,17 @@ export const HEADER_PERSIST = "MF-Persist";
...
 export const BINDING_TEXT_PERSIST = "MINIFLARE_PERSIST";
 
 export const WORKER_BINDING_SERVICE_LOOPBACK: Worker_Binding = {
-  name: BINDING_SERVICE_LOOPBACK,
+  name: CoreBindings.SERVICE_LOOPBACK,
   service: { name: SERVICE_LOOPBACK },
 };
 
 // TODO: make this an inherited worker in core plugin
+// TODO: do we need to URI encode namespace?
```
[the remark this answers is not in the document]
mrbbot: Yep, it is user settable. Fixed and added tests for namespaces with potentially unsafe characters. 1299c88fc98e8778f28f542391fceb88c24d23c3

packages/tre/src/plugins/shared/gateway.ts
```diff
@@ -22,6 +23,11 @@ import {
 export const PersistenceSchema = z.boolean().or(z.string()).optional();
 export type Persistence = z.infer<typeof PersistenceSchema>;
 
+export type DispatchFetch = (
+  input: RequestInfo,
+  init?: RequestInit<Partial<IncomingRequestCfProperties>>
```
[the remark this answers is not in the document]
mrbbot: `IncomingRequestCfProperties` marks some properties as required, but when calling `dispatchFetch()`, we don't want users to have to specify them all. They may not be depending on all of them, or just using custom `cf` properties.

packages/tre/src/plugins/shared/index.ts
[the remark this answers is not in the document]
mrbbot: Queues does not have a limit. 🙃 It doesn't let you define a dead letter consumer as the same queue, but it does let you do infinite cycles with multiple queues, so this is the _correct_ behaviour.

packages/tre/src/plugins/queues/errors.ts
```diff
@@ -0,0 +1,9 @@
+import { HttpError, MiniflareError } from "../../shared";
+
+export type QueuesErrorCode =
+  | "ERR_V8_UNSUPPORTED" // V8 version too old
```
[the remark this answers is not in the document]
mrbbot: By the `assertCompatibleV8Version()` function in `src/plugins/queues/index.ts`.

packages/tre/src/plugins/queues/gateway.ts
mrbbot: Sure, fixed in 1af5499c8c81cceb09fd418b6695c83a0e890960. 🙂

packages/tre/src/plugins/queues/gateway.ts
mrbbot: `devalue` is ESM-only, but we bundle Miniflare as CommonJS. 🙁

packages/tre/src/plugins/queues/gateway.ts
mrbbot: That is an excellent point, we could totally do that. We can actually simplify this further, and just stringify everything at once. 🙂 3dab2a2e9dd211f3fc24e630ec72cc02595e2e1d

packages/tre/src/runtime/config/workerd.capnp
mrbbot: Yes, we were thinking about including it in the `workerd` npm package right?

packages/tre/src/runtime/config/workerd.capnp.d.ts
mrbbot: It should be generated with `dev`/`build`, but it's a bit of a pain to. You need to downgrade to some old TypeScript version supported by `capnpc-ts` then run the `capnp` tool.

packages/tre/src/workers/core/entry.worker.ts
mrbbot: Oh nice! Doesn't look like `api-extractor` supports TypeScript 5 yet, but we should definitely switch to this when we upgrade. Alternatively, we just scrap API extractor and ship multiple `*.d.ts` files for our single bundled `*.js` file?

packages/tre/test/plugins/queues/index.spec.ts
mrbbot: Not really, we have tests that check message values later on. We're more interested in the batching behaviour here. We could check message ordering, but again, this is covered by the retry test later on.

packages/tre/src/plugins/queues/gateway.ts
mrbbot: Actually, we could solve this by marking `devalue` as a `devDependency` and getting `esbuild` to bundle it in.

mrbbot: 138f3a4abe10e4a56f2da2075d7e320b19fd01c7

penalosa:
No, I was wondering how this works if Wrangler only has access to one user Worker—do the queue producer and consumer have to be defined in the same script?

packages/tre/src/plugins/shared/gateway.ts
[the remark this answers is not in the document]
penalosa: Oh :facepalm:—this is setting _incoming_ properties. What about potentially merging the provided cf object with the miniflare default one? That way Workers relying on cf properties can have confidence they're always there (and if you just want to customise a single cf property you don't need to redefine the whole thing)?

packages/tre/src/runtime/config/workerd.capnp
penalosa: Yeah, I think so. Not a blocker for this PR obviously, but might be nice as a fast follow.

packages/tre/src/runtime/config/workerd.capnp.d.ts
penalosa: Fair... We should investigate if there's a better way to do this in the near future

packages/tre/src/workers/core/devalue.ts
penalosa: Yeah, it's very small—not a big deal to keep it duplicated

packages/tre/src/workers/core/entry.worker.ts
penalosa: What would that look like? What's the benefit of api extractor over that approach?

packages/tre/src/plugins/shared/gateway.ts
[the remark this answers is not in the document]
mrbbot: Good idea, 244e3f51cb767d7729ef96f315c16a9fac8ee957

packages/tre/src/workers/core/entry.worker.ts
mrbbot: `api-extractor` just means we have a single `index.d.ts` file corresponding to our `index.js` bundle. Without it, the structure of our `src` folder would make it to the final bundle as TypeScript definitions file. Essentially, I think you'd be able to do stuff like `import { ... } from "@miniflare/tre/plugins/core"` as `plugins/core/index.d.ts` would exist, but the code for it would be in the `index.js` bundle.

PR #569, opened 2023-05-10
[Miniflare 3] Ignore `node:` and `cloudflare:` modules when collecting modules

mrbbot:
Previously, Miniflare would complain when specifying a `script`/ `scriptPath` that used `node:`/`cloudflare:` imports with `modules: true` set. This change updates Miniflare's module collector to ignore them, as they're automatically included by `workerd`.


packages/tre/test/index.spec.ts
```diff
@@ -234,3 +234,19 @@ test("Miniflare: custom service binding to another Miniflare instance", async (t
     body: null,
   });
 });
+
+test("Miniflare: can use `node:` modules", async (t) => {
+  const mf = new Miniflare({
+    port: await getPort(),
+    modules: true,
+    compatibilityFlags: ["nodejs_compat"],
+    script: `import { Buffer } from "node:buffer";
```
RamIdeas, line 243: Is there a "cloudflare:" import you can also add to this test script?

RamIdeas: Answer: not available to import yet, but will be soon

LGTM

RamIdeas (approved):

PR #571, opened 2023-05-10
[Miniflare 3] Remove legacy storage system

mrbbot:
Now that all gateways have been migrated to the new storage system, we aren't using the legacy one. Also removes remote storage, which will need rethinking with the new storage system.

Split into 2 commits: the first removes the old storage system, then the second renames the new storage system to the old names.

_(to be merged after #566)_


penalosa (approved):
🎉

petebacondarwin:
Windows is not happy...

PR #572, opened 2023-05-10
[Miniflare 3] Fix source mapping of service workers

mrbbot:
cloudflare/workerd#190 changed the script name for service workers to their service name. This change updates Miniflare's heuristics for locating source maps to work with this. It also adds some tests to ensure source mapping is working correctly for the different ways of specifying a script.

_(to be merged after #566)_

### DIFF
--- a/packages/tre/src/index.ts
+++ b/packages/tre/src/index.ts
@@ -43,7 +43,7 @@ import {
 } from "./plugins";
 import {
   JsonErrorSchema,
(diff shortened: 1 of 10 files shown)


penalosa (approved):
All looks great! I'm not sure about the tests—I think it would be useful to check that the passed message is actually present in the error stack, since right now it doesn't look like that's being tested.

PR #606, opened 2023-06-14
Fix nested waitUntil

### LINKED ISSUE #605: Nested `waitUntil` not awaited
When calling `waitUntil` inside another `waitUntil`, the nested `waitUntil` should also be awaited. This does not happen in Miniflare v2.

Minimal repro:

```ts
import { Miniflare } from "miniflare";

const mf = new Miniflare({
  script: `
  const delay = (n) => new Promise((res) => setTimeout(res, n))

  export default {
    async fetch(req, env, ctx) {
      const fn = async () => {
        await delay(2000);

        ctx.waitUntil(delay(2000).then(()=>{
          console.log("nested waitUntil promise resolved")
        }))

        console.log("fn resolved")
      };

      ctx.waitUntil(fn())

      return new Response("ok", { status: 200 })
    }
  }`,
  modules: true,
  compatibilityDate: "2023-03-30"
});


const res = await mf.dispatchFetch("https://test", { method: "GET" });
console.log("received response")
await res.waitUntil()
console.log("wait until done")
```

`nested waitUntil promise resolved` is never logged.

hansottowirtz:
Fixes #605

- I've tested that this behavior also occurs in actual Workers (see screenshot)
- I've added a test
- The values in `await res.waitUntil()` are in the order they have been `waitUntil`'ed
- `@miniflare/core` now exports `waitUntilAll`, which handles awaiting arrays that change in length

Screenshot 2023-06-14 at 11 26 33


packages/core/test/index.spec.ts
```diff
@@ -1384,3 +1384,37 @@ test("MiniflareCore: dispose: cleans up watcher", async (t) => {
...
+  const mf = useMiniflare(
+    {},
+    {
+      script: `
+        const delay = (n) => new Promise((res) => setTimeout(res, n))
+
+        export default {
+          async fetch(req, env, ctx) {
+            const fn = async () => {
+              await delay(2000);
```
mrbbot, line 1397: I believe `delay(0)` should still be enough to fail this test without the fix. Could you try that, or a much smaller value here, so this test doesn't take too long?

packages/core/src/wait-until-all.ts
```diff
@@ -0,0 +1,14 @@
+/** This function is like Promise.all, but also handles nested changes to the promises array */
+export async function waitUntilAll<WaitUntil extends any[] = unknown[]>(
```
mrbbot, line 2: I'd probably just stick this function in `src/standards/event.ts`. I don't think it needs its own file. 👍

packages/core/test/index.spec.ts
```diff
@@ -1384,3 +1384,37 @@ test("MiniflareCore: dispose: cleans up watcher", async (t) => {
...
+            return new Response("ok", { status: 200 })
+          }
+        }
+      `,
+      modules: true,
+    },
+    log
+  );
+
+  const res = await mf.dispatchFetch("https://test", { method: "GET" });
```
mrbbot, line 1417: Could you add tests for the other event types here too?

.changeset/wet-taxis-cough.md
mrbbot, line 1: Despite what the changeset bot says, Miniflare doesn't actually use changesets for its releases. You shouldn't need to add a changeset.

packages/core/test/index.spec.ts
```diff
@@ -1384,3 +1384,37 @@ test("MiniflareCore: dispose: cleans up watcher", async (t) => {
...
+        const delay = (n) => new Promise((res) => setTimeout(res, n))
+
+        export default {
+          async fetch(req, env, ctx) {
+            const fn = async () => {
+              await delay(2000);
+
+              ctx.waitUntil(delay(2000).then(()=>{
+                console.log("nested waitUntil promise resolved")
+              }))
```
mrbbot, line 1401: Instead of logging, could you return a number here, then assert on the contents of `await res.waitUntil()` later in the test?

packages/core/test/index.spec.ts
mrbbot, line 1417: Nit: the method doesn't really matter here, and defaults to `GET`:
```suggestion
  const res = await mf.dispatchFetch("https://test");
```

mrbbot (commented):
Hey! 👋 Thanks so much for fixing this and including a test! 😃 I've added a few suggestions.

PR #610, opened 2023-06-18
Allow `setOptions()` to set runtime port

GregBrimble:
No idea if this is even vaguely right, but here's a first stab at fixing #608.

### DIFF
--- a/packages/miniflare/src/index.ts
+++ b/packages/miniflare/src/index.ts
@@ -455,6 +455,17 @@ export class Miniflare {
     // noinspection JSObjectNullOrUndefined
     this.#loopbackPort = address.port;
 
+    await this.#startRuntime();
+
+    // Update config and wait for runtime to start
+    await this.#assembleAndUpdateConfig(/* initial */ true);
+  }
+
+  async #startRuntime() {
+    await this.#runtime?.dispose();
+
+    assert(this.#loopbackPort !== undefined);
+
     // Start runtime
     const port = this.#sharedOpts.core.port ?? 0;
     const opts: RuntimeOptions = {
@@ -466,9 +477,6 @@ export class Miniflare {
     };
     this.#runtime = new Runtime(opts);
     this.#removeRuntimeExitHook = exitHook(() => void this.#runtime?.dispose());
-
-    // Update config and wait for runtime to start
-    await this.#assembleAndUpdateConfig(/* initial */ true);
   }
 
   async #handleLoopbackCustomService(
@@ -847,6 +855,8 @@ export class Miniflare {
     this.#sharedOpts = sharedOpts;
     this.#workerOpts = workerOpts;
     this.#log = this.#sharedOpts.core.log ?? this.#log;
+
+    await this.#startRuntime();
 
     // Send to runtime and wait for updates to process
     await this.#assembleAndUpdateConfig();
--- a/packages/miniflare/src/index.ts
+++ b/packages/miniflare/src/index.ts
@@ -69,6 +69,7 @@ import {
   Timers,
   defaultTimers,
   formatResponse,
+  maybeApply,
 } from "./shared";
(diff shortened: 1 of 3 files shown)


packages/miniflare/src/runtime/index.ts
```diff
@@ -98,18 +101,24 @@ export class Runtime {
...
   }
 
   async updateConfig(
     configBuffer: Buffer,
-    options?: Abortable
+    options?: Abortable,
+    {
+      entryHost = this.opts.entryHost,
+      entryPort = this.opts.entryPort,
+    }: Partial<Pick<RuntimeOptions, "entryHost" | "entryPort">> = {}
```
mrbbot, line 113: Could we merge these options objects?
```suggestion
    opts?: Abortable & Partial<Pick<RuntimeOptions, "entryHost" | "entryPort">>
```

packages/miniflare/src/runtime/index.ts
```diff
@@ -98,18 +101,24 @@ export class Runtime {
...
-    this.#args = args;
+    return args;
   }
 
   async updateConfig(
     configBuffer: Buffer,
-    options?: Abortable
+    options?: Abortable,
+    {
+      entryHost = this.opts.entryHost,
```
mrbbot, line 111: I don't think we need to worry about `entryHost` here, this comes from `#host` which is assigned a fixed value in the `Miniflare` `constructor`: https://github.com/cloudflare/miniflare/blob/cfddac3e0661f3e327a5a2f4b7256513fb399df9/packages/miniflare/src/index.ts#L372
```suggestion
```

packages/miniflare/src/index.ts
```diff
@@ -785,9 +785,18 @@ export class Miniflare {
...
     const configBuffer = serializeConfig(config);
-    const maybePort = await this.#runtime.updateConfig(configBuffer, {
-      signal: this.#disposeController.signal,
-    });
+    const maybePort = await this.#runtime.updateConfig(
+      configBuffer,
+      {
+        signal: this.#disposeController.signal,
+      },
+      !initial && this.#runtimeEntryURL !== undefined
```
mrbbot, line 793: We should only need to check `this.#runtimeEntryURL !== undefined` here. _(In hindsight, the `initial` parameter is kinda redundant, since `this.#runtimeEntryURL` should always be undefined when `initial` is `true`)_.

packages/miniflare/src/index.ts
```diff
@@ -785,9 +785,18 @@ export class Miniflare {
...
+        ? {
+            entryHost: this.#runtimeEntryURL.host,
+            entryPort: parseInt(this.#runtimeEntryURL.port),
+          }
+        : {}
```
mrbbot, line 798: Should be able to simplify this using the `maybeApply` helper.
```suggestion
      {
        signal: this.#disposeController.signal,
        entryPort: maybeApply(parseInt, this.#runtimeEntryURL?.port),
      },
```

packages/miniflare/src/runtime/index.ts
```diff
@@ -98,18 +101,21 @@ export class Runtime {
...
   async updateConfig(
     configBuffer: Buffer,
-    options?: Abortable
+    options?: Abortable & Partial<Pick<RuntimeOptions, "entryPort">>
   ): Promise<number | undefined> {
     // 1. Stop existing process (if any) and wait for exit
     await this.dispose();
     // TODO: what happens if runtime crashes?
 
+    if (options?.entryPort) {
```
mrbbot, line 115: Passing `0` here should probably be accepted.
```suggestion
    if (options?.entryPort !== undefined) {
```

mrbbot (approved):
One tiny thing, then looks good to me! ✅

mrbbot (approved):

PR #612, opened 2023-06-20
Add `https` option to have miniflare accept https requests

jspspike:


packages/miniflare/src/index.ts
```diff
@@ -716,13 +717,37 @@ export class Miniflare {
...
 
+    let httpOptions: any = {
+      http: {
+        // Even though we inject a `cf` object in the entry worker, allow it to
+        // be customised via `dispatchFetch`
+        cfBlobHeader: HEADER_CF_BLOB,
+      },
+    };
+
+    if (sharedOpts.core.https) {
```
mrbbot, line 728: To match Miniflare 2's behaviour, HTTPS should be enabled if any of the `https*` options are specified (i.e. if you specify `httpsKey`, you shouldn't need to specify `https: true`).

packages/miniflare/src/index.ts
mrbbot, line 1: `dispatchFetch()` will need updating to accept the configured key/pair. See https://undici.nodejs.org/#/docs/best-practices/client-certificate for details on how to do this with `undici`. The tricky thing here is that we'll also need to pass the certificate configuration when establishing WebSocket connections.

I'd do this by creating an `undici` `Agent` dispatcher that allows unauthorised connections:

```ts
export const allowUnauthorizedAgent = new Agent({
  connect: { rejectUnauthorized: false },
});
```

`undici` allows you to pass a custom dispatcher when calling `fetch`. In `dispatchFetch()`, if the protocol was `https:`, you could use this dispatcher instead of the default.

```ts
const forwardInit = forward as RequestInit;
if (url.protocol === "https:") {
  forwardInit.dispatcher = allowUnauthorizedAgent;
}
const response = await fetch(url, forwardInit);
```

Then in `src/http/fetch.ts`, you could check if the dispatcher matched this constant agent, and pass the required options to the WebSocket constructor.

```ts
const requestInit = init as RequestInit;
let secureOpts: NodeWebSocket.ClientOptions | undefined;
if (requestInit.dispatcher === allowUnauthorizedAgent) {
  secureOpts = { rejectUnauthorized: false };
}
const ws = new NodeWebSocket(url, protocols, {
  ...,
  ...secureOpts,
});
```

You may also need to pass the `dispatcher` through to the `baseFetch` call, as I don't think it will be included in the `new Request()` call at the top of the `fetch` function.

```ts
const response = await baseFetch(request, {
  dispatcher: requestInit.dispatcher,
});
```

mrbbot (commented):
Looking good! It would be nice to see a test of a `dispatchFetch()` call with the `https: true` option set in `test/index.spec.ts`. Ideally, the test would also check the log message for the server starting included the correct protocol too. See [here](https://github.com/cloudflare/miniflare/blob/377feaa16df9f35e333be15e74bcf25ee5b647e3/packages/miniflare/test/plugins/queues/index.spec.ts#L285) for an test asserting log output.

packages/miniflare/src/index.ts
jspspike: Would it make sense to just always pass `rejectUnauthorized: false` in `baseFetch` and the `NodeWebSocket`? According to the docs it seems like it would only matter when using https which we would always want to have this setting enabled for

mrbbot: I don't think. `wrangler pages dev` uses Miniflare's `fetch` to proxy to upstream services. If those are using HTTPS, we do actually want to reject unauthorised certificates.

packages/miniflare/test/index.spec.ts
```diff
@@ -15,6 +15,7 @@ import {
   MessageEvent as StandardMessageEvent,
   WebSocketServer,
 } from "ws";
+import { TestLog } from "./test-shared";
 import { useServer } from "./test-shared";
```
mrbbot, line 19: These imports should be collapsed into a single statement.

mrbbot (commented):
`KeyPair` should be removed from here too: https://github.com/cloudflare/miniflare/blob/c86b79a70c710501974ee143ea96b68f9bcf4469/packages/miniflare/src/plugins/index.ts#L98

packages/miniflare/src/http/server.ts
```diff
@@ -0,0 +1,68 @@
+import fs from "fs/promises";
+import { PluginSharedOptions } from "../index";
+import { HEADER_CF_BLOB, SERVICE_ENTRY, SOCKET_ENTRY } from "../plugins";
+import { HttpOptions, Socket, Socket_Https } from "../runtime";
+import { Awaitable } from "../shared";
+import { CERT, KEY } from "./cert";
+
+export async function configureEntrySocket(
+  sharedOpts: PluginSharedOptions
```
mrbbot, line 9: To avoid a cyclic dependency on `../index`, can you replace this with `coreOptions: z.infer<typeof CORE_PLUGIN.options>`? Then pass `sharedOpts.core` instead of `sharedOpts`.

packages/miniflare/src/index.ts
```diff
@@ -82,7 +82,7 @@ export type MiniflareOptions = SharedOptions &
 type PluginWorkerOptions = {
   [Key in keyof Plugins]: z.infer<Plugins[Key]["options"]>;
 };
-type PluginSharedOptions = {
+export type PluginSharedOptions = {
```
mrbbot, line 85: Ideally this shouldn't be exported, see the comment above.
```suggestion
type PluginSharedOptions = {
```

.gitguardian.yml
mrbbot, line 1: Given these files aren't working and we've split up `KEY`, could we remove this file and `.github/secret_scanning.yml`?

packages/miniflare/src/index.ts
```diff
@@ -883,7 +880,12 @@ export class Miniflare {
       forward.headers.delete("Content-Length");
     }
 
-    const response = await fetch(url, forward as RequestInit);
+    const forwardInit = forward as RequestInit;
+    if (url.protocol === "https:" && forwardInit) {
```
mrbbot, line 884: Super tiny nit: `forwardInit` will always be truthy, as `forward` is constructed as `new Request()` earlier on in this function.
```suggestion
    if (url.protocol === "https:") {
```

mrbbot (approved):
LGTM! ✅

PR #613, opened 2023-06-27
[Miniflare 3] Allow custom services to respond with `Content-Encoding` and multiple `Set-Cookie` headers

mrbbot:
Previously, if the response from a custom service included a `Content-Encoding` header, we would respond with decoded content to `workerd` without removing the `Content-Encoding` header. This caused `workerd` to try to decode already decoded content, resulting in an error. This was particularly a problem with `wrangler pages dev`, as upstream dev servers could very easily return gzipped content.

This change updates the `writeResponse()` function with code from Miniflare 2, to re-encode the body before responding.

Closes https://github.com/cloudflare/workers-sdk/issues/3408

---

Previously, if a custom service returned a response with multiple `Set-Cookie` headers, they would be combined into a single header.

This change uses the `set-cookie-parser` package to split the header, ensuring `Headers#getSetCookie()`/ `Headers#getAll("Set-Cookie")` return all cookies in a Worker.


RamIdeas (approved):

petebacondarwin:
Why was Ubuntu latest cancelled?

packages/miniflare/test/index.spec.ts
```diff
@@ -106,6 +108,50 @@ test("Miniflare: routes to multiple workers with fallback", async (t) => {
...
+    },
+  });
+  t.teardown(() => mf.dispose());
+
+  const test = async (encoding: string) => {
+    const res = await mf.dispatchFetch("http://localhost", {
+      headers: { "X-Test-Encoding": encoding },
+    });
+    t.is(res.headers.get("Content-Encoding"), encoding);
+    t.is(await res.text(), testBody, encoding);
```
petebacondarwin, line 142: Here the `encoding` variable is passed as the "message" of the `t.is()`. 
Is that what was meant?

petebacondarwin (commented):
I see this is merged. I was wondering how the test for the _transformsForContentEncoding()` works. Are you just using that to setup the encodings but then checking that the actual `writeResponse()` code does the correct thing? In other words, does this test actually check the `writeResponse()` code?

packages/miniflare/test/index.spec.ts
mrbbot: Yep, since the assertion on this line will be called multiple times, it wouldn't immediately be clear which encoding caused the failure. Passing `encoding` as the message parameter will include it in the error message. 👍

PR #616, opened 2023-06-27
[Miniflare 3] Use `http.STATUS_CODES` for logging status text

mrbbot:
Some responses, especially those from custom services, don't have a `statusText`. This means some logs only include a status code. We can derive the text from just the code using the `http.STATUS_CODES` constants built into Node.


packages/miniflare/src/workers/core/entry.worker.ts
```diff
@@ -144,7 +144,6 @@ function maybeLogRequest(
       headers: { "Content-Type": "application/json" },
       body: JSON.stringify({
         status: response.status,
-        statusText: response.statusText,
```
petebacondarwin, line 147: Is there any scenario where `response.statusText` is different to what would be retrieved from `http.STATUS_CODES`? I.e. should we try using this first and fall back to the lookup if not available?

petebacondarwin (approved):

packages/miniflare/src/workers/core/entry.worker.ts
mrbbot: Technically, you can specify any value you want I think. Trying the value then falling back seems like a good idea. 🙂 👍

petebacondarwin (approved):

PR #617, opened 2023-06-27
[Miniflare 3] Correctly resolve relative source mapped URLs

mrbbot:
Previously, if source maps included relative paths, those paths would not be resolved relative to the source map's own file path before inclusion into stack traces. This meant stack traces included relative paths that couldn't be resolved for click-to-definition by IDEs.

This change includes the URL of the source map upon retrieval, allowing `source-map-support` to include absolute source URLs in stack traces.

### DIFF
--- a/packages/miniflare/src/plugins/core/errors/index.ts
+++ b/packages/miniflare/src/plugins/core/errors/index.ts
@@ -177,7 +177,7 @@ function getSourceMappedStack(
     const sourceMapFile = maybeGetDiskFile(sourceMapPath);
     if (sourceMapFile === undefined) return null;
 
-    return { map: sourceMapFile.contents };
+    return { map: sourceMapFile.contents, url: sourceMapFile.path };
   }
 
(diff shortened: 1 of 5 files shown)


packages/miniflare/src/plugins/core/errors/index.ts
```diff
@@ -177,7 +177,7 @@ function getSourceMappedStack(
     const sourceMapFile = maybeGetDiskFile(sourceMapPath);
     if (sourceMapFile === undefined) return null;
 
-    return { map: sourceMapFile.contents };
+    return { map: sourceMapFile.contents, url: sourceMapFile.path };
```
petebacondarwin, line 180: This is literally it?

petebacondarwin (approved):

packages/miniflare/src/plugins/core/errors/index.ts
mrbbot: Yep 😅

PR #618, opened 2023-06-28
[Miniflare 3] Update API docs with `routes`, `queue*`, `https*`, and `upstream`

mrbbot:
This PR adds some missing API docs for #520, #566, #612, and #568. 👍

### DIFF
--- a/packages/miniflare/README.md
+++ b/packages/miniflare/README.md
@@ -435,11 +435,11 @@ Options shared between all Workers/"nanoservices".
 
 - `https?: boolean`
 
-  Start an HTTPS server using a pre-generated self-signed certificate for
-  `localhost`. Note this certificate is not valid for any other hostnames or IP
-  addresses. If you need to access the HTTPS server from another device, you'll
-  need to generate your own certificate and use the other `https*` options
-  below.
+  If `true`, start an HTTPS server using a pre-generated self-signed certificate
+  for `localhost`. Note this certificate is not valid for any other hostnames or
+  IP addresses. If you need to access the HTTPS server from another device,
+  you'll need to generate your own certificate and use the other `https*`
+  options below.
 
   ```shell
   $ openssl req -newkey rsa:2048 -nodes -keyout key.pem -x509 -days 365 -out cert.pem
--- a/packages/miniflare/README.md
+++ b/packages/miniflare/README.md
@@ -398,7 +398,7 @@ parameter in module format Workers.
 
 > :warning: Queues are only supported with Node.js 18 or above.
 
-- `quueueProducers?: Record<string, string> | string[]`
+- `queueProducers?: Record<string, string> | string[]`
 
   Record mapping binding name to queue names to inject as `WorkerQueue` bindings
   into this Worker. Different Workers may bind to the same queue name with
--- a/packages/miniflare/README.md
+++ b/packages/miniflare/README.md
@@ -187,7 +187,7 @@ level will be logged.
 - `maxRetries?: number`
 
   Maximum number of times to retry dispatching a message, if handling it throws,
-  or it is explicitly retired. Defaults to `2`.
+  or it is explicitly retried. Defaults to `2`.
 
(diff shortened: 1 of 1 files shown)


packages/miniflare/README.md
```diff
@@ -173,6 +173,27 @@ level will be logged.
...
+
+- `maxBatchTimeout?: number`
+
+  Maximum number of seconds to wait for a full batch. If a message is sent, and
+  this timeout elapses, a partial batch will be dispatched. Defaults to `1`.
+
+- `maxRetries?: number`
+
+  Maximum number of times to retry dispatching a message, if handling it throws,
+  or it is explicitly retired. Defaults to `2`.
```
petebacondarwin, line 190: In the docs it also says that a retry can be triggered by `retryAll()` being called.

packages/miniflare/README.md
```diff
@@ -173,6 +173,27 @@ level will be logged.
...
+
+- `deadLetterQueue?: string`
+
+  Name of another Queue to send a message on if it fails processing after
+  `maxRetries`. If this isn't specified, failed messages will be discarded.
```
petebacondarwin, line 195: Additional line in the docs for this configuration item:

> If there is no Queue with the specified name, it will be created automatically.

packages/miniflare/README.md
```diff
@@ -173,6 +173,27 @@ level will be logged.
...
+
```
petebacondarwin, line 196: The docs also describe a property called `max_concurrency`. Does Miniflare support this too?

packages/miniflare/README.md
```diff
@@ -366,7 +394,25 @@ parameter in module format Workers.
   ID with different binding names. If a `string[]` of binding names is
   specified, the binding name and database ID are assumed to be the same.
 
-#### Analytics Engine and Queues
+#### Queues
+
+> :warning: Queues are only supported with Node.js 18 or above.
+
+- `quueueProducers?: Record<string, string> | string[]`
```
petebacondarwin, line 401:
```suggestion
- `queueProducers?: Record<string, string> | string[]`
```

packages/miniflare/README.md
mrbbot: I figured explicitly retried would cover `Message#retry()` or `MessageBatch#retryAll()`.

packages/miniflare/README.md
mrbbot: This doesn't really matter for Miniflare, since we don't really have a concept of resource "creation". We implicitly create resources based on the names/IDs specified.

packages/miniflare/README.md
mrbbot: It does not. I believe Miniflare effectively implements a fixed value of `1` here.

packages/miniflare/README.md
mrbbot: 1c16010263476feb020b4b6051b14da62bae64bd

packages/miniflare/README.md
```diff
@@ -387,6 +433,45 @@ Options shared between all Workers/"nanoservices".
   specified port is in use, Miniflare throws an error, rather than attempting to
   find a free port.
 
+- `https?: boolean`
+
+  Start an HTTPS server using a pre-generated self-signed certificate for
```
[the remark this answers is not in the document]
mrbbot: 27a6a1001036893e9adae2b1bdf8336fe5be35da

packages/miniflare/README.md
petebacondarwin: Oh I now why I was confused. You have a typo: retired -> retried!

mrbbot: Ah 😂

mrbbot: f7d5f08c45e2d209f0efb7de60178d3749b30ca6

petebacondarwin (approved):
Glad you retired that typo!

PR #620, opened 2023-06-28
[Miniflare 3] Mark WebSocket `fetch()` close test as flaky

mrbbot:
We're seeing this `t.is(code, 3003)` assertion flake quite a bit in CI. Instead of being `3003`, the close code is `1006` which means "Abnormal Closure". I haven't been able to reproduce this locally, so suspect this is just some transient network failure with the Actions runner. Retrying the test seems to help, so this PR introduces a `flaky` helper for wrapping test implementations with. This retires a test 3 times.

### DIFF
--- a/packages/miniflare/test/test-shared/asserts.ts
+++ b/packages/miniflare/test/test-shared/asserts.ts
@@ -1,5 +1,6 @@
 import assert from "assert";
 import { ExecutionContext } from "ava";
+import { Awaitable } from "miniflare";
 
 export function isWithin(
   t: ExecutionContext,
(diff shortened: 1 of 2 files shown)


packages/miniflare/test/test-shared/asserts.ts
```diff
@@ -21,3 +22,22 @@ export function escapeRegexp(value: string): RegExp {
...
   return new RegExp(value.replace(/[.*+?^${}()|[\]\\]/g, "\\$&"));
 }
+
+export function flaky(
+  impl: (t: ExecutionContext) => Awaitable<void>
+): (t: ExecutionContext) => Promise<void> {
+  const maxAttempts = 3;
+  return async (t) => {
+    for (let attempt = 1; attempt <= maxAttempts; attempt++) {
+      const result = await t.try(impl);
```
mrbbot, line 32: Docs on `t.try()` can be found here: https://github.com/avajs/ava/blob/main/docs/03-assertions.md#trytitle-implementation--macro-args. It's kinda neat... 🙂

penalosa (approved):

PR #628, opened 2023-07-13
Removing shim from D1

geelen:
Now that workerd [includes the D1 shim](https://github.com/cloudflare/workerd/blob/main/src/cloudflare/internal/d1-api.ts), we no longer need wrangler to bundle it, and no longer require all bindings to be prefixed with `__D1_BETA__`.

This will need to be released alongside the Wrangler version that drops support for the shim as well, currently being tracked in this PR: https://github.com/cloudflare/workers-sdk/pull/3595


packages/miniflare/test/fixtures/d1/README.md
```diff
@@ -1,11 +1,11 @@
 # D1 Worker Fixture
```
mrbbot, line 1: The only reason we had this fixture was to get the D1 shim from Wrangler for testing. `miniflareTest()` (see `test/plugins/d1/index.spec.ts`) can accept a request handler function like:
https://github.com/cloudflare/miniflare/blob/c0f32f331dad9569a91adeabfb632b8b3937a4b6/packages/miniflare/test/plugins/cache/index.spec.ts#L13-L58
This handler gets wrapped with all the `MF-Experimental-Error-Stack` stuff automatically too. Could tidy this up in a future PR though.

packages/miniflare/src/plugins/d1/index.ts
```diff
@@ -32,7 +32,15 @@ export const D1_PLUGIN: Plugin<
     const databases = namespaceEntries(options.d1Databases);
     return databases.map<Worker_Binding>(([name, id]) => ({
       name,
-      service: { name: `${SERVICE_DATABASE_PREFIX}:${id}` },
+      wrapped: {
+        moduleName: "cloudflare-internal:d1-api",
+        innerBindings: [
```
mrbbot, line 37: What you've written here is correct. Looks like the type definition for `Worker_Binding_WrappedBinding` is missing an `s`:
https://github.com/cloudflare/miniflare/blob/c0f32f331dad9569a91adeabfb632b8b3937a4b6/packages/miniflare/src/runtime/config/workerd.ts#L137
That file is currently handwritten, would you mind adding it?

mrbbot (commented):
Nice! 🥳 Couple minor comments...

packages/miniflare/test/fixtures/d1/README.md
geelen: I ended up just reusing the existing test file, before Wrangler had compiled it. What do you think?

geelen:
Updated the PR to make Miniflare backwards-compatible with the older style of bindings.

I also broke the test apart into a `shared.ts`, which has the basic test infra, and then a `suite.ts`, since the two test cases differ only on which file they're including, and the binding name.

Not sure this is great but for right now, all I want to confirm is that the runtime shim has the exact same behaviour as the Wrangler-based shim...

packages/miniflare/src/plugins/d1/index.ts
```diff
@@ -30,10 +30,27 @@ export const D1_PLUGIN: Plugin<
...
-      name,
-      service: { name: `${SERVICE_DATABASE_PREFIX}:${id}` },
-    }));
+    return databases.map<Worker_Binding>(([name, id]) => {
+      const binding = name.startsWith("__D1_BETA__")
+        ? // Used before Wrangler 3.3
+          {
+            service: { name: `${SERVICE_DATABASE_PREFIX}:${id}` },
+          }
+        : // Used after Wrangler 3.3
```
geelen, line 39: Kinda assuming this will be the version where the D1 shim drops...

mrbbot (approved):
LGTM! ✅

PR #629, opened 2023-07-14
[Miniflare 3] Add `outboundService` option for customising global `fetch()` target

mrbbot:
`workerd` exposes a `globalOutbound` option for customising which service global `fetch()` calls should be dispatched to. For tests, it's useful to be able to mock responses to certain routes or assert certain requests are made. This new option allows `fetch()` call to be sent to another configured Worker. See the added test for an example.

Note support for `undici`'s [`MockAgent`](https://undici.nodejs.org/#/docs/api/MockAgent) API in Miniflare 3 is still planned. This is a lower-level primitive for customising `fetch()` responses.

/cc @vlovich

### DIFF
--- a/packages/miniflare/README.md
+++ b/packages/miniflare/README.md
@@ -315,6 +315,12 @@ parameter in module format Workers.
     handler. This allows you to access data and functions defined in Node.js
     from your Worker.
 
+- `outboundService?: string | { network: Network } | { external: ExternalServer } | { disk: DiskDirectory } | (request: Request) => Awaitable<Response>`
+
+  Dispatch this Worker's global `fetch()` and `connect()` requests to the
+  configured service. Service designators follow the same rules above for
+  `serviceBindings`.
+
 - `routes?: string[]`
 
   Array of route patterns for this Worker. These follow the same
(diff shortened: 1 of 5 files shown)


vlovich:
Omg. I added so much custom code to emulate this. If I'd know I would have asked for this sooner. LGTM

vlovich (approved):

kentonv:
Note that this setting affects more than just fetch() -- it will also affect where connect() goes, and would presumably affect similar functions that we add in the future. Hence why the setting in `workerd` is called `globalOutbound` and not `globalFetch`.

What you want to name it is up to you but just wanted to make sure this was understood.

vlovich:
@mrbbot is there a way to make it easier to have a `globalInternet` option that takes a function & sets up the requisite magic worker to tunnel out and invoke the function?

penalosa (commented):
Looks good! Total bike shedding, but why the change in name from `globalOutbound`?

penalosa (approved):

limzykenneth:
Has this feature landed in the latest release yet? The documentation already contains the `outboundService` option but it does not seem to work yet.

limzykenneth:
I doubled checked and turns out I was missing some bindings and fetch is just failing instead. 

Sorry another question though, is it possible to filter which request get passed to the `outboundService` while the rest will route as usual, for example only intercept requests to a certain hostname but for everything else make the network request as usual?

PR #631, opened 2023-07-17
[Miniflare 3] Fix standalone support for Workers Sites using modules

### LINKED ISSUE #630: Cannot get sitePath to work
Trying to start Miniflare serving a site too, with this code:

```js
const mf = new Miniflare({
    host: options.address,
    port: options.port,
    workers: [
        {
            name: 'myapp',
            script: script.toString('utf8'),
            scriptPath: 'index.mjs',
            modules: true,
            modulesRules: [{type: 'ESModule', include: ['**/*.js', '**/*.mjs']}],
            compatibilityDate: '2023-05-18',
            sitePath: options.sitePath,
        },
    ],
})
```

However, this fails with:

> MiniflareCoreError [ERR_MODULE_RULE]: Unable to resolve "index.mjs" dependency "__STATIC_CONTENT_MANIFEST": no matching module rules.
> If you're trying to import an npm package, you'll need to bundle your Worker first

If I modify `getServices` to add a `console.log`, I can see that `additionalModules` is populated correctly:

```js
[
  {
    name: '__STATIC_CONTENT_MANIFEST',
    text: '{"index.html":"$__MINIFLARE_SITES__$/index.html"}'
  }
]
```

However, it appears that the next line, invoking `getWorkerScript`, doesn't get this manifest and panics.

mrbbot:
Previously, attempting to use Workers Sites with `modules: true` would fail as the automatic module locator didn't know about `__STATIC_CONTENT_MANIFEST`. This change passes through additional modules to the locator. This change also adds support for internal `"workerd:"` modules to the locator (e.g. `workerd:rtti`).

Additionally, `import manifest from "__STATIC_CONTENT_MANIFEST"` was only allowed in the "root" directory of the modules virtual file-system. This change adds stub modules for each additional module in each subdirectory, that re-export the module. This allows the manifest to be imported in any directory.

Closes #630

### DIFF
--- a/packages/miniflare/src/plugins/core/index.ts
+++ b/packages/miniflare/src/plugins/core/index.ts
@@ -563,10 +563,10 @@ function getWorkerScript(
   options: SourceOptions,
   workerIndex: number
 ): { serviceWorkerScript: string } | { modules: Worker_Module[] } {
+  const modulesRoot =
+    ("modulesRoot" in options ? options.modulesRoot : undefined) ?? "";
   if (Array.isArray(options.modules)) {
     // If `modules` is a manually defined modules array, use that
-    const modulesRoot =
-      ("modulesRoot" in options ? options.modulesRoot : undefined) ?? "";
     return {
       modules: options.modules.map((module) =>
         convertModuleDefinition(sourceMapRegistry, modulesRoot, module)
@@ -589,7 +589,11 @@ function getWorkerScript(
 
   if (options.modules) {
     // If `modules` is `true`, automatically collect modules...
-    const locator = new ModuleLocator(sourceMapRegistry, options.modulesRules);
+    const locator = new ModuleLocator(
+      sourceMapRegistry,
+      modulesRoot,
+      options.modulesRules
+    );
     // If `script` and `scriptPath` are set, resolve modules in `script`
     // against `scriptPath`.
     locator.visitEntrypoint(
--- a/packages/miniflare/src/plugins/core/index.ts
+++ b/packages/miniflare/src/plugins/core/index.ts
@@ -374,10 +374,15 @@ export const CORE_PLUGIN: Plugin<
     sourceMapRegistry,
   }) {
     // Define regular user worker
+    const additionalModuleNames = additionalModules.map(({ name }) => {
+      assert(name !== undefined);
+      return name;
+    });
     const workerScript = getWorkerScript(
       sourceMapRegistry,
       options,
-      workerIndex
+      workerIndex,
+      additionalModuleNames
     );
     // Add additional modules (e.g. "__STATIC_CONTENT_MANIFEST") if any
     if ("modules" in workerScript) {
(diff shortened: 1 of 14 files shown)


packages/miniflare/src/plugins/core/index.ts
```diff
@@ -374,14 +375,39 @@ export const CORE_PLUGIN: Plugin<
...
+        // subdirectory re-exporting each additional module. These allow
+        // additional modules to be imported in every directory.
+        for (const subDir of subDirs) {
+          const relativePath = path.posix.relative(subDir, module.name);
+          const relativePathString = JSON.stringify(relativePath);
+          workerScript.modules.push({
+            name: path.posix.join(subDir, module.name),
+            // TODO(someday): if we ever have additional modules without
+            //  default exports, this may be a problem, for now, our only
+            //  additional module is `__STATIC_CONTENT_MANIFEST` so this is fine
```
RamIdeas, line 406: I guess we could read the file and look for occurrences of either `/\bexport\s+default\b/` or `\bas\s+default\b` and use that to determine whether we include the second ` export { default } from` line.

packages/miniflare/test/plugins/kv/sites.spec.ts
```diff
@@ -0,0 +1,302 @@
...
+const FIXTURES_PATH = path.resolve(
+  __dirname,
+  "..",
+  "..",
+  "..",
+  "..",
+  "test",
+  "fixtures",
+  "sites"
+);
```
RamIdeas, line 18: nit: path.resolve will normalise `/` to `\` on Windows so you can write this as:

```suggestion
const FIXTURES_PATH = path.resolve(__dirname, "../../../../test/fixtures/sites");
```

packages/miniflare/test/plugins/kv/sites.spec.ts
```diff
@@ -0,0 +1,302 @@
...
+  t.is(res1.headers.get("CF-Cache-Status"), "MISS");
+  t.is(await res1.text(), "1");
+
+  await fs.writeFile(testPath, "2", "utf8");
+  const res2 = await mf.dispatchFetch("http://localhost:8787/test.txt");
+  t.is(res2.headers.get("CF-Cache-Status"), "MISS");
+  t.is(await res2.text(), "2");
+});
+
+test("gets assets with module worker", async (t) => {
```
RamIdeas, line 151: is it worth adding a test for a nested asset?

RamIdeas (approved):
minor comments

packages/miniflare/src/plugins/core/index.ts
mrbbot: That seems like a reasonable heuristic. Updated the comment so we know what to do in the future. 👍

packages/miniflare/test/plugins/kv/sites.spec.ts
mrbbot: That's fair, think I'll leave this as is for now, since there are quite a few other instances like this in the codebase.

packages/miniflare/test/plugins/kv/sites.spec.ts
mrbbot: Can't hurt, added 👍

PR #637, opened 2023-07-24
[Miniflare 3] Split certificates from `NODE_EXTRA_CA_CERTS`

mrbbot:
If `NODE_EXTRA_CA_CERTS` contains multiple certificates, they [need to be split](https://github.com/cloudflare/miniflare/pull/587/files#r1271579671) before being passed to `workerd`. This change fixes that, and adds a test for `NODE_EXTRA_CA_CERTS` handling in general [as promised](https://github.com/cloudflare/miniflare/pull/587#discussion_r1206509748).

### DIFF
--- a/packages/miniflare/src/plugins/core/index.ts
+++ b/packages/miniflare/src/plugins/core/index.ts
@@ -65,7 +65,12 @@ if (process.env.NODE_EXTRA_CA_CERTS !== undefined) {
   // https://nodejs.org/api/cli.html#node_extra_ca_certsfile
   try {
     const extra = readFileSync(process.env.NODE_EXTRA_CA_CERTS, "utf8");
-    trustedCertificates.push(extra);
+    // Split bundle into individual certificates and add each individually:
+    // https://github.com/cloudflare/miniflare/pull/587/files#r1271579671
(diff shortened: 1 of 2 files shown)


RamIdeas (approved):

packages/miniflare/src/plugins/core/index.ts
```diff
@@ -65,7 +65,12 @@ if (process.env.NODE_EXTRA_CA_CERTS !== undefined) {
   // https://nodejs.org/api/cli.html#node_extra_ca_certsfile
   try {
     const extra = readFileSync(process.env.NODE_EXTRA_CA_CERTS, "utf8");
-    trustedCertificates.push(extra);
+    // Split bundle into individual certificates and add each individually:
+    // https://github.com/cloudflare/miniflare/pull/587/files#r1271579671
+    const pemBegin = "-----BEGIN";
```
kentonv, line 70: Does PEM actually specify exactly 5 `-` and no whitespace? I've always assumed the format was looser than that but I guess I don't know.

mrbbot: Looks like it: https://datatracker.ietf.org/doc/html/rfc7468#section-2

PR #650, opened 2023-08-09
Switch default to `localhost` instead of `127.0.0.1`

jspspike:
Also allow host to use ipv6 addresses


packages/miniflare/src/index.ts
```diff
@@ -438,9 +438,14 @@ export class Miniflare {
     this.#workerOpts = workerOpts;
     this.#log = this.#sharedOpts.core.log ?? new NoOpLog();
     this.#timers = this.#sharedOpts.core.timers ?? defaultTimers;
-    this.#host = this.#sharedOpts.core.host ?? "127.0.0.1";
+    this.#host = this.#sharedOpts.core.host ?? "localhost";
```
mrbbot, line 441: A slight problem here is that `this.#host` is also used as the hostname for the loopback server. https://github.com/cloudflare/miniflare/blob/39fd9109150a4ce5b7fc1d5fcf47b1674ba1d969/packages/miniflare/src/index.ts#L525
This needs to be accessible on the same hosts as the runtime itself for live reload to work. Unfortunately, unlike `workerd`, passing `localhost` as the hostname to Node's `http.createServer().listen()` only listens on the first address returned by DNS lookup. In Node.js 17+, this is usually the IPv6 address `::1`, making the loopback server unaccessible on `127.0.0.1`.

You can perform DNS lookup in Node using something like...

```js
import dns from "dns/promises";
const results = await dns.lookup("localhost", { all: true });
```

We could use the results of this to start multiple loopback servers on all the required hosts, but the trick there is to make sure they're all started on the same port, even if `port: 0` is specified. You could do that by starting the first, then using `server.address().port` to get the port, and then start the rest. This all seems quite susceptible to race conditions though... 😕

packages/miniflare/src/index.ts
```diff
@@ -440,7 +440,14 @@ export class Miniflare {
     this.#timers = this.#sharedOpts.core.timers ?? defaultTimers;
     this.#host = this.#sharedOpts.core.host ?? "127.0.0.1";
     this.#accessibleHost =
-      this.#host === "*" || this.#host === "0.0.0.0" ? "127.0.0.1" : this.#host;
+      this.#host === "*" || this.#host === "0.0.0.0" || this.#host === "::"
+        ? "localhost"
```
mrbbot, line 444: It looks like `dispatchFetch()` doesn't work on Node 17+ if `host: "0.0.0.0"` is set with this change. I think this makes sense, as `undici` will resolve `localhost` to `::1` which `0.0.0.0` won't be listening on.  I'd change `localhost` to `127.0.0.1` here, which should be valid for all the `#host`'s checked here.

mrbbot: It also looks like `*` isn't a valid hostname for `http.createServer().listen()`, but it is supported by `workerd`. I think we'll need to add `if (hostname === "*") hostname = "::";` at the start of `#startLoopbackServer()` in this file.

jspspike:
Small change but do you think we should maybe change [this](https://github.com/cloudflare/miniflare/blob/tre/packages/miniflare/src/index.ts#L880) to host or something else? Just to let the user know that it's listening on more than `127.0.0.1`?

mrbbot (approved):
> maybe change [this](https://github.com/cloudflare/miniflare/blob/tre/packages/miniflare/src/index.ts#L880) to host or something else?

Hmmm, we could do. The reason it's accessible host is so that when IDE's autolink it in integrated terminals, it goes straight to the browser. Miniflare 2 solved this by logging the listening host first, then all IPs this was accessible on:
https://github.com/cloudflare/miniflare/blob/7e4d906e19cc69cd3446512bfeb7f8aee3a2bda7/packages/http-server/src/index.ts#L481-L484
Maybe we should do something like that?

mrbbot (approved):

PR #652, opened 2023-08-10
Trigger scheduled handler on '/cdn-cgi/mf/scheduled'

jspspike:


packages/miniflare/test/index.spec.ts
```diff
@@ -498,3 +498,26 @@ test("Miniflare: Accepts https requests", async (t) => {
...
+
+  const mf = new Miniflare({
+    log,
+    modules: true,
+    script: `export default {
+      fetch() {
+        return new Response("Hello world");
+      },
+      scheduled() {
+        return new Response("Hello cron");
```
mrbbot, line 513: Could you have the `scheduled()` handler set a module level variable in this `script`, that is then returned by the `fetch()` handler, then swap the order of the `disatchFetch()`es below? This would make sure the `scheduled()` handler is being called in this test.

packages/miniflare/test/plugins/d1/suite.ts
```diff
@@ -192,7 +192,7 @@ export default (binding: string, WORKER_PATH: string) => {
     const select = await db.prepare(`SELECT * FROM ${tableColours}`);
     let result: ColourRow | null = await select.first<ColourRow>();
     t.deepEqual(result, { id: 1, name: "red", rgb: 0xff0000 });
-    let id = await select.first<number>("id");
+    let id: number | null = await select.first<number>("id");
```
mrbbot, line 195: Is this type annotation needed? It should get inferred.

mrbbot (commented):
Nice! Added a few comments... 🙂

packages/miniflare/test/plugins/d1/suite.ts
jspspike: When I remove it I get a type error. I think because `first` returns a `number` type when id is reassigned below it says can't assign `number | null` to a `number` type

packages/miniflare/test/index.spec.ts
```diff
@@ -498,3 +498,32 @@ test("Miniflare: Accepts https requests", async (t) => {
...
 });
+
+test("Miniflare: Manually triggered scheduled events", async (t) => {
+  const log = new TestLog(t);
+
+  const mf = new Miniflare({
+    log,
+    modules: true,
+    script: `
+    var scheduledRun = false;
```
mrbbot, line 509: Nit: generally you don't want to use `var`. Can you switch this to `let`?

packages/miniflare/test/index.spec.ts
```diff
@@ -498,3 +498,32 @@ test("Miniflare: Accepts https requests", async (t) => {
...
+    export default {
+      fetch() {
+        return new Response(scheduledRun);
+      },
+      scheduled() {
+        scheduledRun = true;
+        return new Response("Hello cron");
```
mrbbot, line 516: Nit: since this return value is ignored, could you remove it?

packages/miniflare/test/plugins/d1/suite.ts
mrbbot: Ah ok, that makes sense. 👍

packages/miniflare/src/workers/core/entry.worker.ts
```diff
@@ -159,11 +159,29 @@ async function handleQueue(
...
     Date.now() - startTime;
   return Response.json(queueResponse);
 }
 
+async function handleScheduled(
+  params: URLSearchParams,
+  service: Fetcher
+): Promise<Response> {
+  const date = params.get("time");
+  const scheduledTime = date ? new Date(date) : undefined;
```
mrbbot, line 172: `date` is a string here, so `new Date()` will parse it as a date string. What we want here is for `time` to be a unix timestamp. You'll want `new Date(parseInt(date))` instead. 👍

mrbbot (approved):

PR #654, opened 2023-08-11
Change r2 put response encoding

robbertvanginkel:
Without this, the `await env.MY_BUCKET.put(...)` of https://developers.cloudflare.com/r2/api/workers/workers-api-usage/ hangs indefinitely

This problem matches the description in https://github.com/cloudflare/workers-sdk/issues/3688 and https://discord.com/channels/595317990191398933/1138945222873731264/1138945222873731264.

### DIFF
--- a/packages/miniflare/src/plugins/r2/router.ts
+++ b/packages/miniflare/src/plugins/r2/router.ts
@@ -144,7 +144,7 @@ export class R2Router extends Router<R2Gateway> {
         valueSize,
         metadata
       );
-      return encodeResult(result);
+      return encodeJSONResult(result);
     } else if (metadata.method === "createMultipartUpload") {
       const result = await gateway.createMultipartUpload(
         metadata.object,


robbertvanginkel:
The `tre` branch seems to be where the currently released version of miniflare is? Are there any tests for this behaviour to verify?

I'm not completely sure why `encodeResult` doesn't work here, but with it I observed that `writeResponse` would be get stuck in this loop: https://github.com/cloudflare/miniflare/blob/cf080a81ff72902f3b2f86e344beeaeffddebfdd/packages/miniflare/src/index.ts#L350-L353

robbertvanginkel:
Thanks for the response! I was already getting a bit confused as I thought this had worked on wrangler v3 before but after returning to a project after a while it had suddenly stopped. Couldn't get it to work by downgrading wrangler/miniflare but didn't think node would have been the issue...

With my node fixed to the LTS version the old code seems to work again, I'll close this out and lets wait for the node fix then.

PR #661, opened 2023-08-16
fix: add `type: "module"` to vitest-environment-miniflare

### LINKED ISSUE #645: [Bug] vitest-environment-miniflare breaks when using vitest 0.34.1
## Package versions
vitest: 0.34.1
vitest-environment-miniflare: 2.14.0

NOTE: I know for a fact that everything was working correctly with vitest `0.33.0`. The issue surfaced just in the latest release.
As I'll state later, the error does seem to be with the `vitest-environment-miniflare` package, hence I'm making the issue here.

## Bug description
When upgrading to vitest 0.34.1 the following error shows up
```
SyntaxError: Cannot use import statement outside a module
 ❯ internalCompileFunction node:internal/vm:73:18
 ❯ wrapSafe node:internal/modules/cjs/loader:1178:20
 ❯ Module._compile node:internal/modules/cjs/loader:1220:27
 ❯ Module._extensions..js node:internal/modules/cjs/loader:1310:10
 ❯ Module.load node:internal/modules/cjs/loader:1119:32
 ❯ Module._load node:internal/modules/cjs/loader:960:12
 ❯ ModuleWrap.<anonymous> node:internal/modules/esm/translators:169:29
 ❯ ModuleJob.run node:internal/modules/esm/module_job:194:25

Module /PATH_TO_PROJECT/node_modules/vitest-environment-miniflare/dist/src/index.js:10 seems to be an ES Module but shipped in a CommonJS package. You might want to create an issue to the package "vitest-environment-miniflare" asking them to ship the file in .mjs extension or add "type": "module" in their package.json.

As a temporary workaround you can try to inline the package by updating your config:

// vitest.config.js
export default {
  test: {
    deps: {
      inline: [
        "vitest-environment-miniflare"
      ]
    }
  }
}
```

Looking into the node_modules folder it seems like vitest is correct? The dist/src/index.js file does use import statements and seems to present itself as an ES module, but no `"type": "module"` config is present in package.json.

The curious part is that vitest has never complained about this until the latest release. I'm making the issue here since it does look like the error message is highlighting a legit flaw in `vitest-environment-miniflare`

## Reproduction repository
The project I'm working on is small enough that I think I can point to it for anyone wanting a repository to reproduce this issue.
Here is a link to the branch with the issue present - https://github.com/AdiRishi/turborepo-remote-cache-cloudflare/tree/arishi-dev
Just clone it, run `yarn install` and then `yarn build && yarn test` to trigger the bug.

It's not just this project however, pretty much all my worker projects saw the same bug when I tried to update to the latest vitest version.

Hope this is enough detail. Let me know if I can provide any more information.

Averethel:
fixes #645

### DIFF
--- a/.changeset/fuzzy-poets-share.md
+++ b/.changeset/fuzzy-poets-share.md
@@ -1,0 +1,5 @@
+---
+"vitest-environment-miniflare": patch
+---
+
+fix: add `type: "module"` to vitest-environment-miniflare
--- a/packages/vitest-environment-miniflare/package.json
+++ b/packages/vitest-environment-miniflare/package.json
@@ -11,6 +11,7 @@
   ],
   "author": "MrBBot <me@mrbbot.dev>",
   "license": "MIT",
+  "type": "module",
   "main": "./dist/src/index.js",
   "module": "./dist/src/index.js",
   "types": "./dist/src/index.d.ts",


Averethel:
@mrbbot thank you! I was about to reach out for help.
I tried to figure it out independently, but I couldn't because of my lack of experience with the tooling.
Another approach I thought of (but couldn't explore) is some post-processing of the package.json in the build process. Not sure if your build pipeline supports it though

PR #678, opened 2023-09-06
[Miniflare 3] Remove unused `handleQueue()` function from entry worker

mrbbot:
With #656, the Queues dispatcher is now implemented as part of the broker Durable Object. We no longer send message batches directly from Node.js, so can remove queue handling from the entry worker.

Note the magic proxy enqueues messages through queue producer bindings like regular workers, so never used this endpoint directly.


RamIdeas (approved):

petebacondarwin (approved):
No changes to tests because this is never used and makes no externally measurable change to behaviour?

PR #679, opened 2023-09-07
Add dispatchNamespace binding support

jspspike:


mrbbot (commented):
Thanks for doing this! 🎉  Apologies for the delay in reviewing, I now understand Workers for Platforms enough to review this... 😅 I've added some comments. 👍

PR #681, opened 2023-09-08
[Miniflare 3] Improved source map handling and support for VSCode's debugger

mrbbot:
Hey! 👋 This PR includes a bunch of changes aimed at improving debugging support. Notably, this PR adds support for VSCode's built-in debugger, and step-through debugging of Miniflare's internal Workers. See the individual commit descriptions for more details.


scripts/build.mjs
```diff
@@ -102,6 +101,8 @@ const embedWorkersPlugin = {
           format: "esm",
           target: "esnext",
           bundle: true,
+          sourcemap: true,
+          sourcesContent: false,
```
mrbbot, line 105: This change increases Miniflare's bundle size from ~910KB to ~1080KB (~20% increase). We could only include these source maps in development, but having them in production is very useful for issue reports.

packages/miniflare/src/plugins/core/modules.ts
```diff
@@ -133,6 +133,16 @@ function moduleName(modulesRoot: string, modulePath: string) {
...
   return path.sep === "\\" ? name.replaceAll("\\", "/") : name;
 }
+export function withSourceURL(script: string, scriptPath: string): string {
+  let scriptURL: URL | string = scriptPath;
+  if (maybeGetStringScriptPathIndex(scriptPath) === undefined) {
+    scriptURL = pathToFileURL(scriptPath);
+  }
+  let sourceURL = `//# sourceURL=${scriptURL}\n`;
+  // Make sure `//# sourceURL` comment is on its own line
+  if (!script.endsWith("\n")) sourceURL = "\n" + sourceURL;
```
RamIdeas, line 143: can the preceding `\n` be added unconditionally?

```suggestion
  const sourceURL = `\n//# sourceURL=${scriptURL}\n`;
```

RamIdeas (approved):

scripts/build.mjs
RamIdeas: I think that's a fair cost for the inspectability. Compressed is probably much less significant

packages/miniflare/src/plugins/core/modules.ts
mrbbot: Yep, looks like that works. 👍 That's much cleaner. 🙂

PR #685, opened 2023-09-09
[Miniflare 3] Switch Miniflare's CI back to Node 20

mrbbot:
Hey! 👋 We switched to Node 18 as part of #623, to get around cloudflare/workers-sdk#3626. Now that Miniflare's simulators are implemented as Durable Objects, we no longer hit the erroneous Node code paths, so can switch back to version 20.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,7 +19,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        node: [16.13.0, 18]
+        node: [16.13.0, 20]
         # See supported Node.js release schedule at https://nodejs.org/en/about/releases/
 
     steps:


.github/workflows/test.yml
```diff
@@ -19,7 +19,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        node: [16.13.0, 18]
+        node: [16.13.0, 20]
```
rozenmd, line 22: With Node 16 going EOL today, do we still want to build for it?

mrbbot: I would think changing our minimum supported Node version would require a major version bump. We'd also probably want to bump Wrangler's minimum support version too.

penalosa: However, could we bump this to 16.14 to be inline with Wrangler?

penalosa (approved):

PR #687, opened 2023-09-12
[Miniflare 3] Assorted port-related improvements

mrbbot:
Hey! 👋 This PR includes 3 changes primarily intended to improve the reliability of `wrangler dev`. See the commit descriptions for the rationale behind each change.

### DIFF
--- a/packages/miniflare/src/index.ts
+++ b/packages/miniflare/src/index.ts
@@ -81,6 +81,7 @@ import {
   Service,
   Socket,
   SocketIdentifier,
+  SocketPorts,
   Worker_Binding,
   Worker_Module,
   kInspectorSocket,
@@ -546,6 +547,7 @@ export function _initialiseInstanceRegistry() {
 }
 
 export class Miniflare {
+  #previousSharedOpts?: PluginSharedOptions;
   #sharedOpts: PluginSharedOptions;
   #workerOpts: PluginWorkerOptions[];
   #log: Log;
@@ -553,7 +555,7 @@ export class Miniflare {
   readonly #runtime?: Runtime;
   readonly #removeRuntimeExitHook?: () => void;
   #runtimeEntryURL?: URL;
-  #socketPorts?: Map<SocketIdentifier, number>;
+  #socketPorts?: SocketPorts;
   #runtimeClient?: Client;
   #proxyClient?: ProxyClient;
 
@@ -877,6 +879,21 @@ export class Miniflare {
     });
   }
 
+  #getSocketAddress(
+    id: SocketIdentifier,
+    previousRequestedPort: number | undefined,
+    host = DEFAULT_HOST,
+    requestedPort?: number
+  ) {
+    // If `port` is set to `0`, was previously set to `0`, and we previously had
+    // a port for this socket, reuse that random port
+    if (requestedPort === 0 && previousRequestedPort === 0) {
+      requestedPort = this.#socketPorts?.get(id);
+    }
+    // Otherwise, default to a new random port
+    return `${host}:${requestedPort ?? 0}`;
+  }
+
   async #assembleConfig(loopbackPort: number): Promise<Config> {
     const allWorkerOpts = this.#workerOpts;
     const sharedOpts = this.#sharedOpts;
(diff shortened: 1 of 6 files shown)


packages/miniflare/src/index.ts
```diff
@@ -956,28 +981,46 @@ export class Miniflare {
...
-    if (maybePort === undefined) {
+    if (maybeSocketPorts === undefined) {
       throw new MiniflareCoreError(
         "ERR_RUNTIME_FAILURE",
         "The Workers runtime failed to start. " +
           "There is likely additional logging output above."
       );
     }
+    // Note: `updateConfig()` doesn't resolve until ports for all required
+    // sockets have been recorded. At this point, `maybePorts` contains all of
```
RamIdeas, line 1013: `maybePorts` -> `maybeSocketPorts`

packages/miniflare/test/index.spec.ts
```diff
@@ -96,22 +96,40 @@ test("Miniflare: ready returns copy of entry URL", async (t) => {
...
+  opts.host = "0.0.0.0";
   await mf.setOptions(opts);
   const updatedURL = await mf.ready;
+  res = await mf.dispatchFetch("http://localhost");
+  const updatedLoopbackPort = loopbackPortRegexp.exec(await res.text())?.[1];
 
+  // Make sure a new was allocated when `port: 0` was passed to `setOptions()`
   t.not(initialURL.port, "0");
-  t.is(initialURL.port, updatedURL.port);
+  t.not(initialURL.port, updatedURL.port);
```
RamIdeas, line 127: I don't think this is actually what we were shooting for.

I think the previous behaviour of: if option `port: 0` was set originally and remained `0` on `setOptions`, retaining the original port is very helpful and expected.

What we need for wrangler, if option `port` was set to something other that `0` originally and changed to another number (anything, even `0`), the port must change (the bug before was that miniflare retained the previous port, even if both times they were non zero and different)

RamIdeas (changes requested):

packages/miniflare/test/index.spec.ts
```diff
@@ -96,22 +96,40 @@ test("Miniflare: ready returns copy of entry URL", async (t) => {
...
+  let res = await mf.dispatchFetch("http://localhost");
+  const initialLoopbackPort = loopbackPortRegexp.exec(await res.text())?.[1];
 
+  opts.host = "0.0.0.0";
   await mf.setOptions(opts);
   const updatedURL = await mf.ready;
+  res = await mf.dispatchFetch("http://localhost");
+  const updatedLoopbackPort = loopbackPortRegexp.exec(await res.text())?.[1];
 
+  // Make sure a new was allocated when `port: 0` was passed to `setOptions()`
```
RamIdeas, line 125: "new was" -> "new port was"

packages/miniflare/test/index.spec.ts
mrbbot: So if the previous port was `0`, and the new `setOptions()` port is `0`, Miniflare should use the previously allocated port?

There's a small race condition here if the OS allocates that previously allocated port somewhere else, between when the old `workerd` instance is terminated, and the new `workerd` instance is started. Admittedly, that's quite unlikely, but it's what I was trying to avoid with this change.

I guess my thinking was if `port: 0` is being used (the default for Miniflare's API), I'd rather `setOptions()`s was guaranteed to be safe. The previous behaviour can be achieved with something like...

```ts
const opts: MiniflareOptions = {
  port: 0
};
const mf = new Miniflare(opts);
const url = await mf.ready;
// ...
opts.port = parseInt(url.port);
await mf.setOptions(opts);
```

 We could maybe have this behaviour if `port` was `undefined`, but that might complicate things. 🤔

RamIdeas: I agree that race condition is quite unlikely. More so because the port was recently bound by ourselves so we can assume we can rebind to it

The DX of changing the existing config to reset the config without changes is suboptimal. The user would also have to do the same for the inspectorPort and the unsafeDirectPorts on each subworker which gets even harder.

I made this ticket https://jira.cfdata.org/browse/DEVX-927

mrbbot: Updated with this behaviour, except for...
- `unsafeDirectPort: 0`: it's not really clear which allocated port we should reuse here
- `port: undefined`/`inspectorPort: undefined`: in these cases, the `undefined` suggests the user really doesn't care what the port is, so allocating a random one each time should be fine. This also makes the mentioned above race condition impossible in tests where users are unlikely to be specifying ports, but are likely be creating lots of `Miniflare` instances and frequently reloading them.

RamIdeas (approved):

PR #689, opened 2023-09-20
decouple MF-Original-URL header from disabling the pretty-error page

RamIdeas:
See https://jira.cfdata.org/browse/DEVX-945 for context

TL;DR this PR decouples the `MF-Original-URL` header from disabling the pretty-error page and instead uses an explicit `MF-Disable-Pretty-Error` header to disable it.

This allows `MF-Original-URL` to be set without disabling the pretty-error page.

### DIFF
--- a/packages/miniflare/src/index.ts
+++ b/packages/miniflare/src/index.ts
@@ -1155,6 +1155,7 @@ export class Miniflare {
     const forward = new Request(input, init);
     const url = new URL(forward.url);
     forward.headers.set(CoreHeaders.ORIGINAL_URL, url.toString());
+    forward.headers.set(CoreHeaders.DISABLE_PRETTY_ERROR, "");
     url.protocol = this.#runtimeEntryURL.protocol;
     url.host = this.#runtimeEntryURL.host;
     if (forward.cf) {
--- a/packages/miniflare/src/index.ts
+++ b/packages/miniflare/src/index.ts
@@ -1155,7 +1155,7 @@ export class Miniflare {
     const forward = new Request(input, init);
     const url = new URL(forward.url);
     forward.headers.set(CoreHeaders.ORIGINAL_URL, url.toString());
-    forward.headers.set(CoreHeaders.DISABLE_PRETTY_ERROR, "");
+    forward.headers.set(CoreHeaders.DISABLE_PRETTY_ERROR, "true");
     url.protocol = this.#runtimeEntryURL.protocol;
     url.host = this.#runtimeEntryURL.host;
     if (forward.cf) {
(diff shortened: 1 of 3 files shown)


packages/miniflare/src/workers/core/entry.worker.ts
```diff
@@ -208,11 +209,11 @@ export default <ExportedHandler<Env>>{
...
     if (isProxy) return handleProxy(request, env);
 
-    // `dispatchFetch()` will always inject the passed URL as a header. When
+    // `dispatchFetch()` will always inject this header. When
     // calling this function, we never want to display the pretty-error page.
     // Instead, we propagate the error and reject the returned `Promise`.
-    const isDispatchFetch =
-      request.headers.get(CoreHeaders.ORIGINAL_URL) !== null;
+    const disablePrettyErrorPage =
+      request.headers.get(CoreHeaders.DISABLE_PRETTY_ERROR) !== null;
```
RamIdeas, line 216: There was no example to follow with boolean headers so I used the existing check, as is.

Another options is:
```suggestion
    const enablePrettyErrorPage =
      Boolean(request.headers.get(CoreHeaders.PRETTY_ERROR) ?? 'true');
```

where `CoreHeaders.PRETTY_ERROR === "MF-Pretty-Error"` and `dispatchFetch` would set it with `forward.headers.set(CoreHeaders.PRETTY_ERROR, "false");`

packages/miniflare/src/index.ts
```diff
@@ -1155,6 +1155,7 @@ export class Miniflare {
     const forward = new Request(input, init);
     const url = new URL(forward.url);
     forward.headers.set(CoreHeaders.ORIGINAL_URL, url.toString());
+    forward.headers.set(CoreHeaders.DISABLE_PRETTY_ERROR, "");
```
mrbbot, line 1158: I know this isn't really needed, but might be nice to give this header a value.
```suggestion
    forward.headers.set(CoreHeaders.DISABLE_PRETTY_ERROR, "true");
```

packages/miniflare/src/workers/core/entry.worker.ts
mrbbot: This is great, it's what we do for `CoreHeaders.ERROR_STACK` earlier in this file:
https://github.com/cloudflare/miniflare/blob/868c60a088b7cc5e698d5bfc02c3c8af8e95d835/packages/miniflare/src/workers/core/entry.worker.ts#L79

mrbbot (approved):

packages/miniflare/src/index.ts
RamIdeas: This was my dilemma. If we give stringified values, we should infer them as booleans, else "false" would be considered truthy. And then we're faced with a double negative. So, what do you think of my suggestion above (PRETTY_ERROR instead of DISABLE_PRETTY_ERROR with values cast/inferred via `Boolean()`)?

mrbbot: I don't think that works, `Boolean("false") === true`. 😅 I'd just stick with presence detection for now, especially considering this header is meant to be internal, and only written and read in one place. 👍

RamIdeas: 🤦

PR #691, opened 2023-09-22
Fix `KV#list` with long (≥50ch) prefixes

### LINKED ISSUE #690: KV `list` with prefix over 50 chars causes SQL `LIKE or GLOB pattern too complex` error
Here is a `.list()` with a prefix less than 50 characters, which works correctly:

```ts
await environment.KV.list({
  prefix: "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx",
});
```

Adding one more `x`:

```ts
await environment.KV.list({
  prefix: "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx",
});
```

Causes:

```
[mf:err] Error: LIKE or GLOB pattern too complex
    at [object Object]
    at Object.list (/code/node_modules/.pnpm/miniflare@3.20230918.0/node_modules/miniflare/src/workers/shared/sql.worker.ts:81:14)
    at KeyValueStorage.list (/code/node_modules/.pnpm/miniflare@3.20230918.0/node_modules/miniflare/src/workers/shared/keyvalue.worker.ts:253:36)
    at KVNamespaceObject.list (/code/node_modules/.pnpm/miniflare@3.20230918.0/node_modules/miniflare/src/workers/kv/namespace.worker.ts:202:36)
    at KVNamespaceObject.fetch (/code/node_modules/.pnpm/miniflare@3.20230918.0/node_modules/miniflare/src/workers/shared/router.worker.ts:42:54)
    at KVNamespaceObject.fetch (/code/node_modules/.pnpm/miniflare@3.20230918.0/node_modules/miniflare/src/workers/shared/object.worker.ts:132:26)
```

This is very likely to have been caused by afe3da0, which changed [this line](https://github.com/cloudflare/miniflare/blob/afe3da0e82cdf72834bb1702bb3a491d926a9c6e/packages/miniflare/src/workers/shared/keyvalue.worker.ts#L81) to `db.stmt`.

(When I attach a debugger in VS Code—thank you!—it unwraps the following `workerd` error: `workerd/jsg/util.c++:275: error: e = workerd/util/sqlite.c++:830: failed: expected _ec != SQLITE_MISUSE [21 != 21]; SQLite misused: sqlite3_bind_text(statement, i+1, text.begin(), text.size(), SQLITE_STATIC)`, which is why I’m assuming it’s related to the change from `db.prepare`)

huw:
Resolves #690. See comments for more.

### DIFF
--- a/packages/miniflare/src/workers/r2/bucket.worker.ts
+++ b/packages/miniflare/src/workers/r2/bucket.worker.ts
@@ -15,7 +15,6 @@ import {
   all,
   base64Decode,
   base64Encode,
-  escapeLike,
   get,
   maybeApply,
   readPrefix,
@@ -230,12 +229,12 @@ function sqlStmts(db: TypedSql) {
     ];
     // TODO: consider applying same `:start_after IS NULL` trick to KeyValueStore
     return db.stmt<
-      { limit: number; escaped_prefix: string; start_after: string | null },
+      { limit: number; prefix: string; start_after: string | null },
       Omit<ObjectRow, "blob_id"> & Pick<ObjectRow, ExtraColumns[number]>
     >(`
       SELECT ${columns.join(", ")}
       FROM _mf_objects
-      WHERE key LIKE :escaped_prefix || '%' ESCAPE '\\'
+      WHERE substr(key, 1, length(:prefix)) = :prefix
       AND (:start_after IS NULL OR key > :start_after)
       ORDER BY key LIMIT :limit
     `);
@@ -379,7 +378,6 @@ function sqlStmts(db: TypedSql) {
     listMetadata: db.stmt<
       {
         limit: number;
-        escaped_prefix: string;
         start_after: string | null;
         prefix: string;
         delimiter: string;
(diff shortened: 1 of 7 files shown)


huw:
@mrbbot Is there an issue I can follow? I’m not sure if I made it clear but this completely blocks me from using local mode, which means I can’t adopt the Birthday Week features

huw:
I narrowed this down to https://github.com/cloudflare/workerd/blob/efa6c737e0aa78af29e353cecc88e05537cb44b3/src/workerd/util/sqlite.c%2B%2B#L526, which sets the `LIKE` clause limit to 50 characters for security reasons. Unless @kentonv is willing to raise it, I think the easier solution is just to avoid `LIKE` clauses unless we need complex patterns. As such, I updated the PR to run `WHERE SUBSTR(key, 1, LENGTH(:prefix)) = :prefix` instead and dropped the escaping code. I can’t see many reasons why it would be slower or otherwise undesirable, but you’ll know this better than I will. Thoughts?

mrbbot (approved):

huw:
Goated, cheers mate 🐐

Issue #698, opened 2023-09-17
🐛 BUG:


Silviu-Marian:
### Which Cloudflare product(s) does this pertain to?

Workers Runtime, Wrangler core

### What version(s) of the tool(s) are you using?

3.7.0

### What version of Node are you using?

18.6.0

### What operating system are you using?

Windows

### Describe the Bug

fresh install, default everything

main set to `worker.js` with a simple `addEventListener('fetch', event => { console.log('hello world') })`

navigate to `http://127.0.0.1:8787/NYSE:V`

500 error:

```
TypeError: Fetch API cannot load: nyse:V
    at Object.fetch (entry.worker.js:1004:36)
```

obviously the culprit is the semicolon in the URL but it shouldn't 500. 
briefly tested live though, and it works just fine when deployed, but not locally. 

what's wrong? what can we do? we need those semicolons to work in there.

mrbbot:
Hey! 👋 This should be fixed by #699, which will be included in the next release. Please note, Miniflare 3 now lives in the `cloudflare/workers-sdk` repo, and new releases will be published there.

PR #699, opened 2023-09-29
Ensure path is relative

penalosa:
Fixes https://github.com/cloudflare/miniflare/issues/698. Previously paths which included a colon were being parsed as hostname + port

### DIFF
--- a/packages/miniflare/src/workers/core/entry.worker.ts
+++ b/packages/miniflare/src/workers/core/entry.worker.ts
@@ -40,7 +40,7 @@ function getUserRequest(
     // If a custom `upstream` was specified, make sure the URL starts with it
     let path = url.pathname + url.search;
     // Remove leading slash, so we resolve relative to `upstream`'s path
-    if (path.startsWith("/")) path = path.substring(1);
+    if (path.startsWith("/")) path = `./${path.substring(1)}`;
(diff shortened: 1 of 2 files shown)


RamIdeas (approved):

packages/miniflare/src/workers/core/entry.worker.ts
```diff
@@ -40,7 +40,7 @@ function getUserRequest(
     // If a custom `upstream` was specified, make sure the URL starts with it
     let path = url.pathname + url.search;
     // Remove leading slash, so we resolve relative to `upstream`'s path
-    if (path.startsWith("/")) path = path.substring(1);
+    if (path.startsWith("/")) path = `./${path.substring(1)}`;
```
RamIdeas, line 43: That must've been fun to figure out 😅

RamIdeas (approved):

packages/miniflare/src/workers/core/entry.worker.ts
mrbbot, line 43: Could you add a regression test here too?
https://github.com/cloudflare/miniflare/blob/92c22a0aad5c80b6d1d1b07cc6b598785a42c4ba/packages/miniflare/test/index.spec.ts#L518

mrbbot (approved):

PR #708, opened 2023-10-05
[Miniflare 3] Add `getFetcher()` for dispatching `fetch`/`scheduled`/`queue` events

mrbbot:
Hey! 👋  This change adds back support for dispatching `scheduled` and `queue` events directly. Miniflare 2 previously provided similar `dispatchScheduled()` and `dispatchQueue()` methods, but these implemented an inconsistent, non-standard API.

With `getFetcher()`, we're able to reuse magic proxy code to support arbitrary Workers APIs. This is important for `queue()`, which supports sending any structured serialisable as a message `body`.

`getFetcher()` also provides an idiomatic Miniflare API for dealing with multiple workers, matching that provided by `getBindings()`.

### DIFF
--- a/packages/miniflare/README.md
+++ b/packages/miniflare/README.md
@@ -598,6 +598,18 @@ defined at the top-level.
   bindings, for all bindings in the Worker with the specified `workerName`. If
   `workerName` is not specified, defaults to the entrypoint Worker.
 
+- `getFetcher(workerName?: string): Promise<Fetcher>`
+
+  Returns a `Promise` that resolves with a
+  [`Fetcher`](https://workers-types.pages.dev/experimental/#Fetcher) pointing to
+  the specified `workerName`. If `workerName` is not specified, defaults to the
+  entrypoint Worker. Note this `Fetcher` uses the experimental
+  [`service_binding_extra_handlers`](https://github.com/cloudflare/workerd/blob/1d9158af7ca1389474982c76ace9e248320bec77/src/workerd/io/compatibility-date.capnp#L290-L297)
+  compatibility flag to expose
+  [`scheduled()`](https://workers-types.pages.dev/experimental/#Fetcher.scheduled)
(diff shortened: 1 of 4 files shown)


packages/miniflare/src/index.ts
```diff
@@ -1399,6 +1400,21 @@ export class Miniflare {
 
     return bindings as Env;
   }
+  async getFetcher(workerName?: string): Promise<ReplaceWorkersTypes<Fetcher>> {
```
mrbbot, line 1403: Not 100% sure on this name, maybe `getWorker()` or `getService()`?

packages/miniflare/README.md
```diff
@@ -598,6 +598,18 @@ defined at the top-level.
...
+
+  Returns a `Promise` that resolves with a
+  [`Fetcher`](https://workers-types.pages.dev/experimental/#Fetcher) pointing to
+  the specified `workerName`. If `workerName` is not specified, defaults to the
+  entrypoint Worker. Note this `Fetcher` uses the experimental
+  [`service_binding_extra_handlers`](https://github.com/cloudflare/workerd/blob/1d9158af7ca1389474982c76ace9e248320bec77/src/workerd/io/compatibility-date.capnp#L290-L297)
+  compatibility flag to expose
+  [`scheduled()`](https://workers-types.pages.dev/experimental/#Fetcher.scheduled)
+  and [`queue()`](https://workers-types.pages.dev/experimental/#Fetcher.queue)
+  methods for dispatching `scheduled` and `queue` events.
```
mrbbot, line 611: Might be nice to add an example here, with a `<details>`.

penalosa (approved):

packages/miniflare/src/index.ts
RamIdeas: ➕ `getWorker`

PR #711, opened 2023-10-08
fix: unhandled error thrown by busboy

b-marques:
As per, https://fetch.spec.whatwg.org/#dom-body-formdata, TypeErrors are expected to be thrown. Busboy errors were being unhandled.

### DIFF
--- a/packages/core/src/standards/http.ts
+++ b/packages/core/src/standards/http.ts
@@ -311,41 +311,51 @@ export class Body<Inner extends BaseRequest | BaseResponse> {
       );
     }
     const formData = new FormData();
-    await new Promise<void>(async (resolve) => {
-      const Busboy: typeof import("busboy") = require("busboy");
-      const busboy = Busboy({
-        headers: headers as http.IncomingHttpHeaders,
-        preservePath: true,
-      });
-      busboy.on("field", (name, value) => {
-        formData.append(name, value);
-      });
-      busboy.on("file", (name, value, info) => {
-        const { filename, encoding, mimeType } = info;
-        const base64 = encoding.toLowerCase() === "base64";
-        const chunks: Buffer[] = [];
-        let totalLength = 0;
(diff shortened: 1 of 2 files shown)


packages/core/src/standards/http.ts
```diff
@@ -311,41 +311,51 @@ export class Body<Inner extends BaseRequest | BaseResponse> {
...
+            });
+          });
+          busboy.on("finish", resolve);
+
+          const body = this[_kInner].body;
+          if (body !== null)
+            for await (const chunk of body) busboy.write(chunk);
+          busboy.end();
+        } catch (e) {
+          reject(e);
```
mrbbot, line 352: This will need to be converted to a `TypeError` here. If the `Promise` is rejected, the `if (busboyPromiseResult instanceof Error) {` line won't be hit:
```suggestion
          reject(new TypeError(e instanceof Error ? e.message : String(e)));
```

b-marques:
@mrbbot FYI, from Nodejs v16.19.0 onwards, the undici is included with the `multipart/form-data` parser.

mrbbot (approved):

packages/core/test/standards/http.spec.ts
```diff
@@ -382,7 +382,31 @@ test("Body: formData: respects Content-Transfer-Encoding header for base64 encod
...
+  const body = new Body(
+    new BaseResponse(["second value2", "--boundary--"].join("\r\n"), {
+      headers: { "content-type": "application/json" },
+    })
+  );
+  await t.throwsAsync(body.formData()),
+    {
+      instanceOf: TypeError,
+      message: "Unsupported content type: application/json",
+    };
```
mrbbot, line 408: These expectations aren't actually included in the `throwsAsync()`.
```suggestion
  await t.throwsAsync(body.formData(), {
    instanceOf: TypeError,
    message: "Unsupported content type: application/json",
  });
```

PR #712, opened 2023-10-09
Miniflare3: Add Content-Length and FixedLengthStream to R2 get response

gabivlj:
Before this, we would see that the following code wouldn't work:
```
  const res = await env.BUCKET.get(name);
  await env.BUCKET.put(name, res!.body);
```

As it would throw:
	TypeError: Provided readable stream must have a known length (request/response body or readable half of FixedLengthStream)


mrbbot (approved):
Looks good! As part of reviewing this, I wrote a test. Rather than asking you to rewrite this, I've just committed it. 😃

PR #718, opened 2023-10-17
Add Hyperdrive Bindings

tmthecoder:
Adds `Hyperdrive` binding support in Miniflare.

### DIFF
--- a/packages/miniflare/src/plugins/hyperdrive/index.ts
+++ b/packages/miniflare/src/plugins/hyperdrive/index.ts
@@ -1,0 +1,109 @@
+import { z } from "zod";
+import { Service, Worker_Binding } from "../../runtime";
+import { Plugin } from "../shared";
+
+export const HYPERDRIVE_PLUGIN_NAME = "hyperdrive";
+
+export const HyperdriveSchema = z
+  .string()
+  .url()
+  .transform((urlString, ctx) => {
+    const url = new URL(urlString);
+    if (url.protocol === "") {
+      ctx.addIssue({
+        code: z.ZodIssueCode.custom,
+        message: "You must specify the database protocol - e.g. 'postgresql'.",
+      });
+    } else if (
+      url.protocol !== "postgresql:" &&
+      url.protocol !== "postgres:" &&
+      url.protocol !== ""
+    ) {
+      ctx.addIssue({
+        code: z.ZodIssueCode.custom,
+        message:
+          "Only PostgreSQL or PostgreSQL compatible databases are currently supported.",
+      });
+    }
+    if (url.host === "") {
+      ctx.addIssue({
+        code: z.ZodIssueCode.custom,
+        message:
+          "You must provide a hostname or IP address in your connection string - e.g. 'user:password@database-hostname.example.com:5432/databasename",
+      });
+    }
+    if (url.port === "") {
+      ctx.addIssue({
+        code: z.ZodIssueCode.custom,
+        message:
+          "You must provide a port number - e.g. 'user:password@database.example.com:port/databasename",
+      });
+    }
+    if (url.pathname === "") {
+      ctx.addIssue({
+        code: z.ZodIssueCode.custom,
+        message:
+          "You must provide a database name as the path component - e.g. /postgres",
+      });
+    }
+    if (url.username === "") {
+      ctx.addIssue({
+        code: z.ZodIssueCode.custom,
+        message:
+          "You must provide a username - e.g. 'user:password@database.example.com:port/databasename'",
+      });
+    }
+    if (url.password === "") {
+      ctx.addIssue({
+        code: z.ZodIssueCode.custom,
+        message:
+          "You must provide a password - e.g. 'user:password@database.example.com:port/databasename' ",
+      });
+    }
+    return {
+      database: url.pathname.replace("/", ""),
+      user: url.username,
+      password: url.password,
+      scheme: url.protocol.replace(":", ""),
+      host: url.host,
+      port: url.port,
+    };
+  });
+
+export const HyperdriveInputOptionsSchema = z.object({
+  hyperdrives: z.record(z.string(), HyperdriveSchema).optional(),
+});
+
+export const HYPERDRIVE_PLUGIN: Plugin<typeof HyperdriveInputOptionsSchema> = {
+  options: HyperdriveInputOptionsSchema,
+  getBindings(options) {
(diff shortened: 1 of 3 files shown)


packages/miniflare/src/plugins/hyperdrive/index.ts
```diff
@@ -0,0 +1,120 @@
...
+          port: url.port,
+        };
+      })
+  )
+  .optional();
+
+export const HYPERDRIVE_PLUGIN: Plugin<typeof HyperdriveInputOptionsSchema> = {
+  options: HyperdriveInputOptionsSchema,
+  getBindings(options) {
+    const configs = HyperdriveOptions.parse(options.hyperdrives ?? {});
```
mrbbot, line 88: `options` will already be validated at this point against this plugin's `options` schema, so you shouldn't need to call `parse()` again. Same in `getServices()` below.

packages/miniflare/src/plugins/hyperdrive/index.ts
```diff
@@ -0,0 +1,120 @@
...
+        },
+      })
+    );
+  },
+  getNodeBindings(options) {
+    const bindingNames = Object.keys(options.hyperdrives ?? {});
+    return Object.fromEntries(
+      bindingNames.map((name) => [name, kProxyNodeBinding])
+    );
+  },
```
mrbbot, line 109: As above, `Socket`s and `WritableStream`s can't really be proxied yet, so I'd remove these bindings for now:
```suggestion
  getNodeBindings() {
    return {};
  },
```

packages/miniflare/test/plugins/hyperdrive/index.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+test.beforeEach(async (t) => {
+  const ns = `${Date.now()}_${Math.floor(
+    Math.random() * Number.MAX_SAFE_INTEGER
+  )}`;
+  t.context.ns = ns;
+  t.context.hyperdrive = namespace(
+    ns,
+    await t.context.mf.getHyperdrive("hyperdrive")
+  );
+});
```
mrbbot, line 39: You shouldn't need to `namespace` anything here. `namespace()` wraps a binding with a proxy that automatically prefixes the first argument with `ns` when calling any method. Since `Hyperdrive` bindings only provide a single `connect()` method, and this doesn't take a `string` parameter, namespacing like this won't work. Considering we're removing the `getHyperdrive()` function too, I've advise returning a JSON-serialisable version of the hyperdrive binding, and then making requests to the worker in your tests.

```suggestion
const test = miniflareTest<{ hyperdrive: Hyperdrive }>(opts, async (global, env) => {
  return global.Response.json({
    connectionString: env.hyperdrive.connectionString,
    user: env.hyperdrive.user,
    ...
  });
});

test("configuration", async (t) => {
  const res = await t.context.mf.dispatchFetch("http://localhost/");
  const config = await res.json();
  t.not(config.connectionString, TEST_CONN_STRING);
  t.is(config.user, "user");
  ...
});
```

mrbbot (commented):
Hey! 👋 Thanks for adding support for this! 🎉 I've added some comments and suggestions. 👍

packages/miniflare/test/plugins/hyperdrive/index.spec.ts
tmthecoder: Should be set now

tmthecoder:
@mrbbot Thanks for the review! All the changes should be set

packages/miniflare/src/plugins/hyperdrive/index.ts
```diff
@@ -0,0 +1,109 @@
...
+    }
+    if (url.password === "") {
+      ctx.addIssue({
+        code: z.ZodIssueCode.custom,
+        message:
+          "You must provide a password - e.g. 'user:password@database.example.com:port/databasename' ",
+      });
+    }
+    return {
+      database: url.pathname.replace("/", ""),
```
mrbbot, line 64: I think `/`s are valid in database names. We probably just want to remove the first character in `pathname` here.

packages/miniflare/test/plugins/hyperdrive/index.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+      connectionString: env.hyperdrive.connectionString,
+      user: env.hyperdrive.user,
+      password: env.hyperdrive.password,
+      database: env.hyperdrive.database,
+      host: env.hyperdrive.host,
+    });
+  }
+);
+
+test("connectionString: different from configuration", async (t) => {
```
mrbbot, line 26: Non-blocking, but you might want to consider merging these tests into one.

mrbbot (commented):
Thanks for addressing those comments, a few more, then I'll approve. 👍

packages/miniflare/src/plugins/hyperdrive/index.ts
```diff
@@ -0,0 +1,109 @@
...
+    const url = new URL(urlString);
+    if (url.protocol === "") {
+      ctx.addIssue({
+        code: z.ZodIssueCode.custom,
+        message: "You must specify the database protocol - e.g. 'postgresql'.",
+      });
+    } else if (
+      url.protocol !== "postgresql:" &&
+      url.protocol !== "postgres:" &&
+      url.protocol !== ""
```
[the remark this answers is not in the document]
tmthecoder: I believe it already does. Isn't the `if` statement preceding the `else if` block checking for empty protocols and erroring. I think it just assumed `postgres:` to be the protocol and failed in the remainder of the parse. This logic was used directly from the `wrangler` implementation to create bindings (https://github.com/cloudflare/workers-sdk/pull/3999/files), so I don't think there should be much issue here

packages/miniflare/src/plugins/hyperdrive/index.ts
tmthecoder: This was also used directly from the PR I mentioned above

packages/miniflare/src/plugins/hyperdrive/index.ts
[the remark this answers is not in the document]
mrbbot: Ah I see! Whoops! 🤦

packages/miniflare/src/plugins/hyperdrive/index.ts
mrbbot: Fair enough 👍

mrbbot (approved):

PR #727, opened 2023-10-27
Fix Address Parsing for Hyperdrive

tmthecoder:
I mistakenly listed `url.host` as the hostname while it should have been `url.hostname` as the `url.host` contains the port, causing a formatting issue when we append the post to the address when creating the runtime service

### DIFF
--- a/packages/miniflare/src/plugins/hyperdrive/index.ts
+++ b/packages/miniflare/src/plugins/hyperdrive/index.ts
@@ -32,7 +32,15 @@ export const HyperdriveSchema = z
           "You must provide a hostname or IP address in your connection string - e.g. 'user:password@database-hostname.example.com:5432/databasename",
       });
     }
-    if (url.port === "") {
+    let port: string | undefined;
+    if (
+      url.port === "" &&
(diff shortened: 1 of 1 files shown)


packages/miniflare/src/plugins/hyperdrive/index.ts
```diff
@@ -32,7 +32,15 @@ export const HyperdriveSchema = z
           "You must provide a hostname or IP address in your connection string - e.g. 'user:password@database-hostname.example.com:5432/databasename",
       });
     }
-    if (url.port === "") {
+    let port;
```
mrbbot, line 35: `port` will be implicitly `any` typed here. You'll want to add an explicit type annotation:
```suggestion
    let port: string | undefined;
```

mrbbot (approved):

packages/miniflare/src/plugins/hyperdrive/index.ts
tmthecoder: Should be all good now, thanks!

PR #731, opened 2023-10-30
Version bump 3.20231025.1

RamIdeas:


mrbbot (changes requested):
You'll need to bump the version in `packages/miniflare/package.json` too and re-run `npm install` to update the lockfile.

penalosa:
I imagine this will need a corresponding package-lock change?

mrbbot (approved):

PR #744, opened 2023-11-16
Remove reference to `Miniflare#getDurableObjectStorage()`

mrbbot:
We don't support this in Miniflare 3 yet, so let's remove it from the docs for now.


RamIdeas (approved):
There were a lot of redundant formatting changes in this diff – was that prettier or did you do it manually? 😅

PR #746, opened 2023-11-28
(docs|fix): Fix incorrect link to legacy miniflare v2 docs

### LINKED ISSUE #745: [Docs] Miniflare v2 docs removed?
Given that Miniflare 3 currently does not integrate well with `Jest` and `vitest`, a lot of us are still using Miniflare v2 for unit test simulation of the cloudflare tools.

However, it looks like those old docs have all been removed? Previous links like https://miniflare.dev/testing/jest no longer work.

The official docs link also suggest using miniflare v2, but the link on the testing page (https://miniflare.dev/testing) redirects straight back to the miniflare v3 docs (the link in question is [https://v2.miniflare.dev/testing](https://v2.miniflare.dev/testing))

Not sure if this is an oversight or not, but it would be really great if we could still access the old miniflare v2 docs.

AdiRishi:
Resolves #745

### DIFF
--- a/docs/src/content/testing/index.md
+++ b/docs/src/content/testing/index.md
@@ -6,4 +6,4 @@ order: 5
 
 <!-- <DirectoryListing path="/testing"/> -->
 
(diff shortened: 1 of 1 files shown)


mrbbot (approved):
Thank you! 😃

PR #752, opened 2023-12-27
docs: update example in getting started

pi0:
Randomly noticed when blindly copy-pasting example there is an extra `)`. Also added `mf.dispose()` to make example more complete.


mrbbot (approved):
Great catch, thanks for fixing this! ✅

olljanat:
@pi0 do you happen to have more complete example how to get started with this? 

I mean things like:
1. Do I need create Worker project first and install to miniflare package to there?
2. Which file that example should be put.
3. Any other needed configuration files needed to get started.

I ended up to see just a lot of errors which does not make any sense.

PR #768, opened 2024-06-17
Docs: add missing worker variable declaration

unicoder88:
Hi! I was reading [Getting started](https://miniflare.dev/get-started) when I came across using undefined variable `worker`, which made me wondering where it comes from.
Tried examples locally and indeed it reproduces:
Screenshot 2024-06-17 at 19 01 34

Luckily, I found an example on [Scheduled Events](https://miniflare.dev/core/scheduled) page.

Please review :)


penalosa:
Thanks for this! We'll follow up and add this in the Cloudflare docs: https://github.com/cloudflare/cloudflare-docs/issues/20783

PR #789, opened 2024-12-07
Fixed code snippet mistake on docs

KyGuy2002:
It looks like the D1 prepared statement API requires a `.run()` to dispatch the query.  Not sure if this is a relic from an older API version or just a typo.


penalosa:
Thanks for this! We'll follow up and add this in the Cloudflare docs: https://github.com/cloudflare/cloudflare-docs/issues/20783

